PR #23, opened 2016-09-07
Add Array.First() Array.Last()

oalmali:
Hello gavv;

I've been enjoying httpexpect to test API's and today I've come across with a situation which I needed to get first and/or last element of an array within a JSON response. So you will find my PR contains related code and the tests which also covers an empty array case too.

Thanks for httpexpect again!


gavv:
Thanks, LGTM!

Issue #26, opened 2016-10-23
Prevent example dependencies from being vendored along side github.com/gavv/httpexpect


JamesMura:
When you try to install github.com/gavv/httpexpect in a vendor folder most tools with install all the recursive dependencies , even dependencies for packages only used for the examples , hence you get lots of packages in your vendor folder that you don't need to use the tool. Not sure how to get around this on my end , but some projects prefix the example folder with an underscore so that it is ignored by the build tool.

kataras:
Let me fix it

kataras:
Reference: https://github.com/gavv/httpexpect/pull/27 . Thanks!

gavv:
@JamesMura What package manager(s) are you using?

JamesMura:
I am using gb https://getgb.io/

gavv:
I've renamed `example/` to `_examples/`. It should work at least with `gb` and `glade`.

`gb` is now happy with the following dependencies:

```
└── vendor
    └── src
        ├── github.com
        │   ├── ajg
        │   │   └── form
        │   ├── davecgh
        │   │   └── go-spew
        │   ├── fatih
        │   │   └── structs
        │   ├── gavv
        │   │   ├── gojsondiff
        │   │   ├── httpexpect
        │   │   └── monotime
        │   ├── google
        │   │   └── go-querystring
        │   ├── imkira
        │   │   └── go-interpol
        │   ├── klauspost
        │   │   ├── compress
        │   │   ├── cpuid
        │   │   └── crc32
        │   ├── moul
        │   │   └── http2curl
        │   ├── pmezard
        │   │   └── go-difflib
        │   ├── sergi
        │   │   └── go-diff
        │   ├── stretchr
        │   │   └── testify
        │   ├── valyala
        │   │   ├── bytebufferpool
        │   │   └── fasthttp
        │   ├── xeipuuv
        │   │   ├── gojsonpointer
        │   │   ├── gojsonreference
        │   │   └── gojsonschema
        │   ├── yalp
        │   │   └── jsonpath
        │   └── yudai
        │       └── golcs
        └── golang.org
            └── x
                └── net
```

gavv:
Closing this, feel free to reopen.

PR #27, opened 2016-10-23
Fix https://github.com/gavv/httpexpect/issues/26

kataras:
Prevent example dependencies from being vendored along side github.com/gavv/httpexpect

### DIFF
--- a/example/echo.go
+++ b/example/echo.go
@@ -1,3 +1,5 @@
+// +build ignore
+
 package example
 
 import (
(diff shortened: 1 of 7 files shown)


mnemonicflow:
I think it would be a good idea to squash all the commits into a single commit

gavv:
@kataras 

I've renamed `example/` to `_examples/` instead. It should work at least with `gb` and `glade`.

Let me know if there is any case when `build ignore` works better.

kataras:
@gavv Ofcourse,  what ever you think, I just went to solve @JamesMura 's problem

PR #31, opened 2016-12-20
Add Google App Engine Testing Instance support (with RequestFactory interface{})

mattes:
This PR adds support for Google App Engine Testing.

#### Background
Each `*http.Request` needs to be registered with the GAE test instance. Otherwise calls to `appengine.NewContext(*http.Request)` will panic with:

```
appengine: NewContext passed an unknown http.Request
```

#### References
* https://github.com/golang/appengine/blob/master/aetest/instance_vm.go#L60
* https://cloud.google.com/appengine/docs/go/tools/localunittesting/

#### Changes

* Add `GaeTestInstance` to `struct Config`
* If `Config.GaeTestInstance` is specified (not nil), use `GaeTestInstance.NewRequest` to create `*http.Request` that is associated with the GAE test instance.
* Otherwise create plain `*http.Request`.
* In order to keep the correct reference, `struct Request` needs pointer to request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Workflow:
 * URL query parameters (encoding using [`go-querystring`](https://github.com/google/go-querystring) package).
 * Headers, cookies, payload: JSON,  urlencoded or multipart forms (encoding using [`form`](https://github.com/ajg/form) package), plain text.
 * Create custom [request builders](#reusable-builders) that can be reused.
+* Create requests for Google App Engine testing with `RequestFactory`.
 
 ##### Response assertions
 
--- a/_examples/gae_test.go
+++ b/_examples/gae_test.go
@@ -1,0 +1,67 @@
+package examples
+
+import (
+	"github.com/gavv/httpexpect"
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/aetest"
+	"net/http"
+	"os"
+	"testing"
+)
+
+// init is used by GAE to start serving the app
+// added here for illustration purposes
+// func init() {
+// 	http.Handle("/", Router())
+// }
+
+func Router() http.Handler {
+	m := http.NewServeMux()
+
+	m.HandleFunc("/ping", func(w http.ResponseWriter, r *http.Request) {
+		_ = appengine.NewContext(r)
+		w.Write([]byte("pong"))
+	})
+
+	return m
+}
+
+// GaeInstance is our global dev_appserver instance
+var GaeInstance aetest.Instance
+
+// TestMain is called first to create the GaeInstance
+func TestMain(m *testing.M) {
+
+	// INFO: Remove the return to actually run the tests.
+	// Requires installed Google Appengine SDK.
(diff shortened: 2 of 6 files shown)


mattes:
This works for me so far. `_examples/gae.go` is more like an example usage, not actual tests. Tests would require the Google App Engine Go SDK. 

Leaving this PR here for discussion and feedback. Happy to change things if necessary. Once finalized, I will change the import paths, too.

_examples/gae.go
```diff
@@ -0,0 +1,43 @@
...
+	GaeInstance.Close()
+	os.Exit(c)
+}
+
+func Router() http.Handler {
+	// return your actual handlers
+	return nil
+}
+
+func NewHttpExpect(t *testing.T) *httpexpect.Expect {
```
mattes: I wonder if this could be a "first class" func, like 

```
func NewGae(t *testing.T, r http.Handler, i aetest.Instance) *httpexpect.Expect
```
Thoughts?

gavv:
Hi, thanks for the PR!

I'd prefer to hide request creation behind an interface, say:

```
type RequestFactory interface {
    NewRequest(method, urlStr string, body io.Reader) (*http.Request, error)
}

...

type Config struct {
    ...
    RequestFactory RequestFactory
}
```

and use it in request.go instead of GaeTestInstance.

This interface is compatible with the aetest.Instance, thus the user can just pass the instance as a factory.

If the user didn't specify a factory, the default RequestFactory implementation should be used (say, DefaultRequestFactory) that just calls http.NewRequest.

With this approach, we can also eliminate aetest dependency.

gavv:
Meanwhile, we can still keep aetest dependency in examples.

gavv:
> I wonder if this could be a "first class" func, like 
> [...]
> Thoughts?

We could a more generic wrapper:

```
func WithHandler(t LoggerReporter, r http.Handler, f RequestFactory) *httpexpect.Expect
```

mattes:
Very good thoughts. I will update that.

mattes:
* Changed everything to use a `RequestFactory` interface
* Added tests
* Updated `_examples/gae_test.go` to include fully working example 
* Couldn't decide on the `func WithHandler(t LoggerReporter, r http.Handler, f RequestFactory) *httpexpect.Expect` idea, so left it out for now. Might open a new PR for it in the future. 

I think my work is done here :-)

gavv:
Great.

This PR had two issues:

* the Request struct was not prepared to handle nil http.Request, which is now possible since RequestFactory can fail

* it's not correct to use "path" as an URL in NewRequest, since it may contain `{named}` arguments for go-interpol that will be called later from WithPath

PR is merged and this two issues are fixed as well.

I've also extracted DefaultRequestFactory struct from NewRequest and edited README a bit.

mattes:
Awesome! Thanks for fixing the issues and merging! One of the cleanest code bases I've seen in a while. Keep it up!

PR #40, opened 2017-02-24
Update travis for go 1.8 - read description

kataras:
@gavv Most of the web frameworks (first Iris) uses explicity the http.Pusher interface, we need go 1.8 because we can't use the *http.PushOptions on go 1.7 too.  So I **removed** the go 1.7, if you don't agree with that just let it as it is*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: go
 sudo: false
 
 go:
-    - 1.7
+    - 1.8
 
 before_install:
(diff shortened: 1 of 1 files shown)


gavv:
Thanks. I've added go1.8 to the travis build. I don't want to drop go1.7 right now, so I kept it and disabled iris examples on go < 1.8.

kataras:
@gavv Excellent! Sorry for being late to answer, I'm just working on the next month's new, tactical, iris release

PR #49, opened 2018-01-30
Replace github.com/yalp/jsonpath with github.com/PaesslerAG/jsonpath

### LINKED ISSUE #48: switching to PaesslerAG/jsonpath
We love `httpexpect`, and are using it in our project. However, currently `httpexpect` depends on `github.com/yalp/jsonpath` which does not come with a license. From a legal standpoint, this is a big red flag, and we need a solution for us to be able to use `httpexpect`.

Someone raised this issue directly on the `jsonpath` repo (https://github.com/yalp/jsonpath/issues/3), but there has been no movement.

Can `httpexpect` remove the `jsonpath` dependency so that we can use this legally? One possibility might be to use another one like https://github.com/PaesslerAG/jsonpath which is licensed under BSD-3. Thoughts?

sjlee:
Fix #48.

It replaces the github.com/yalp/jsonpath dependency which does not have a license with github.com/PaesslerAG/jsonpath.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -6,8 +6,8 @@ import (
 	"reflect"
 	"regexp"
 
+	"github.com/PaesslerAG/jsonpath"
 	"github.com/xeipuuv/gojsonschema"
-	"github.com/yalp/jsonpath"
 	"github.com/yudai/gojsondiff"
 	"github.com/yudai/gojsondiff/formatter"
 )
@@ -28,7 +28,7 @@ func getPath(chain *chain, value interface{}, path string) *Value {
 		return &Value{*chain, nil}
 	}
 
-	result, err := jsonpath.Read(value, path)
+	result, err := jsonpath.Get(path, value)
 	if err != nil {
 		chain.fail(err.Error())
 		return &Value{*chain, nil}


gavv:
Thanks. LGTM, but I think I need to test it a bit before merging to ensure that we don't break compatibility.

sjlee:
Thanks. Absolutely.

gavv:
Hi, sorry for delay.

I've added some tests for jsonpath (based on original yalp/jsonpath tests).

Unfortunately, this small PR breaks the compatibility.

Below you can find the list of cases where yalp/jsonpath and PaesslerAG/jsonpath work differently.

Test JSON:

```
{
    "A":["string",23.3,3,true,false,null],
    "B":"value",
    "C":3.14,
    "D":{
        "C":3.1415,
        "V":["string2",{"C":3.141592}]
    }
}
```

```
special -1 index

expression:
$.A[:-1]

yalp/jsonpath:
["string", 23.3, 3.0, true, false]

PaesslerAG/jsonpath:
[]
```

```
unquoted key

expression:
$[A][0]

yalp/jsonpath:
"string"

PaesslerAG/jsonpath:
null
```

```
unquoted key

expression:
$[C,B]

yalp/jsonpath:
[3.14, "value"]

PaesslerAG/jsonpath:
[]
```

```
recursive search

expression:
$..C

yalp/jsonpath:
[3.1415, 3.141592]

PaesslerAG/jsonpath:
[3.14, 3.1415, 3.141592]
```

```
recursive search

expression:
$..["C"]

yalp/jsonpath:
[3.1415, 3.141592]

PaesslerAG/jsonpath:
[3.14, 3.1415, 3.141592]
```

You can check these cases using this online tester: https://codebeautify.org/jsonpath-tester.

So what we have so far:

* PaesslerAG/jsonpath doesn't handle special -1 index
* PaesslerAG/jsonpath doesn't handle unquoted keys
* yalp/jsonpath has a bug with recursive search

I'm not sure if the first two features are standard, but if we reject them we will break the compatibility.

So we have several options:

* stay on yalp/jsonpath (it seems that it has the LICENSE file now)
* fix PaesslerAG/jsonpath before switching to it
* add a possibility to switch the JSONPath engine (but this is an overkill I think)

gavv:
https://github.com/yalp/jsonpath/issues/6

gavv:
All bugs I've found in yalp/jsonpath are now fixed in upstream. Since it also has a LICENSE file, I guess there is no urgency now to make a switch.

On the other hand, all important JSONPath features are now covered with httpexpect tests, so making a switch would be safe.

Switching to PaesslerAG/jsonpath makes sense since it seems to be a more complete implementation. However, compatibility issues should be addressed first. If someone is interested, pull requests are welcome.

gavv:
Closing this for now.

PR #51, opened 2018-04-04
WIP: Basic WebSocket support

### LINKED ISSUE #50: WebSocket support
It would be nice to have one for testing WS connections.

tyranron:
Resolves #50

### Why this PR is here?

I haven't found any handy solutions for E2E [WebSocket]s testing on Go. The only solution I've found is [github.com/posener/wstest](https://github.com/posener/wstest) which is for unit tests only, however.

### What does this PR do?

Adds first-class library support for [WebSocket] testing. Built on top of [github.com/gorilla/websocket](https://github.com/gorilla/websocket) library, which is de facto Go standard for [WebSocket]s at the moment.

This supports both real [WebSocket] server implemented in any language (so allows E2E tests) and `http.Handler` / `fasthttp.RequestHandler` directly without running server (so allows unit tests).

#### Features

- Send WS HTTP request: `req := e.WS("/my/path/to/ws")`
- Obtain WS connection from response: `conn := req.Expect().Connection()`
- Close WS connection normally/abnormally: `conn.Close()` / `conn.Disconnect()`
- Write to WS connection: `conn.WriteText("hi")`, `conn.WriteJSON(object)`, `conn.WriteBinary(bytes)`, and so on...
- Read from WS connection: `msg := conn.Expect()`
- `Binder` / `FastBinder` implementations for `websocket.Dialer` (to test handlers directly without running WS server)
- `WsPrinter` interface extension for `Printer` which prints sent and received [WebSocket] messages (implemented for `DebugPrinter`)

#### Example

```go
func TestEchoServer(t *testing.T) {
	e := httpexpect.New(t, "http://echo.server/")

	conn := e.WS("/endpoint/ws").
		Expect().
		Status(http.StatusSwitchingProtocols).
		Connection()
	defer conn.Disconnect()

	conn.WriteText("hi").
		Expect().Text().Body().Equal("hi")

	conn.WriteJSON(struct {
		Message string `json:"message"`
	}{"hi"}).
		Expect().
		Text().JSON().Object().ValueEqual("message", "hi")

	conn.CloseWithText("bye")
}
```

### If you want to try it now

Modify your `Gopkg.toml`:

```toml
[[constraint]]
  name = "github.com/gavv/httpexpect"
  branch = "ws-support"
  source = "github.com/instrumentisto/httpexpect"
```

[WebSocket]: https://en.wikipedia.org/wiki/WebSocket

### DIFF
--- a/request.go
+++ b/request.go
@@ -19,8 +19,8 @@ import (
 	"github.com/fatih/structs"
 	"github.com/gavv/monotime"
 	"github.com/google/go-querystring/query"
-	"github.com/imkira/go-interpol"
 	"github.com/gorilla/websocket"
+	"github.com/imkira/go-interpol"
 )
 
 // Request provides methods to incrementally build http.Request object,
@@ -37,7 +37,7 @@ type Request struct {
 	bodysetter string
 	typesetter string
 	forcetype  bool
-	isWS       bool
+	isWs       bool
 }
 
 // NewRequest returns a new Request object.
@@ -96,9 +96,9 @@ func NewRequest(config Config, method, path string, pathargs ...interface{}) *Re
 		chain.fail(err.Error())
 	}
 
-	isWS := false
+	isWs := false
 	if method == "ws" {
-		isWS = true
+		isWs = true
 		method = http.MethodGet
 	}
 
@@ -112,7 +112,7 @@ func NewRequest(config Config, method, path string, pathargs ...interface{}) *Re
 		chain:  chain,
 		path:   path,
 		http:   hr,
-		isWS:   isWS,
+		isWs:   isWs,
 	}
 }
 
@@ -132,7 +132,7 @@ func (r *Request) WithClient(client Client) *Request {
 	switch {
 	case r.chain.failed():
 		return r
-	case r.isWS:
+	case r.isWs:
 		r.chain.fail(
 			"\nunexpected WithClient usage for WebSocket request, " +
 				"use WithDialer instead")
@@ -158,7 +158,7 @@ func (r *Request) WithDialer(dialer Dialer) *Request {
 	switch {
 	case r.chain.failed():
 		return r
-	case !r.isWS:
+	case !r.isWs:
 		r.chain.fail(
 			"\nunexpected WithDialer usage for non-WebSocket request, " +
 				"use WithClient instead")
@@ -183,7 +183,7 @@ func (r *Request) WithHandler(handler http.Handler) *Request {
 	switch {
 	case r.chain.failed():
 		return r
-	case r.isWS:
+	case r.isWs:
 		r.chain.fail(
 			"\nunexpected WithHandler usage for WebSocket request, " +
 				"which is not supported currently") // TODO: support maybe?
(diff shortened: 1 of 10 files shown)


tyranron:
@gavv I know this is huge one, but I'd like to ask you to look through the design and if you like it, I'll cover new stuff with unit tests and improve documentation.

request.go
```diff
@@ -827,22 +946,31 @@ func (r *Request) encodeRequest() {
...
+	resp *http.Response, conn *websocket.Conn, elapsed time.Duration,
+) {
 	if r.chain.failed() {
 		return
 	}
 
+	// TODO: print actual request made by Dialer
 	for _, printer := range r.config.Printers {
 		printer.Request(r.http)
 	}
```
tyranron: @gavv the only place I'm not sure what to do...

The problem is that `websocket.Dialer` [constructs the final HTTP request on-fly](https://github.com/gorilla/websocket/blob/v1.2.0/client.go#L199-L247) and there is no way to specify our request there. So, actually, we can't pass to `Printer` the real request (with `Upgrade` header and etc.), just preconfigured values.

The only option I see here is to repeat (copy-paste) `http.Request` construction from `gorilla/websocket`, but this will come with increased cost of maintaining (need to check changes in their lib every time).

gavv:
Hi,

I finally was able to find some time for httpexpect (well, ok, actually just quit my job :) and now I'm looking at the request.

The overall implementation looks very nice, but I'd like to rearrange and rename a few things.

I think I'll rebase and merge this in a few days.

gavv:
I've rebased and squashed these commits and merged them. I've also pushed several follow-up commits with some refactoring, fixes, and tests.

What I've changed:

* Renamed some structs and methods to be more similar to the rest code. The user code usually don't include httpexpect type names explicitly, so they can be a bit more verbose.

* Removed Expect.WS() and the special "ws" http method and added Request.WithWebsocketUpgrade(). Now the user should create a regular HTTP GET request, as usual, and then upgrade it to a WebSocket request using this method. I think this approach reflects better how the things work under the hood and it eliminates the need of a special magic http method.

* Removed multiple `isWs` checks from various Request methods that set the request body. There is a simpler way to do it -- we can check if the body is set when we're encoding a request (we even record who set the body to produce a nice error message).

* Refactored the request encoding and response creation a bit.

* Removed default WebSocket timeout. I decided to mimic the standard library here and don't provide any defaults just like in the default http.Transport.

* Removed the code that closes the connection and waits the final message from server when we're sending a close message to the server. This code didn't allow to match the final message from server in tests, which would be an expected feature I think. I there any reason why we really should have this code?

* Removed Content() method from message. It allowed to get a Value object from a non-JSON message, however the Value type is originally designed to represent *JSON values*. I there any reason why we really should have this method?

* Reworked dialer constructors a bit. First, there is no need to make them methods of binders, so I moved them out. Second, they now support multiple Dial calls. Third, they now start one goroutine per Dial call and wait until it finish in the Conn.Close method to provide a guarantee that the handler will not be used after the connection is closed.

What I've added:

* Some unit tests. The coverage is not 100% though. If you get a chance to add more, you're welcome :)

* Some end2end tests. They are based on the example from this request but extend it a bit.

* A note and a thank you in README about the WebSocket support.

gavv:
Related commits:

* 67c7006ee32d878fb24e77ddadbdd896e9046dbc
* 14adb4f5ab2eedf32d5de4ce020605c9fbe45724
* 2ef7ef17287bed8cf4d4cf0fad7d7076d96f2259
* fbbe5fc6ae1f2260fd7493fc419dc8ce00164c06
* c3000bc283cd364de398ca4e3f8d91f4a7591114

request.go
gavv: I thinks it's OK. Since the user can provide a custom dialer, we really don't control how a websocket request is formed, so it's not our responsibility to print it.

gavv:
One more follow-up commit: 82350969d5d20e9048a418cb62407d31bf74e953

gavv:
I'm closing this, thanks for contribution!

PR #57, opened 2019-09-25
chore: bump moul.io/http2curl dependency

moul:
See https://github.com/moul/http2curl/pull/19 for details


gavv:
Thanks, I'll take a look. There is a mess with go.mod: "go test" and "go mod tidy" are producing conflicting changes.

moul:
Ok, tell me if I can help

gavv:
Follow-up commit: 8387429857db3db0139e6f1b3811038b2cfe1aa8

gavv:
This fixes are also merged into `v1` and `v2` branches and two new releases are tagged: `v1.1.1` and `v2.0.1`.

PR #58, opened 2019-09-25
Fix http2curl import

Pantani:
# Headline
Fix http2curl import for `printer.go`

## Problem
`package github.com/moul/http2curl: code in directory $GOPATH/src/github.com/moul/http2curl expects import "moul.io/http2curl"`

## Solution
User the package `moul.io/http2curl` instead `github.com/moul/http2curl`

### DIFF
--- a/printer.go
+++ b/printer.go
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
-	"github.com/moul/http2curl"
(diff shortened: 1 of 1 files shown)


gavv:
Hi,

We already have a PR for this issue: #57

I have merged the fix to master, v1, and v2, see my comment in that PR.

Thanks.

Issue #63, opened 2019-12-16
Gin example


veeral-patel:
Hi, please consider adding an example of using `httpexpect` with `gin`: https://github.com/gin-gonic/gin!

Thanks for this great tool!

gavv:
Hi, thanks.

Unfortunately I haven't worked with gin, but pull requests are welcome :)

circa10a:
@gavv 
```go
// Init all context paths
func routes(router *gin.Engine) {
	root := router.Group("/")
	{
	root.GET("/rules", rulesHandler())
	// For every route not defined, forward to swagger docs
	router.NoRoute(noRouteHandler())
}

func TestRulesEndpoint(t *testing.T) {
	g := gin.Default()
        routes(g)
	server := httptest.NewServer(g)
	defer server.Close()
	e := httpexpect.New(t, server.URL)
	e.GET("/rules").
		Expect().
		Status(http.StatusOK).JSON().Array()
}
```

gavv:
PR is merged.

PR #64, opened 2019-12-28
Add Gin example

circa10a:
Addresses #63


gavv:
Thanks!

_examples/gin.go
```diff
@@ -0,0 +1,28 @@
...
+
+// Init routes, bind to global G variable to be used in tests
+func init() {
+	G.GET("/example", ginHandler())
+}
+
+// Start server
+func main() {
+	G.Run()
+}
```
gavv: All examples share the same package, so they should not define main(). Otherwise, only one example could have main. Can we construct gin instance in the test instead?

_examples/gin_test.go
```diff
@@ -0,0 +1,24 @@
+package examples
+
+import (
+	"github.com/gavv/httpexpect"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestGinHandler(t *testing.T) {
```
gavv: If we're adding a gin example, it would be reasonable to demonstrate something unique to gin.

One thing we could demonstrate is using httpexpext.NewBinder to test gin handler directly.

Other ideas?

README.md
```diff
@@ -102,6 +102,10 @@ See [`_examples`](_examples) directory for complete standalone examples.
 
     Testing a server running under the [Google App Engine](https://en.wikipedia.org/wiki/Google_App_Engine).
 
+* [`gin_test.go`](_examples/gin_test.go)
+
+    Testing a server utilizing the [gin web framework](https://github.com/gin-gonic/gin).
```
gavv: Nitpick: I suggest to move it upper, right after the echo example, so that framework examples will be grouped.

Makefile
```diff
@@ -1,5 +1,6 @@
 GO111MODULE := on
-export GO111MODULE
+GIN_MODE := release
```
gavv: Do we need it on build time or at runtime? If on build time, it's OK, but if at runtime, I'd suggest to set the var in the example itself instead of the makefile.

_examples/gin.go
circa10a: Good point. Test works fine without it. I've removed the `main` function.

_examples/gin_test.go
circa10a: I'm not familiar  with the `NewBinder` method. You have a snippet to demonstrate?

README.md
circa10a: Resolved. Have moved under echo

Makefile
circa10a: WIthout it, there's a couple of extra lines of debug output that gin puts out when it instantiates the server.
`os.Setenv("GIN_MODE", "release")` doesn't seem to resolve. I removed it from the `Makefile` though.

_examples/gin_test.go
gavv: * https://github.com/gavv/httpexpect#use-http-handler-directly
* https://godoc.org/github.com/gavv/httpexpect#Binder

https://github.com/gavv/httpexpect/blob/0387206792097a6dc6fd830ac268a6f1953194f1/_examples/echo_test.go#L72-L87

Makefile
gavv: I guess you need to call setenv before calling gin.Default(). Or maybe you can call gin.SetMode(gin.ReleaseMode) (if it works, it'd be preferred).

gavv:
Please also squash commits after fixing all the issues.

_examples/gin_test.go
circa10a: I've added the binder method for the test by following the echo example, the only issue I see with it is that it creates a new gin instance for each handler and doesn't feel very realistic, since you would normally use the same gin instance in every handler.

Makefile
circa10a: Got this working by using an `init` function and keeping the gin instance inside the handler.

_examples/gin_test.go
circa10a: I've modified the test to use `NewBinder` following the echo example. I did modify it a bit to pass along a `gin` instance to demonstrate that you would use the same instance in all of your route handlers and not create a new instance per route(hope that makes sense).

gavv:
LGTM, thanks!

PR #73, opened 2020-04-29
WithRetry capability

teivah:
First of all, thanks for this cool library.

In my context, we would have been interested in a retry capability before to assert that a test is failing. We are running a test, synchronizing an internal cache and then we trigger an HTTP request. As we don't have the possibility to query the internal cache, we don't know when the HTTP query will succeed. Hence, we had to add an ugly time.Sleep() before calling httpexpect library.

Instead, it would be way cleaner to be able to retry a failing query. That would be something similar to what we have for example in Gomega with the `Eventually` assertion: https://onsi.github.io/gomega/#making-assertions

`WithRetry` would accept a max retry value and a retry delay.

### DIFF
--- a/mocks_test.go
+++ b/mocks_test.go
@@ -6,13 +6,30 @@ import (
 )
 
 type mockClient struct {
-	req  *http.Request
+	req        *http.Request
+	resp       http.Response
+	err        error
+	retries    []retry
+	retryIndex int
+}
+
+type retry struct {
 	resp http.Response
 	err  error
 }
 
 func (c *mockClient) Do(req *http.Request) (*http.Response, error) {
 	c.req = req
+	if c.retries != nil {
+		currentRetry := c.retries[c.retryIndex]
+		c.retryIndex++
+		if currentRetry.err != nil {
+			return nil, currentRetry.err
+		}
+		c.resp.Header = c.req.Header
+		c.resp.Body = c.req.Body
+		return &currentRetry.resp, nil
+	}
 	if c.err == nil {
 		c.resp.Header = c.req.Header
 		c.resp.Body = c.req.Body
--- a/request.go
+++ b/request.go
@@ -38,6 +38,8 @@ type Request struct {
 	forceType  bool
 	wsUpgrade  bool
 	matchers   []func(*Response)
+	maxRetry   int
+	retryDelay time.Duration
 }
 
 // NewRequest returns a new Request object.
@@ -851,6 +853,18 @@ func (r *Request) WithMultipart() *Request {
 	return r
 }
 
+// WithRetry allows to retry an HTTP request until a max retry value.
+// It also accepts a retry delay that is performed before to retry on failure.
+func (r *Request) WithRetry(maxRetry int, retryDelay time.Duration) *Request {
+	if r.chain.failed() {
+		return r
+	}
+
+	r.maxRetry = maxRetry
+	r.retryDelay = retryDelay
+	return r
+}
+
 // Expect constructs http.Request, sends it, receives http.Response, and
 // returns a new Response object to inspect received response.
 //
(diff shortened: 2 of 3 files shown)


teivah:
@gavv Any problem with the approach?

gavv:
Hi, being a bit busy currently with other projects. I'll find time for httpexpect in week or two. Sorry for inconvenience and thanks for the PR!

teivah:
No worries, thank you :)

request.go
```diff
@@ -980,7 +994,20 @@ func (r *Request) sendRequest() *http.Response {
...
 	}
 
-	resp, err := r.config.Client.Do(r.http)
+	var err error
+	var resp *http.Response
+	if r.maxRetry == 0 {
+		resp, err = r.config.Client.Do(r.http)
+	} else {
+		for i := 0; i < r.maxRetry; i++ {
+			resp, err = r.config.Client.Do(r.http)
```
gavv, line 1003: The problem here is that, in general, http.Request is not reusable after calling http.Client.Do, even if it failed. In particular, http.Request.Body is io.Reader, and http.Client may read the whole body, so that the retry will see an EOF, like the body is empty.

This problem doesn't appear with requests without body. We should add a test with a POST request with non-empty body which will catch this bug.

The typical way to solve this problem is:

1. read the whole body into buffer
2. create a new reader from the buffer
3. call Client.Do
4. on failure, go to step (2)

Some optimizations are possible, e.g. if the client buffer is ReaderSeeker, we can just seek to the beginning before each retry.

Quick googling showed that there is also a package the handles this issue: https://github.com/hashicorp/go-retryablehttp

Maybe we can use it here? Or at least reproduce part of its logic.

request.go
```diff
@@ -851,6 +853,18 @@ func (r *Request) WithMultipart() *Request {
 	return r
 }
 
+// WithRetry allows to retry an HTTP request until a max retry value.
+// It also accepts a retry delay that is performed before to retry on failure.
+func (r *Request) WithRetry(maxRetry int, retryDelay time.Duration) *Request {
```
gavv, line 858: Is the retryDelay parameter useful in practice? (In your case or in general). Maybe instead remove it and use exponential back-off, to make the API a bit simpler?

request.go
```diff
@@ -980,7 +994,20 @@ func (r *Request) sendRequest() *http.Response {
...
+	var resp *http.Response
+	if r.maxRetry == 0 {
+		resp, err = r.config.Client.Do(r.http)
+	} else {
+		for i := 0; i < r.maxRetry; i++ {
+			resp, err = r.config.Client.Do(r.http)
+			if err != nil {
+				time.Sleep(r.retryDelay)
+				continue
+			}
```
gavv, line 1007: It usually makes sense to retry only temporary network errors, like:

```
if nerr, ok := err.(net.Error); ok && nerr.Temporary() {
```

Also, it makes sense to retry timeout errors (err.Timeout()), I don't remember whether timeout error is considered temporary or not.

In addition, it usually makes sense to retry some of HTTP status codes, likely all 5xx errors except 501. However, for httpexpect it doesn't always make sense, so probably we need a separate option for this.

I guess we need to allow the user to specify something like "retry policy" (what to retry and how).

mocks_test.go
```diff
@@ -6,13 +6,30 @@ import (
...
+
+type retry struct {
 	resp http.Response
 	err  error
 }
 
 func (c *mockClient) Do(req *http.Request) (*http.Response, error) {
 	c.req = req
+	if c.retries != nil {
+		currentRetry := c.retries[c.retryIndex]
```
gavv, line 24: Let's add an explicit panic with meaningful message when index is out of bounds.

request_test.go
```diff
@@ -165,6 +167,53 @@ func TestRequestClient(t *testing.T) {
 	req3.chain.assertFailed(t)
 }
 
+func TestRequestWithRetrySuccess(t *testing.T) {
```
gavv, line 170: Can we also check the resulting `retryIndex` in these tests? To ensure that httpexpect made the desired number of retry attempts.

request.go
teivah: Not necessarily. Please check at my use-case in the description. It's not simply a transient retry.

gavv: I see. But I guess this is not the most common use case, so maybe handle it as a special retry policy?

E.g. we could have the following policies:

* retry temporary network errors
* retry temporary network errors and temporary server errors (5xx) - likely should be default when retrying is enabled
* retry any errors

What exact error are you retrying in your case?

gavv:
I've added an issue: #85.

Issue #69, opened 2020-04-02
Rework error reporting and formatting


gavv:
## Problem

Currently we're missing the following features:

* the user can't register a hook that will be invoked on every performed assertion including successful ones; e.g. this could be used for goconvey integration (#143) or for other reasons;

* when an error is passed to user-defined `Reporter`, it only receives the error message, but no additional context: e.g. test name (#59), request and response (#137), value which didn't pass the assertion, etc;

* since every component formats error message by its own, there is no single point where we can implement various formatting improvements like colored output.

## Solution

The suggested solution is to unify error handling by introducing error context and formatter:

* Add `Context` struct:

  * Context will accumulate the assertion context, assertion name (i.e. method name like "Equal"), request, response, round trip time, etc. E.g., when the user calls Expect(), the encoded request and received response are stored into context.

  * Context should be inherited by nested objects, e.g. in Array().Element() the returned element inherits the context of the array.

  * Context will be passed to Formatter when an assertion fails.

  * Context will contain `Reporter` field, which will be used to report failures.

  * Context will also contain private `Formatter` field, to allow nested components to inherit formatter.

  * We already have struct `chain` which is automatically inherited by all nested objects. We should just replace `chain.reporter` with `chain.context`.

* Add `Failure` struct:

  * Failure will contain information about the failed assertion: error message, actual value, and expected value (for assertions like Equal) or the value which triggered the failure (for assertions like Contains).

  * Failure will be passed to Formatter when an assertion fails.

* Add `Formatter` interface with four methods:

      BeginAssertion(Context)
      Success(Context)
      Failure(Context, Failure)
      EndAssertion(Context)

* Add `DefaultFormatter` implementation of the `Formatter` interface:

  * BeginAssertion(), EndAssertion(), and Success() may be no-op for now.

  * Failure() should format info from Context and Failure structs into a string and pass it to Context.Reporter.

  * It should use dumpValue() to print values from Failure struct. If both actual and expected values are present, it should use diffValues() to print the diff. Currently this logic is duplicated in every component and now it should be implemented once in Formatter.

* Switch all components to new structs:

  * Make all components using Context.

  * Add Formatter to Config. If it's nil, automatically use DefaultFormatter. Pass Formatter and Reporter from Config to to Context.

  * Incrementally, one-by-one, switch every component from doing its own error formatting and reporting to using Context.Formatter. They should call BeginAssertion() before doing a check, then call Success() and Failure(), and then call EndAssertion().

## Notes

[...]

PR #78, opened 2020-06-28
Default Formatter implementation

goku321:
Added default implementation for `Formatter` interface.

Fixes: #69


expect.go
```diff
@@ -237,6 +238,50 @@ func (DefaultRequestFactory) NewRequest(
...
+// added later.
+func (DefaultFormatter) EndAssertion(ctx Context) {}
+
+// Failure implements Formatter.Failure and reports failure.
+//
+// It formats the info from Context and Failure struct into
+// a string and passes it to Context.Reporter for reporting.
+func (DefaultFormatter) Failure(ctx Context, f Failure) {
+	errString := ""
+	if f.actual != nil {
```
gavv, line 268: I think we should also add assertionName to Failure struct (e.g. "Equals", "Contains", etc) and print it here.

gavv: Also, I suggest to rename Context.name to Context.TestName. And I guess we need to make all Context fields, except formatter, exported.

gavv (commented):
Hi, sorry for delay! If you're still interested in this, here are a few comments.

goku321:
Hi @gavv Thanks for your comments. Yes, I'm still interested but caught up with other tickets at the moment. Will surely return to this as soon as possible.

Leryan:
@goku321 @gavv hi 👋 i’m very interested in these features, i could bring help if task sharing is possible :) i haven’t checked the current state of this PR in detail but i'm ready to spend several hours during the weekends. just ping me :)

gavv:
@Leryan Thanks for your interest, the help is appreciated! You could help with finishing this PR, I hope @goku321 don't mind. After this is done, it would be possible to parallelize the work of switching various components to the Context, since those components can be changed independently.

expect.go
Leryan: the test (`testing.T`) reference could be inside the context?

gavv:
Closing with respect to #150. See details in https://github.com/gavv/httpexpect/pull/98#issuecomment-1320941335

Issue #68, opened 2020-02-26
Support adding middleware to finalize HTTP requests


mrosales:
This request might seem a little strange, so I'll outline the use case first. I have an API that requires the requests to sign the body, and I'd like to use `httpexpect` to create one of these signed requests.

In order to do this, I need to construct a signature from the request body, http headers, full path, query string, HTTP protocol. See the [AWS Request Signing docs](https://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html) for an example of a similar scheme.

I'm not suggesting something too specific to my use case, but rather suggesting to implement something like a "request finalizer" or "request middleware" that lets you perform a function to modify the outgoing HTTP request object. I currently have to viable workarounds.

1. Prepend a logger, since is called basically when I want:

```go

func TestSignedRequest(t *testing.T) {
	// ...

	e := httpexpect.WithConfig(httpexpect.Config{
		BaseURL: baseURL,
		Client: &http.Client{
			Jar:     httpexpect.NewJar(),
			Timeout: time.Second * 30,
		},
		Reporter: httpexpect.NewAssertReporter(t),
		Printers: []httpexpect.Printer{
			requestFinalizerFunc(func(r *http.Request) {
				// copy request body
				// compute signature headers
				// set header
			}),
			httpexpect.NewCompactPrinter(t),
		},
	})

	// e.POST(...)
}


type requestFinalizerFunc func(r *http.Request)

// Request implements httpexpect.Printer.Request.
func (finalizer requestFinalizerFunc) Request(req *http.Request) {
	if req != nil {
		finalizer(req)
	}
}
```

2. Override the `http.Client`, so we can transform the outgoing request.

I opted for the logger approach since it allows the request to be transformed after it is constructed but before being logged or sent. Would you consider adding a hook that exposes something like the following:

```
e.POST("/foo/bar").
    WithBytes(someBytes).
    WithRequestTransformer(func(r *http.Request){
        // modify the request here
    })
```

It could be used with something like:

```
e := httpexpect.New(t, "http://example.com")

signed := e.Builder(func (req *httpexpect.Request) {
	req.WithRequestTransformer(func(r *http.Request) {
		// sign my request
	})
})

signed.POST("/foo/bar").WithBytes([]byte("some bytes to sign"))
```

gavv:
Hi, sorry for late reply.

Yes, `WithRequestTransformer` idea looks good! I'll assign `help wanted` tag so that someone could implement it.

[...]

PR #79, opened 2020-07-09
feat: adds WithRequestTransformer

t0mburton:
I found myself needing a way to construct a header based on the request body and noticed this concept had been discussed under https://github.com/gavv/httpexpect/issues/68.

Nice lib btw, I hope the PR is how you envisaged it :+1:

### DIFF
--- a/request.go
+++ b/request.go
@@ -851,6 +851,26 @@ func (r *Request) WithMultipart() *Request {
 	return r
 }
 
+// WithRequestTransformer executes the given transform on the underlying http.Request.
+//
+// Example:
+//  req := NewRequest(config, "PUT", "http://example.com/path")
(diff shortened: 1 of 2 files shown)


gavv:
Sorry for delay.

I think it'd be more reasonable to apply transformers only after encoding request, i.e. in Expect call. This way transformers will play well together with other request builders (e.g. WithPath or WithQuery) - they will see the final result of all other builders and will be able to override it.

Since I was so late, I've rebased and fixed PR by myself.

Everything merged into master, here are commits: 7891e197e7d7b713fc8a0fab78ca3ca3b77075d4 (original commit, rebased), b11e30f2c9e3f19dd84b4db9bc05b03b779268ef (my follow-up).

Thanks!!

PR #71, opened 2020-04-16
Add Context, Failure, Formatter types

goku321:
Added following types:

1. `Context` struct
2. `Failure` struct
3. `Formatter` interface

Fixes: #69

PR #82, opened 2020-10-02
Add Context, Failure, Formatter types (#71)

Vedurumudi-Priyanka:
added context, failure & formatter types


gavv:
Hi, sorry, but we already have #71 and #78. Thanks.

Issue #85, opened 2020-12-21
Add retrying capability


gavv:
Add three new Request methods:

* WithRetry(maxRetry int)
* WithRetryDelay(minDelay, maxDelay time.Duration)
* WithRetryPolicy(RetryNetworkErrors|RetryNetworkAndServerErrors|RetryAnyErrors)

WithRetry enabled retrying. If it's not called, test fails after first round trip failure. If it's called, test will try up to maxRetry times and only then fail.

WithRetryDelay controls pauses between retries. If it's not called, default pause is 1 second. If it's called, pause starts from minDelay and grows exponentially until it reaches maxDelay or we're out of retries.

WithRetryPolicy controls what to retry. If it's not called, default policy is RetryNetworkAndServerErrors, i.e. retry temporary network errors and temporary server errors (5xx). RetryNetworkErrors means retry only temporary network errors. RetryAnyErrors means retry absolutely all errors.

Previous attempt was here: #73. In particular, see https://github.com/gavv/httpexpect/pull/73#discussion_r434367446

We should also add documentation comments for new methods, cover them with tests (ideally both unit and e2e, but at least e2e), and add a new section to quick start in README.

ShreyanGoswami:
Hello, can I give it a try?

bmihaylov:
Hey,
taking a look at your review of the previous attempt I saw the reference to the `retryablehttp` lib.
I poked around and it turns out they introduced the ability to get their client wrapped as a standard `http.Client`.
Combining that with `httpexpect` ability to set client on a per `Request` or `Expect` instance basis there is no need to change `httpexpect` .
Here is a minimal example which would suffice for a lot of cases in my opinion, given that their defaults seem sensible:
```go
package main

import (
	"net/http"
	"net/http/httptest"
	"testing"

	"github.com/gavv/httpexpect/v2"
	rhttp "github.com/hashicorp/go-retryablehttp"
)

func TestRetry(t *testing.T) {
	fail := true
	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if fail {
			fail = false
			http.Error(w, "Try again!!", 500)
		}
	})
	server := httptest.NewServer(handler)
	defer server.Close()

	e := httpexpect.New(t, server.URL)
	e.GET("/").WithClient(rhttp.NewClient().StandardClient()).
		Expect().
		Status(http.StatusOK)
}

```
It's your call obviously, but talking as a user I see no reason to add code to support this "natively".
[...]

PR #89, opened 2021-02-13
WIP: Use RetryableHttp lib to add WithRetry functionality

nakulpathak3:
References https://github.com/gavv/httpexpect/issues/85

This is what I could gather so far but it has a failing test. If someone better at Go gets a chance to debug this, please do and I'll learn something from you :)


gavv:
Hi, thanks for the PR.

Unfortunately, after looking closely at the PR and `retryablehttp` source code, it appeared that we can't use this library.

In your implementation, if retries are enabled, you ignore `r.config.Client` and use `retryablehttp.NewClient()`. We can't do this: all requests we send should go through the client provided by user. On the other hand, we can't configure `retryablehttp` to use our own client, because it accepts only `http.Client` instances, while our client is of type `Client` interface.

There is another PR implementing the same feature (#90), but it can't be merged as well. Thus, I decided to implement it by myself. See https://github.com/gavv/httpexpect/issues/85#issuecomment-782745049

Thanks.

PR #90, opened 2021-02-16
Retry mothods added.

atenasadat:


gavv:
Hi, thanks for PR.

There are a few problems with this implementation and can't be merged as is:

* exported fields CheckForRetry and Backoff doesn't fit with the design of other APIs in this lib; we use fluent/chain builder pattern;

* I'd prefer to keep retry configuration declarative and simple, so I'd like to use a enum with retry policies instead of functions; if users need something more complicated than builtin policies, they can always provide custom http client;

* this implementation doesn't provide actual policies which I mentioned in the issue (retry only temporary network errors, retry temporary network errors and 5xx codes, retry all);

* tests are failing;  the new feature should be covered with tests too;

* also, it seems this implementation wont resend request body on retry;

* websocket requests should be retried as well.

There is another PR implementing the same feature (#89), but it can't be merged as well. Thus, I decided to implement it by myself. See https://github.com/gavv/httpexpect/issues/85#issuecomment-782745049

Thanks.

Issue #91, opened 2021-02-20
Add unit tests for redirect and retry support


gavv:
Redirect methods (WithRedirectPolicy, WithMaxRedirects) and retry methods (WithRetryPolicy, WithMaxRetries, WithRetryDelay) are already covered with end-to-end tests (e2e_redirect_test.go, e2e_retry_test.go), but unit tests are missing.

It would be nice to add unit tests for these methods to request_test.go. In unit tests, we can use `http.Client` with custom `Transport` set to our own implementation of `http.RoundTripper` interface. There we can check what requests are generated by httpexpect and produce fake responses to trigger redirects and retries.

gopalanvinay:
@gavv could I make a PR to add these tests?

gavv:
@gopalanvinay Sure, you're welcome.

gopalanvinay:
@gavv could I be added as a collaborator? Permission to push to the repo is currently denied to my account :)

gavv:
@123vivekr Hi! Yes, there is unmerged PR with unresolved issues, see #94. Feel free to pick up the task.

123vivekr:
I'm a beginner in Golang. I'll give this a try and submit a PR for review soon.

kaustubhmallik:
Hi @gavv is it still open? If so, I would like to pick it up. :)

123vivekr:
@kaustubhmallik Feel free to take it up.

gavv:
Hi, the issue is still open. See also discussion in #94.

neverbeenthisweeb:
@gavv hi, for testing the `WithRetryDelay` wdyt if we do `time.Sleep` in the unit test? then assert if the sleep time is as expected

or you might have concerns about this?

my concern is that by doing sleep, the unit test will be slower by some miliseconds (approximately, because we sleep)

gavv:
I think short sleeps, maybe below 10-20 ms per test, are OK.

neverbeenthisweeb:
got it, will keep the sleep around that

gavv:
Landed

PR #94, opened 2021-02-28
Add Unit Tests for Request Redirect and Retry Methods

gopalanvinay:
Added unit tests for Redirect methods (`withRedirectPolicy`, `withMaxRedirects`) and Retry methods (`withRetryPolicy`, `withMaxRetries`)

Resolves: https://github.com/gavv/httpexpect/issues/91

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -1594,3 +1594,63 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
 	req3.WithFileBytes("a", "a", []byte("a"))
 	req3.chain.assertFailed(t)
 }
+
+func TestRequestRedirectPolicy(t *testing.T) {
+	factory := DefaultRequestFactory{}
+
+	client := &mockClient{
+		err: errors.New("error"),
+	}
+
+	reporter := newMockReporter(t)
+
+	config := Config{
+		RequestFactory: factory,
+		Client:         client,
+		Reporter:       reporter,
+	}
+
+	req := NewRequest(config, "POST", "/path")
+	assert.Equal(t, RedirectPolicy(0), req.redirectPolicy)
+
+	req.chain.assertOK(t)
+	req.WithRedirectPolicy(FollowAllRedirects)
+	assert.Equal(t, RedirectPolicy(2), req.redirectPolicy)
+
+	req.chain.assertOK(t)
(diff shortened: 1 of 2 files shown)


gavv:
Thanks for the PR!

In these tests, you're checking that fields like `req.redirectPolicy`, `req.retryPolicy`, etc. are set correctly. While this is a valid check, it's not really helpful. What is really important is how these fields are used later to construct `http.Request` struct (not to be confused with `httpexpect.Request`) - as described in the issue.

When you call `Expect()`, it will look into redirection parameters and set `CheckRedirect` and `GetBody` fields of the `http.Request` (see `setupRedirects()`). It will also look into retry parameters and perform retries when request is failed (see `retryRequest()`).

To test this, we need to develop a mock http client (`mockClient` may help), pass it instead of a real client (see `Config`), create `httpexpect.Request`, configure some redirection/retry parameters, and call `Expect()`. After that we should inspect how `Expect()` interacted with our mock client - what were the fields of the generated `http.Request`, did it perform retries, etc.

If you search for `mockClient`, you can find examples of other tests using the same approach.

svaloumas:
Hello @gavv ! Apologies for interfering here. I was experimenting with writing those tests and noticed that in `setupRedirects` a type assertion on the `Client` interface takes place at the first line. 

```
func (r *Request) setupRedirects() {
	httpClient, _ := r.config.Client.(*http.Client)

	if httpClient == nil {
		if r.redirectPolicy != defaultRedirectPolicy {
			r.chain.fail("WithRedirectPolicy can be used only if Client is *http.Client")
			return
		}
        ....
```

Mocking the client while setting a redirect policy other than the default, would make the function return before setting the `CheckRedirect` and `GetBody` fields, as the `httpClient` would be `nil` after the type assertion. I may miss something here, any thoughts?

gavv:
@svaloumas oh, I see, you're right. We have to use http.Client here. However there is still a place where we can insert mock: we can set http.Client.Transport to our own implementation of http.RoundTripper interface. It is used to actually send request, and there we can check what requests are generated and produce fake responses.

gavv:
I'm closing this PR, but keeping the issue open (#91). Also I updated the issue text.

PR #102, opened 2021-08-20
Add environment in expect

zacscoding:
Add environment in `Expect` to manipulate test data.

This pr is a proposal for environment :)
For examples, i want to use a value from previous test like below

```go
func TestIT(t *testing.T) {
	e := httpexpect.New(t, "http://localhost:8080")

	t.Run("/users", func(t *testing.T) {
		obj := e.GET("/users").
			Expect().
			Status(http.StatusOK).JSON().Object()

		userID := obj.Path("$.users[1].id").String().Raw()
		e.Env.Put("user1.id", userID)
	})
	t.Run("/user/{userId}", func(t *testing.T) {
                // use a user's id from above tests.
		userID, err := e.Env.GetString("user1.id")
		assert.NoError(t, err)
		e.GET("/user/{userId}").
			WithPath("userId", userID)
			Expect().
			Status(http.StatusOK)
	})
}
```

If u agree with this concept, i will implement more such as int slice types with tests.
Thanks for providing a good test library!


gavv:
Hi! Thanks a lot for proposal and PR and sorry for delay.

The proposal looks good! There are a couple of gotchas with current implementation:

* instead of returning errors, I'd prefer to design this piece in-line with the rest of the library, and report failures to via `chain`; if value does not exist, Env would report failure and return zero value

* instead of automatic conversion performed by spf13/cast, I'd prefer more strict type checks that don't perform implicit conversions between types; if value has different type, Env would report failure and return zero value

I've cherry-picked your commit to master (13b7a077f3e2e3eba09474ccd74ce0e3b1ded0ef) and added a follow-up commit with changes I described above + tests + documentation (7c17246510c15103d47e4b48c51bda673cd9a6b1).

Also added section to README with example you've provided: https://github.com/gavv/httpexpect#shared-environment

Let me know if you have more suggestions or improvements.

PR #108, opened 2022-04-07
Upgrade valyala/fasthttp to v1.35.0

bajalnyt:
The package github.com/valyala/fasthttp before 1.34.0 is vulnerable to Directory Traversal via the ServeFile function, due to improper sanitization. It is possible to be exploited by using a backslash %5c character in the path. 

Note: This security issue impacts Windows users only.


gavv:
Thanks, fixed in #106.

PR #110, opened 2022-05-10
Add a key to the fail message report

snowlyg:
#### What am I using?

- Web framework: Gin
- Ide: vscode

#### Go version 
go 1.16+

#### Repositorie
[https://github.com/snowlyg/iris-admin](https://github.com/snowlyg/iris-admin)

When I use [https://github.com/snowlyg/httptest](https://github.com/snowlyg/httptest) to test  HTTP API and the HTTP response object or array data, I can't know the fail is which key.

```
func (rks ResponseKeys) Test(object *httpexpect.Object) {
	for _, rk := range rks {
		zap_server.ZAPLOG.Info(rk.Key)
		// to sign the test data key name
		if rk.Value == nil {
			continue
		}
		switch reflect.TypeOf(rk.Value).String() {
		case "string":
			if strings.ToLower(rk.Type) == "notempty" {
				object.Value(rk.Key).String().NotEmpty()
			} else {
				object.Value(rk.Key).String().Equal(rk.Value.(string))
			}
		case "float64":
			if strings.ToLower(rk.Type) == "ge" {
				object.Value(rk.Key).Number().Ge(rk.Value.(float64))
			} else {
				object.Value(rk.Key).Number().Equal(rk.Value.(float64))
			}
		case "uint":
			if strings.ToLower(rk.Type) == "ge" {
				object.Value(rk.Key).Number().Ge(rk.Value.(uint))
			} else {
				object.Value(rk.Key).Number().Equal(rk.Value.(uint))
			}
		case "int":
			if strings.ToLower(rk.Type) == "ge" {
				object.Value(rk.Key).Number().Ge(rk.Value.(int))
			} else {
				object.Value(rk.Key).Number().Equal(rk.Value.(int))
			}
		case "[]base.ResponseKeys":
			if strings.ToLower(rk.Type) == "empty" {
				object.Value(rk.Key).Array().Empty()
				continue
			}
```
image

I will add a key to the fail message report like below :

image #110

### DIFF
--- a/array.go
+++ b/array.go
@@ -9,6 +9,7 @@ import (
 type Array struct {
 	chain chain
 	value []interface{}
+	key   string
 }
 
 // NewArray returns a new Array given a reporter used to report failures
@@ -26,7 +27,7 @@ func NewArray(reporter Reporter, value []interface{}) *Array {
 	} else {
 		value, _ = canonArray(&chain, value)
 	}
-	return &Array{chain, value}
+	return &Array{chain, value, ""}
 }
 
 // Raw returns underlying value attached to Array.
@@ -56,7 +57,7 @@ func (a *Array) Schema(schema interface{}) *Array {
 //  array := NewArray(t, []interface{}{1, 2, 3})
 //  array.Length().Equal(3)
 func (a *Array) Length() *Number {
-	return &Number{a.chain, float64(len(a.value))}
+	return &Number{a.chain, float64(len(a.value)), a.key}
 }
 
 // Element returns a new Value object that may be used to inspect array element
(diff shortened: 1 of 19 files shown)


gavv:
Hi,

Thanks for suggestion and PR!

The feature itself looks good, but unfortunately the implementation in this PR conflicts a lot with changes needed for #69 (see #98, #150).

I recently merged #150 and among other things it covers the use case from your PR. With new changes, httpexpect now tracks path of every assertion, including map keys, and prints this path in failure report.

Here is an example:

```
        expected: value is map
        
        assertion:
          Request("GET").Expect().JSON().Object().Value("foo").Object()
        
        actual value:
          [
            "bar",
            123,
            false,
            null
          ]
```

So I'm going to close this PR, but please let me know if I'm missing something and thanks for your work!

PR #114, opened 2022-05-26
add a new func Number to String

chzhuo:


gavv:
Thanks!

gavv:
Tagged v2.4.0 with this PR included.

Issue #103, opened 2021-10-30
Some tests fail in os windows


EstebanJesus:
When cloning the project on a windows OS and running the tests I get some failures, it is related to the creation of temporary files.

EstebanJesus:
Great, with the version you created it works perfect for me.

clobrano:
@gavv, what am I supposed to do with this?

gavv:
@clobrano the fix is merged. Does this problem still reproduce for you on fresh master?

clobrano:
unfortunately I cannot re-test this anymore

PR #115, opened 2022-06-14
tests: fix expected filename in TestRequestBodyMultipartFile

clobrano:
currently TestRequestBodyMultipartFile expect the filename given by

func (*multipartPart) FileName() string

to be a fullpath in one of the test cases, however the documentation
states that:

> the filename is passed through filepath.Base (which is platform
> dependent) before being returned.

then the test fails.

This change fix the expected filename for this test, using the same
filepath.Base function.

Fixes: #103

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -9,6 +9,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


EstebanJesus (approved):

EstebanJesus:
Hi @aries-zhang, I really don't know what the strategy is like, but I don't see a problem with it.

clobrano:
Hey, let me know if there is more I can do to improve this PR.

gavv:
Follow-up commit a54cf58c872b620d00b71cc961822765ae316d64 to fix the build completely

PR #116, opened 2022-06-16
Fix a test error in go1.17: expected multipart filename should not include its directory path

aries-zhang:
This repo depends on a version of [valyala/fasthttp](https://github.com/valyala/fasthttp) that contains a critical security vulnerability: https://www.cve.org/CVERecord?id=CVE-2022-21221.

`dependabot` has a bump of the version but the test was failing due to a change in [multipart filename in go 1.17](https://cs.opensource.google/go/go/+/dev.boringcrypto.go1.17:src/mime/multipart/multipart.go;l=84).

This PR fixes this test failure and hope the vulnerability can be fixed as well.

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -9,6 +9,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


gavv:
Thanks, both #115 and #106 are merged now.

PR #118, opened 2022-07-16
fix abrupt context cancel when response body is large

ratanphayade:
While reading the response body context gets cancelled die to termination connection termination. 

In here we are wrapping the the response body along with `context.CancelFunc` this will help us read the body regardless of the response size


gavv:
Fixed in 18182d49157016129c6f14806f37d85c75a50016

PR #172, opened 2022-11-24
Add unit tests for redirect support

neverbeenthisweeb:
Issue: https://github.com/gavv/httpexpect/issues/91

Scope:
- Add round tripper implementation that can have various HTTP response in unit tests
- Add unit tests for redirects

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -1610,10 +1610,7 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has no body", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.Equal(t, http.NoBody, r.Body)
-					assert.Nil(t, r.GetBody)
-
-					return true
+					return assert.Equal(t, http.NoBody, r.Body)
 				})
 
 			client := &http.Client{
@@ -1636,6 +1633,9 @@ func TestRequestRedirect(t *testing.T) {
 				Equal("/redirect")
 			resp.chain.assertOK(t)
 
+			// Should set GetBody
+			assert.Nil(t, req.httpReq.GetBody)
+
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
 			assert.NotNil(t, httpClient.CheckRedirect)
@@ -1648,9 +1648,9 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has body", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.NotNil(t, r.Body)
-					assert.NotEqual(t, http.NoBody, r.Body)
-					assert.Nil(t, r.GetBody)
+					b, err := ioutil.ReadAll(r.Body)
+					assert.NoError(t, err)
+					assert.Equal(t, "test body", string(b))
 
 					return true
 				})
@@ -1676,6 +1676,9 @@ func TestRequestRedirect(t *testing.T) {
 				Equal("/redirect")
 			resp.chain.assertOK(t)
 
+			// Should set GetBody
+			assert.Nil(t, req.httpReq.GetBody)
+
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
 			assert.NotNil(t, httpClient.CheckRedirect)
@@ -1690,19 +1693,7 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has no body", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.Equal(t, http.NoBody, r.Body)
-
-					if r.URL.String() == "/url" {
-						rc, err := r.GetBody()
-						assert.NoError(t, err)
-						assert.Equal(t, http.NoBody, rc)
-					} else if r.URL.String() == "/redirect" {
-						assert.Nil(t, r.GetBody)
-					} else {
-						t.Fatalf("invalid request URL")
-					}
-
-					return true
+					return assert.Equal(t, http.NoBody, r.Body)
 				}).WithMaxRedirect(1)
 
 			client := &http.Client{
@@ -1723,6 +1714,11 @@ func TestRequestRedirect(t *testing.T) {
 			resp := req.Expect().
 				Status(http.StatusOK)
 			resp.chain.assertOK(t)
+
+			// Should set GetBody
+			gb, err := req.httpReq.GetBody()
+			assert.NoError(t, err)
+			assert.Equal(t, http.NoBody, gb)
 
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
@@ -1741,19 +1737,7 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has no body and too many redirects", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.Equal(t, http.NoBody, r.Body)
-
-					if r.URL.String() == "/url" {
-						rc, err := r.GetBody()
-						assert.NoError(t, err)
-						assert.Equal(t, http.NoBody, rc)
-					} else if r.URL.String() == "/redirect" {
-						assert.Nil(t, r.GetBody)
-					} else {
-						t.Fatalf("invalid request URL")
-					}
-
-					return true
+					return assert.Equal(t, http.NoBody, r.Body)
 				})
 
 			client := &http.Client{
@@ -1774,6 +1758,11 @@ func TestRequestRedirect(t *testing.T) {
 			resp := req.Expect()
 			resp.chain.assertFailed(t)
 
+			// Should set GetBody
+			gb, err := req.httpReq.GetBody()
+			assert.NoError(t, err)
+			assert.Equal(t, http.NoBody, gb)
+
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
 			assert.NotNil(t, httpClient.CheckRedirect)
@@ -1791,21 +1780,9 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has body", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.NotNil(t, r.Body)
-
-					if r.URL.String() == "/url" {
-						rc1, err := r.Body.(*bodyWrapper).GetBody()
-						assert.NoError(t, err)
-
-						rc2, err := r.GetBody()
-						assert.NoError(t, err)
-
-						assert.Equal(t, rc1, rc2)
-					} else if r.URL.String() == "/redirect" {
-						assert.Nil(t, r.GetBody)
-					} else {
-						t.Fatalf("invalid request URL")
-					}
+					b, err := ioutil.ReadAll(r.Body)
+					assert.NoError(t, err)
+					assert.Equal(t, "test body", string(b))
 
 					return true
 				}).WithMaxRedirect(1)
@@ -1830,6 +1807,13 @@ func TestRequestRedirect(t *testing.T) {
 				Status(http.StatusOK)
 			resp.chain.assertOK(t)
 
+			// Should set GetBody
+			gb, err := req.httpReq.GetBody()
+			assert.NoError(t, err)
+			b, err := ioutil.ReadAll(gb)
+			assert.NoError(t, err)
+			assert.Equal(t, "test body", string(b))
+
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
 			assert.NotNil(t, httpClient.CheckRedirect)
@@ -1847,21 +1831,9 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has body and too many redirects", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.NotNil(t, r.Body)
-
-					if r.URL.String() == "/url" {
-						rc1, err := r.Body.(*bodyWrapper).GetBody()
-						assert.NoError(t, err)
-
-						rc2, err := r.GetBody()
-						assert.NoError(t, err)
-
-						assert.Equal(t, rc1, rc2)
-					} else if r.URL.String() == "/redirect" {
-						assert.Nil(t, r.GetBody)
-					} else {
-						t.Fatalf("invalid request URL")
-					}
+					b, err := ioutil.ReadAll(r.Body)
+					assert.NoError(t, err)
+					assert.Equal(t, "test body", string(b))
 
 					return true
 				})
@@ -1884,6 +1856,13 @@ func TestRequestRedirect(t *testing.T) {
 			// Should error
 			resp := req.Expect()
 			resp.chain.assertFailed(t)
+
+			// Should set GetBody
+			gb, err := req.httpReq.GetBody()
+			assert.NoError(t, err)
+			b, err := ioutil.ReadAll(gb)
+			assert.NoError(t, err)
+			assert.Equal(t, "test body", string(b))
 
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
@@ -1904,10 +1883,7 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has no body", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.Contains(t, []interface{}{nil, http.NoBody}, r.Body)
-					assert.Nil(t, r.GetBody)
-
-					return true
+					return assert.Contains(t, []interface{}{nil, http.NoBody}, r.Body)
 				}).WithMaxRedirect(1)
 
 			client := &http.Client{
@@ -1929,6 +1905,9 @@ func TestRequestRedirect(t *testing.T) {
 				Status(http.StatusOK)
 			resp.chain.assertOK(t)
 
+			// Should set GetBody
+			assert.Nil(t, req.httpReq.GetBody)
+
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
 			assert.NotNil(t, httpClient.CheckRedirect)
@@ -1946,19 +1925,7 @@ func TestRequestRedirect(t *testing.T) {
 		t.Run("request has no body and too many redirects", func(t *testing.T) {
 			tp := newMockTransportRedirect(t).
 				WithAssertFn(func(r *http.Request) bool {
-					assert.Equal(t, http.NoBody, r.Body)
-
-					if r.URL.String() == "/url" {
-						rc, err := r.GetBody()
-						assert.NoError(t, err)
-						assert.Equal(t, http.NoBody, rc)
-					} else if r.URL.String() == "/redirect" {
-						assert.Nil(t, r.GetBody)
-					} else {
-						t.Fatalf("invalid request URL")
-					}
-
-					return true
+					return assert.Contains(t, []interface{}{nil, http.NoBody}, r.Body)
 				})
 
 			client := &http.Client{
@@ -1971,13 +1938,16 @@ func TestRequestRedirect(t *testing.T) {
 			}
 
 			req := NewRequest(config, http.MethodPut, "/url").
-				WithRedirectPolicy(FollowAllRedirects).
+				WithRedirectPolicy(FollowRedirectsWithoutBody).
 				WithMaxRedirects(1)
 			req.chain.assertOK(t)
 
 			// Should error
 			resp := req.Expect()
 			resp.chain.assertFailed(t)
+
+			// Should set GetBody
+			assert.Nil(t, req.httpReq.GetBody)
 
 			// Should set CheckRedirect
 			httpClient, _ := req.config.Client.(*http.Client)
@@ -1997,8 +1967,9 @@ func TestRequestRedirect(t *testing.T) {
 			func(t *testing.T) {
 				tp := newMockTransportRedirect(t).
 					WithAssertFn(func(r *http.Request) bool {
-						assert.NotNil(t, r.Body)
-						assert.Nil(t, r.GetBody)
+						b, err := ioutil.ReadAll(r.Body)
+						assert.NoError(t, err)
+						assert.Equal(t, "test body", string(b))
 
 						return true
 					}).WithMaxRedirect(1)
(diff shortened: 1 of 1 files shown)


neverbeenthisweeb:
hi @gavv would you mind to check on this draft?

here, i start by testng the `.WithRedirectPolicy(DontFollowRedirects)`

i mock the `RoundTripper` and assert for `origReq.GetBody == nil` in it because that is the only operation that happens when we have `.WithRedirectPolicy(DontFollowRedirects)` CMIIW (see https://github.com/gavv/httpexpect/blob/master/request.go#L2025)

if the assertion OK, i return HTTP 200

if the assertion failed, i return HTTP 400

is there something i am missing out or to improve? thank you!

neverbeenthisweeb:
i am still unsure whether should i spawn a httptest server (https://go.dev/src/net/http/httptest/example_test.go) or not to simulate the redirect/retry

because none of the test in `request_test.go` do spawn a httptest server

gavv:
Hi, thanks for PR!

The approach looks good. In addition to checking `GetBody == nil`, for DontFollowRedirects we should also check that CheckRedirect field of http.Client is not nil and if we call it, it immediately returns ErrUseLastResponse (see [here](https://github.com/gavv/httpexpect/blob/master/request.go#L2000)). This is what forces http.Client to disable redirects.

Also, I think instead of StatusOK, it's better to return StatusPermanentRedirect to force http.Client to apply redirect policy. And check that it didn't follow redirect (i.e. didn't make one more call to transport, it was only one call).

Some small improvements that we could make:

* request factory is not needed, it will be created automatically
* we can pass testing.T to mockTransportWithDontFollowRedirects, and it can use t.Fatalf() to report failed assertion and interrupt test, instead of encoding assertion result inside http response; it's cleaner and more flexible

> i am still unsure whether should i spawn a httptest server (https://go.dev/src/net/http/httptest/example_test.go) or not to simulate the redirect/retry

No need for httptest.Server because we replace real transport with a mock that does not do any network operations. All we need is to return correct response from roundtripper.

BTW one concern in this PR is that we don't set all response fields and probably it will look invalid to http.Client. Instead of creating response manually, I suggest to use httputil.ResponseRecorder. It should fill all necessary fields.

neverbeenthisweeb:
> BTW one concern in this PR is that we don't set all response fields and probably it will look invalid to http.Client. Instead of creating response manually, I suggest to use httputil.ResponseRecorder. It should fill all necessary fields.

@gavv sorry, but did you mean [httptest.ResponseRecorder](https://pkg.go.dev/net/http/httptest#ResponseRecorder)?

neverbeenthisweeb:
hi @gavv i have updated this PR based on the previous comments

i added test for cases of redirect:

- dont follow redirects
- follow all redirects

would you mind to take a look? thanks!

request_test.go
```diff
@@ -1601,3 +1603,85 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+		client := &http.Client{
+			Transport: mockTransportWithNoRedirect{t: t},
+		}
+
+		config := Config{
+			Client:   client,
+			Reporter: reporter,
+		}
+
+		req := NewRequest(config, "METHOD", "url").
```
gavv, line 1620: Since we use real http.Client, let's also use real method, e.g. GET or POST.

request_test.go
```diff
@@ -1601,3 +1603,85 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+			WithHeader("Location", "/redirect").
```
gavv, line 1621: Is this header really needed? This header is usually present in responses, not in requests. On the other hand, I suggest to set Location in response, because http.Client probably expects it.

Same for the second test.

request_test.go
```diff
@@ -1601,3 +1603,85 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+			WithRedirectPolicy(DontFollowRedirects)
+		req.chain.assertOK(t)
+
+		resp := req.Expect().
+			Status(http.StatusPermanentRedirect)
+		resp.chain.assertOK(t)
+
+		httpClient, _ := req.config.Client.(*http.Client)
+		assert.Equal(t, http.ErrUseLastResponse, httpClient.CheckRedirect(req.httpReq, nil))
```
gavv, line 1630: Nitpick: let's also check that httpClient.CheckRedirect is not nil, before calling it, to make clear our intentions. We could also add a comment, like "Expect() should set CheckRedirect".

Same for the second test.

request_test.go
```diff
@@ -1601,3 +1603,85 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+
+		assert.Nil(t, req.httpReq.GetBody)
+	})
```
gavv, line 1633: Nitpick: let's also check that tripCount has expected value (not zero), to be sure that our round tripper was really called.

Same for the second test.

request_test.go
```diff
@@ -1601,3 +1603,85 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+
+		config := Config{
+			Client:   client,
+			Reporter: reporter,
+		}
+
+		req := NewRequest(config, "METHOD", "url").
+			WithHeader("Location", "/redirect").
+			WithRedirectPolicy(FollowAllRedirects).
+			WithMaxRedirects(0).
```
gavv, line 1648: It would be nice to add a test for non-zero max-redirects too.

request_test.go
```diff
@@ -1601,3 +1603,85 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+) {
+	mt.tripCount++
+
+	assert.Equal(mt.t, mt.tripCount, uint8(1))
+
+	res := httptest.NewRecorder()
+	res.Result().StatusCode = http.StatusPermanentRedirect
+
+	return res.Result(), nil
+}
```
gavv, line 1687: In the second test, let's also check here that origReq.Body is set and we can read from it and it gives us correct value ("test body").

gavv:
Thanks for update, see some comments above.

> @gavv sorry, but did you mean [httptest.ResponseRecorder](https://pkg.go.dev/net/http/httptest#ResponseRecorder)?

Yep, sorry.

neverbeenthisweeb:
hi thank you for the feedback, will do the update

neverbeenthisweeb:
@gavv but i have a question

it looks like we can access, then assert, the `http.Request` in 2 ways:

(1) using `httpReq` in `httpexpect.Request`

```go
req := NewRequest(config, "METHOD", "url")
req.Expect()
body := req.httpReq.Body // Can proceed to asserting body
```

(2) or inside the `RoundTrip`

```
func (mt mockTransportWithNoRedirect) RoundTrip(origReq *http.Request) (
	*http.Response, error,
) {
    body := origReq.Body // Can proceed to asserting body
}
```

may i know which one do your prefer and why?

i think the RoundTrip one is more preferable, because it really assert what happen when we are doing the HTTP call

request_test.go
neverbeenthisweeb: previously i got error when not setting the `Location` header in request

but let me double check on that

neverbeenthisweeb: i think i read the documentation wrong

yep, you're right, the Location header should be in response not request

```
loc := resp.Header.Get("Location")
			if loc == "" {
				// While most 3xx responses include a Location, it is not
				// required and 3xx responses without a Location have been
				// observed in the wild. See issues #17773 and #49281.
				return resp, nil
			}
```

https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/http/client.go;l=631;drc=d5de62df152baf4de6e9fe81933319b86fd95ae4

gavv:
> i think the RoundTrip one is more preferable, because it really assert what happen when we are doing the HTTP call

Yes, I also think so. http.Client creates a new http.Request for every RoundTrip call, and sets its Body to new value. And that's what we want to check.

neverbeenthisweeb:
got it :+1:

request_test.go
```diff
@@ -1601,3 +1603,171 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+		assert.Equal(t, nil, httpClient.CheckRedirect(req.httpReq, make([]*http.Request, 0)))
+		assert.EqualError(t,
+			httpClient.CheckRedirect(req.httpReq, make([]*http.Request, 1)),
+			fmt.Sprintf("stopped after %d redirects", req.maxRedirects),
+		)
+	})
+
+	t.Run("follow all redirect policy with max redirects 1", func(t *testing.T) {
+		trans := newMockTransportRedirect(t, func(r *http.Request) bool {
+			rc, err := r.GetBody()
```
neverbeenthisweeb, line 1687: hi @gavv i wanna ask something

i still cant figure out why `r.GetBody` is nil in the second redirect (this test `follow all redirect policy with max redirects 1` do 2 redirects)

do you might have any clue?

i think, in the HTTP 308 [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308#:~:text=The%20request%20method%20and%20the%20body%20will%20not%20be%20altered) it is written that the payload should be _forwarded_ to the next redirect

neverbeenthisweeb: here is my log when running the test:

```
=== RUN   TestRequestRedirect
=== RUN   TestRequestRedirect/dont_follow_redirects_policy
=== RUN   TestRequestRedirect/follow_all_redirect_policy_with_max_redirects_0
=== CONT  TestRequestRedirect
    mocks_test.go:84: Fail: 
        failed to send http request
          Post "/redirect": stopped after 0 redirects
        
        assertion:
          Request().Expect()
=== RUN   TestRequestRedirect/follow_all_redirect_policy_with_max_redirects_1
--- FAIL: TestRequestRedirect (0.00s)
    --- PASS: TestRequestRedirect/dont_follow_redirects_policy (0.00s)
    --- PASS: TestRequestRedirect/follow_all_redirect_policy_with_max_redirects_0 (0.00s)
    --- FAIL: TestRequestRedirect/follow_all_redirect_policy_with_max_redirects_1 (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9c6be5]

goroutine 18 [running]:
testing.tRunner.func1.2({0xa4ebe0, 0xfde980})
	/home/rafi/Documents/personal/programming/go/versions/go/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
	/home/rafi/Documents/personal/programming/go/versions/go/src/testing/testing.go:1392 +0x39f
panic({0xa4ebe0, 0xfde980})
	/home/rafi/Documents/personal/programming/go/versions/go/src/runtime/panic.go:838 +0x207
github.com/gavv/httpexpect/v2.TestRequestRedirect.func3.1(0xbe22c0?)
	/home/rafi/Documents/personal/foss/github/httpexpect/request_test.go:1687 +0x25
github.com/gavv/httpexpect/v2.(*mockTransportRedirect).RoundTrip(0xc00028a1e0, 0xbe14c0?)
	/home/rafi/Documents/personal/foss/github/httpexpect/request_test.go:1765 +0x98
net/http.send(0xc000292300, {0xbe14c0, 0xc00028a1e0}, {0xac6ee0?, 0x49a301?, 0x0?})
	/home/rafi/Documents/personal/programming/go/versions/go/src/net/http/client.go:252 +0x5d8
net/http.(*Client).send(0xc000282c60, 0xc000292300, {0x7fd16c7ba460?, 0xc0002844d0?, 0x0?})
	/home/rafi/Documents/personal/programming/go/versions/go/src/net/http/client.go:176 +0x9b
net/http.(*Client).do(0xc000282c60, 0xc000292300)
	/home/rafi/Documents/personal/programming/go/versions/go/src/net/http/client.go:725 +0x8f5
net/http.(*Client).Do(0xc000058810?, 0x48c397?)
	/home/rafi/Documents/personal/programming/go/versions/go/src/net/http/client.go:593 +0x19
github.com/gavv/httpexpect/v2.(*Request).sendRequest.func1()
	/home/rafi/Documents/personal/foss/github/httpexpect/request.go:1797 +0x2d
github.com/gavv/httpexpect/v2.(*Request).retryRequest(0xc0002d2000, 0xc0001379b0)
	/home/rafi/Documents/personal/foss/github/httpexpect/request.go:1892 +0x4f9
github.com/gavv/httpexpect/v2.(*Request).sendRequest(0xc0002d2000)
	/home/rafi/Documents/personal/foss/github/httpexpect/request.go:1796 +0x65
github.com/gavv/httpexpect/v2.(*Request).roundTrip(0xc0002d2000)
	/home/rafi/Documents/personal/foss/github/httpexpect/request.go:1703 +0x105
github.com/gavv/httpexpect/v2.(*Request).Expect(0xc0002d2000)
	/home/rafi/Documents/personal/foss/github/httpexpect/request.go:1664 +0x166
github.com/gavv/httpexpect/v2.TestRequestRedirect.func3(0xc0002d0000)
	/home/rafi/Documents/personal/foss/github/httpexpect/request_test.go:1712 +0x1fa
testing.tRunner(0xc0002d0000, 0xc0002843e0)
	/home/rafi/Documents/personal/programming/go/versions/go/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	/home/rafi/Documents/personal/programming/go/versions/go/src/testing/testing.go:1486 +0x35f
FAIL	github.com/gavv/httpexpect/v2	0.010s
FAIL
make: *** [Makefile:37: ut] Error 1
```

many thanks!

neverbeenthisweeb: my Makefile:

```
ut:
	clear && \
		go test -v -run TestRequestRedirect github.com/gavv/httpexpect/v2
```

neverbeenthisweeb: after some debugging

CMIIW it looks like `.GetBody` is not set for server request, such as redirect

as written in the code ([src/net/http/client.go](https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/http/client.go;l=663;drc=d5de62df152baf4de6e9fe81933319b86fd95ae4))

and as written in the docs ([net/http/Request.GetBody](https://pkg.go.dev/net/http#:~:text=//%20For%20server%20requests%2C%20it%20is%20unused.))

neverbeenthisweeb: i have updated the assertion to use `TeeReader` instead to assert the `r.Body`. wdyt @gavv ? thanks!

https://github.com/gavv/httpexpect/pull/172/commits/7e90c0c3b46688963339bfbb607351efe2d52865

gavv: Yes, exactly. The convention is the following:

* we create http.Request that has both Body and GetBody
* we pass it to http.Client
* http.Client passes request to RoundTripper as is
* if roundtripper returns redirection response, and CheckRedirect reports that redirect should be followed, http.Client creates a new request
* in new request, http.Client doesn't set GetBody; instead, it calls GetBody on original request and sets Body on new request
* the loop with redirects is repeated until CheckRedirect tells to stop

So yes, in round tripper we should check Body, not GetBody. GetBody is only for http.Client, not for roundtripper.

Relates lines:

* https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/http/client.go;l=627-628
* https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/http/client.go;l=652-668

gavv: I think we should change all tests to use r.Body, and ignore r.GetBody in RoundTripper (it first call it will be non-nil, in other calls it will be nil, but that's not really important and can be considered an implementation detail).

On the other hand, I think we should still do a check that .Expect() sets GetBody in the *original* request (req.httpReq.GetBody). We can also call that GetBody and check that the body is correct.

gavv: BTW I think there is no real need for TeeReader. I think http.Client expects that RoundTripper will read the body and close the reader. And each round trip, http.Client will set Body to fresh reader returned by GetBody.

gavv: Ideally we should also add a test for success case with FollowAllRedirects: when a few redirects are followed, limit is NOT reached, Expect() succeeds, and response has status 200. For that, we'll need to teach RoundTripper to return 200 instead of redirect eventually.

gavv: Also ideally we should cover FollowRedirectsWithoutBody (cases when request has and does not have body).

gavv:
> Add unit tests for redirect and retry support

If it would be more convenient to you, we can limit this PR to only redirect tests and merge it when they're done, and move retry tests to separate PR - to make this shorter. I'm OK with any approach, up to you.

request_test.go
gavv: Regarding understanding of GetBody. You can think about it as an optional thing needed solely by http.Client, solely for redirect support.

Normally request doesn't have GetBody (both on client and server). And even if it has one, normally it would be ignored.

However in one special case, when http.Client sent request, and received response with redirect status, and decided to follow redirect, and when request has body (e.g. it's POST or PUT), then http.Client uses GetBody to "rollback" Body to its original value. Without GetBody it wont be possible because Body's contract is that it's read once and then closed, without any possibility to "rollback".

And in this special case http.Client performs this "rollback" magic with Body and then passes to RoundTripper a normal request with Body and without GetBody and RoundTripper works as usual.

neverbeenthisweeb: > Ideally we should also add a test for success case with FollowAllRedirects: when a few redirects are followed, limit is NOT reached, Expect() succeeds, and response has status 200. For that, we'll need to teach RoundTripper to return 200 instead of redirect eventually.

@gavv  does it mean that our round tripper implementation (e.g. `newMockTransportRedirect.RoundTrip`) should be able to do `RoundTrip` with different HTTP response?

something like

initial request --> RoundTrip returns HTTP 308 --> redirect --> RoundTrip returns HTTP 200 --> response OK

gavv: Yep, ideally a few redirects and then HTTP 200.

A simple way to do it is to add a redirect limit, until counter reaches this limit, we'll return redirect, and after limit is reached we return 200.

gavv: This is how it normally works in practice. http.Client follows a few redirects and finally gets desired response. So it makes sense to cover it in unit tests.

neverbeenthisweeb: i see, okok

i will limit the scope of this PR into:

- add round tripper implementation that can have various HTTP response in unit tests
- add unit tests for redirects

regarding retry mechanism will not be included in this PR

thank you for the feeback, will do the update

gavv: Great, thank you.

neverbeenthisweeb:
hi @gavv , i have updated the unit tests to cover these cases:

```
--- PASS: TestRequestRedirect (0.00s)
    --- PASS: TestRequestRedirect/dont_follow_redirects_policy (0.00s)
        --- PASS: TestRequestRedirect/dont_follow_redirects_policy/request_has_no_body (0.00s)
        --- PASS: TestRequestRedirect/dont_follow_redirects_policy/request_has_body (0.00s)
    --- PASS: TestRequestRedirect/follow_all_redirects_policy (0.00s)
        --- PASS: TestRequestRedirect/follow_all_redirects_policy/request_has_no_body (0.00s)
        --- PASS: TestRequestRedirect/follow_all_redirects_policy/request_has_no_body_and_too_many_redirects (0.00s)
        --- PASS: TestRequestRedirect/follow_all_redirects_policy/request_has_body (0.00s)
        --- PASS: TestRequestRedirect/follow_all_redirects_policy/request_has_body_and_too_many_redirects (0.00s)
    --- PASS: TestRequestRedirect/follow_redirects_without_body_policy (0.00s)
        --- PASS: TestRequestRedirect/follow_redirects_without_body_policy/request_has_no_body (0.00s)
        --- PASS: TestRequestRedirect/follow_redirects_without_body_policy/request_has_no_body_and_too_many_redirects (0.00s)
        --- PASS: TestRequestRedirect/follow_redirects_without_body_policy/request_has_body_and_redirected_with_status_permanent_redirect (0.00s)
        --- PASS: TestRequestRedirect/follow_redirects_without_body_policy/request_has_body_and_redirected_with_status_moved_permanently (0.00s)
        --- PASS: TestRequestRedirect/follow_redirects_without_body_policy/request_has_body____and_redirected_with_status_moved_permanently____and_too_many_redirects (0.00s)
```

i exclude unit test for `defaultRedirectPolicy` because it is the same as `FollowRedirectsWithoutBody`. As written in https://github.com/gavv/httpexpect/blob/master/request.go#L421

thank you

request_test.go
```diff
@@ -1601,3 +1602,634 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+
+		t.Run("request has body", func(t *testing.T) {
+			tp := newMockTransportRedirect(t).
+				WithAssertFn(func(r *http.Request) bool {
+					assert.NotNil(t, r.Body)
+					assert.NotEqual(t, http.NoBody, r.Body)
+					assert.Nil(t, r.GetBody)
+
+					return true
+				})
```
gavv, line 1656: Let's also read and check body contents here.

request_test.go
```diff
@@ -1601,3 +1602,634 @@ func TestRequestErrorConflictMultipart(t *testing.T) {
...
+						assert.NoError(t, err)
+
+						assert.Equal(t, rc1, rc2)
+					} else if r.URL.String() == "/redirect" {
+						assert.Nil(t, r.GetBody)
+					} else {
+						t.Fatalf("invalid request URL")
+					}
+
+					return true
```
gavv, line 1810: Inside this WithAssertFn, we can just read r.Body and check that it's equal to "test body".

It's better not to assume that inside RoundTrip r.Body is bodyWrapper - because theoretically nothing prevents http.Client to wrap or replace body reader with something else when it passes body to RoundTripper.

gavv: Same for similar tests below.

gavv (commented):
Thanks for update, great tests.

gavv:
Also I suggest to remove GetBody checks inside RoundTripper (i.e. inside WithAssertFn).

In my understanding, GetBody is not for RoundTripper, it for http.Client. And it's not guaranteed that inside particular RoundTrip call GetBody will be nil or non-nil, I'd rather say it's an implementation detail of http.Client and may change. E.g. hypothetically client is allowed to set GetBody to nil even for first request, why not (although currently it does not touch first request).

So, inside RoundTripper I suggest to check only Body, and outside of RoundTripper, after Expect() returns, also check GetBody of req.httpReq.

gavv:
BTW can we mark PR as ready for review, i.e. remove draft status?

neverbeenthisweeb:
> Also I suggest to remove GetBody checks inside RoundTripper (i.e. inside WithAssertFn).
> 
> In my understanding, GetBody is not for RoundTripper, it for http.Client. And it's not guaranteed that inside particular RoundTrip call GetBody will be nil or non-nil, I'd rather say it's an implementation detail of http.Client and may change. E.g. hypothetically client is allowed to set GetBody to nil even for first request, why not (although currently it does not touch first request).
> 
> So, inside RoundTripper I suggest to check only Body, and outside of RoundTripper, after Expect() returns, also check GetBody of req.httpReq.

ok, i've done the update in https://github.com/gavv/httpexpect/pull/172/commits/f8973de0029dcf05709273c14a39906fb84d9cf9

please check

thank you

gavv:
Thanks a lot, LGTM

gavv:
BTW if you're interested in http/network related features, there are two interesting ideas - adding [support for GraphQL](https://github.com/gavv/httpexpect/discussions/147) and [for EventSource](https://github.com/gavv/httpexpect/discussions/145) (aka EventStream/SSE). We don't have help wanted issues for them right because I didn't find time to think about design and write issue text - but if you're interested you're welcome.

neverbeenthisweeb:
thank you @gavv 

im quite interested

but maybe i will start by taking unit test related issues first to get familiar with this project

🚀

Issue #157, opened 2022-11-23
Implement Array.ContainsAny


gavv:
We already have `Array.Contains` and `Array.ContainsOnly`.

It would be handy to have `ContainsAny`, which succeeds if array contains any of the specified elements.

We also need the counterpart `NotContainsAny`.

Assertions should have documentation comment with example and be covered by unit tests.

ni3k:
Hi @gavv can I work on this issue? I think I could come up with a PR today

gavv:
@ni3k Great, you're welcome, thanks!

gavv:
Landed.

PR #173, opened 2022-11-24
Add Array.ContainsAny and Array.NotContainsAny

ni3k:
This pull request addresses issue #157.

### DIFF
--- a/array.go
+++ b/array.go
@@ -775,8 +775,8 @@ func (a *Array) NotContainsOnly(values ...interface{}) *Array {
 // Example:
 //
 //	array := NewArray(t, []interface{}{"foo", 123, 123})
-//	array.ContainsAny(123, "foo", "FOO")
-//	array.ContainsAny("FOO")
+//	array.ContainsAny(123, "foo", "FOO") // success
+//	array.ContainsAny("FOO") // failure
 func (a *Array) ContainsAny(values ...interface{}) *Array {
 	a.chain.enter("ContainsAny()")
 	defer a.chain.leave()
--- a/array.go
+++ b/array.go
@@ -819,8 +819,8 @@ func (a *Array) ContainsAny(values ...interface{}) *Array {
 // Example:
 //
 //	array := NewArray(t, []interface{}{"foo", 123})
-//	array.NotContainsAny(123)
-//	array.NotContainsAny("bar", 124)
+//	array.NotContainsAny(123) // success
+//	array.NotContainsAny("bar", 124) // failure
 func (a *Array) NotContainsAny(values ...interface{}) *Array {
 	a.chain.enter("NotContainsAny()")
 	defer a.chain.leave()
--- a/array.go
+++ b/array.go
@@ -768,6 +768,52 @@ func (a *Array) NotContainsOnly(values ...interface{}) *Array {
 	return a
 }
 
+// ContainsAny succeeds if array contains at least one element from the given elements.
+// Before comparison, array and all elements are converted
+// to canonical form.
+//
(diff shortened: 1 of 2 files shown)


array.go
```diff
@@ -768,6 +768,97 @@ func (a *Array) NotContainsOnly(values ...interface{}) *Array {
...
+		a.chain.fail(AssertionFailure{
+			Type:      AssertContainsElement,
+			Actual:    &AssertionValue{a.value},
+			Expected:  &AssertionValue{values},
+			Reference: &AssertionValue{values},
+			Errors: []error{
+				errors.New("expected:" +
+					" array should contain at least one element from reference array"),
+			},
+		})
```
gavv, line 812: When assertion type is AssertContainsElement or AssertNotContainsElement, `Expected` is displayed as `expected element` (according to semantics described in comment to these constants).

Thus, the failure message looks strange: it says that expected element was the whole array. You can see it if you run `go test -v -run TestArrayContainsAny`.

I think we can just remove `Expected` field here and keep only `Actual` and `Reference`.

array.go
```diff
@@ -768,6 +768,97 @@ func (a *Array) NotContainsOnly(values ...interface{}) *Array {
...
+
+	if !foundAny {
+		a.chain.fail(AssertionFailure{
+			Type:      AssertContainsElement,
+			Actual:    &AssertionValue{a.value},
+			Expected:  &AssertionValue{values},
+			Reference: &AssertionValue{values},
+			Errors: []error{
+				errors.New("expected:" +
+					" array should contain at least one element from reference array"),
```
gavv, line 810: Nitpick: usually httpexpect errors are phrased like "expected: foo is bar" instead of "expected: foo should be bar", so I suggest to change it to "expected: array contains ...".

array.go
```diff
@@ -768,6 +768,97 @@ func (a *Array) NotContainsOnly(values ...interface{}) *Array {
...
+
+	if foundAny {
+		a.chain.fail(AssertionFailure{
+			Type:      AssertContainsElement,
+			Actual:    &AssertionValue{a.value},
+			Expected:  &AssertionValue{values},
+			Reference: &AssertionValue{values},
+			Errors: []error{
+				errors.New("expected: array should not contain any elements from reference array"),
+			},
```
gavv, line 855: See my comment above regarding Expected field for these type of assertion.

For NotContainsAny, I suggest to set Expected to the element which we have found.

array.go
gavv: Same for error in NotContainsAny

array_test.go
```diff
@@ -564,6 +566,56 @@ func TestArrayContainsOnly(t *testing.T) {
...
+	value.chain.assertFailed(t)
+	value.chain.reset()
+
+	value.NotContainsAny("FOO")
+	value.chain.assertOK(t)
+	value.chain.reset()
+
+	value.ContainsAny([]interface{}{123, "foo"})
+	value.chain.assertFailed(t)
+	value.chain.reset()
```
gavv, line 612: Seems this is the only check in TestArrayContainsAny where `ContainsOnly` fails. This check is good, but it's for "unusual" case (when the argument is array).

I suggest to add a check for "usual" case as well, like this:

```
value.ContainsAny("bar")
// assertFailed
```

gavv (commented):
Thanks for PR!! Here are some comments

array.go
gavv: Yeah, and also let's replace AssertContainsElement with AssertNotContainsElement

gavv:
Thanks!

Issue #166, opened 2022-11-23
Implement Array.Every and Object.Every


gavv:
Add new methods to Array and Object that allow to run function for every element:

```go
func (a *Array) Every(fn func(index int, value *httpexpect.Value)) *Array

...

array.Every(func(index int, value *httpexpect.Value) {
    // do something with value, for example:
    value.String().NotEmpty()
})
```

```go
func (o *Object) Every(fn func(key string, value *httpexpect.Value)) *Object

....

object.Every(func(key string, value *httpexpect.Value) {
    // do something with value, for example:
    value.String().NotEmpty()
})
```

New methods should have documentation comment with example and be covered by unit tests.

When array or object is constructing value to be passed to function, it should clone chain and append new path entry to it, like it is implemented in `Array.Iter()`, something like this:

```go
valueChain := a.chain.clone()
valueChain.enter("Every[%q]", key)

value := newValue(valueChain, o.value[key])
```

See also #168, #169.

angshumanHalder:
Hey @gavv I would like to take up this issue if that's okay.

gavv:
Sure, you're welcome! Thanks.

gavv:
Merged.

Issue #169, opened 2022-11-24
 Implement Array.Filter and Object.Filter


gavv:
Add new methods to Array and Object that allow to filter element(s) by predicate:

```go
func (a *Array) Filter(fn func(index int, value *httpexpect.Value) bool) *Array

...

newArray := array.Filter(func(index int, value *httpexpect.Value) bool {
    // make assertions
    value.String().NotEmpty()

    // return boolean
    return true
})
```

```go
func (o *Object) Filter(fn func(key string, value *httpexpect.Value) bool) *Object

....

newObject := object.Filter(func(key string, value *httpexpect.Value) bool {
    // make assertions
    value.String().NotEmpty()

    // return boolean
    return true
})
```

New methods should have documentation comment with example and be covered by unit tests.

Description:

* Array.Filter returns new Array with only those elements from original array which match predicate
* Object.Filter is similar, but with map instead of array

Predicate is considered to succeed if BOTH conditions are met:

* assertions inside predicate did not fail (like `value.String().NotEmpty()` in the above example)
* predicate returned true

To implement this, we need to do something like this for every predicate call:

```go
predicateChain := a.chain.clone()
predicateChain.enter("Filter[%q]", key)

// make assertions inside predicate non-fatal
// they will be logged, but wont be reported as failures
predicateChain.setFatal(false)

// set failure callback that will be used by chain and will be inherited by its children
// this callback will just set a boolean, which indicates that the chain or one of its
// children had failed
predicateChainFailure := false
predicateChain.setFailCallback(func() {
    predicateChainFailure = true
})

value := newValue(predicateChain, ...)

ok := predicateFunc(key, value)

if !ok || predicateChainFailure {
    // predicate failed
} else {
    // predicate succeeded
}
```

See also #168, #166.

vtech6:
Hi, I would like to help with this issue :)

gavv:
@vtech6 Sure, you're welcome, thanks.

See this discussion: https://github.com/gavv/httpexpect/pull/176#discussion_r1033368911 Likely, the new chain.replace() method will be needed for Filter() as well.

Other discussions in #176 may be helpful too.

vtech6:
@gavv I opened a [pull request](https://github.com/gavv/httpexpect/pull/178), let me know if I you have any suggestions.

gavv:
Merged.

Issue #168, opened 2022-11-24
Implement Find and FindAll


gavv:
Add new methods to Array and Object that allow to find element(s) by predicate:

```go
func (a *Array) Find(fn func(index int, value *httpexpect.Value) bool) *Value
func (a *Array) NotFind(fn func(index int, value *httpexpect.Value) bool) *Array
func (a *Array) FindAll(fn func(index int, value *httpexpect.Value) bool) []Value

...

value := array.Find(func(index int, value *httpexpect.Value) bool {
    // make assertions
    value.String().NotEmpty()

    // return boolean
    return true
})
```

```go
func (o *Object) Find(fn func(key string, value *httpexpect.Value) bool) *Value
func (o *Object) NotFind(fn func(key string, value *httpexpect.Value) bool) *Object
func (o *Object) FindAll(fn func(key string, value *httpexpect.Value) bool) []Value

....

value := object.Find(func(key string, value *httpexpect.Value) bool {
    // make assertions
    value.String().NotEmpty()

    // return boolean
    return true
})
```

All new methods should have documentation comment with example and be covered by unit tests.

Description:

* Find returns first matched element, or fails and returns empty (non-nil) value if element was not found
* NotFind succeeds if no element matched predicate and fails if there is a match
* FindAll is like Find, but returns slice of all matched values

> This design is similar to String.Match/NotMatch/MatchAll

Predicate is considered to succeed if BOTH conditions are met:

* assertions inside predicate did not fail (like `value.String().NotEmpty()` in the above example)
* predicate returned true

To implement this, we need to do something like this for every predicate call:

```go
predicateChain := a.chain.clone()
predicateChain.enter("Find[%q]", key)

// make assertions inside predicate non-fatal
// they will be logged, but wont be reported as failures
predicateChain.setFatal(false)

// set failure callback that will be used by chain and will be inherited by its children
// this callback will just set a boolean, which indicates that the chain or one of its
// children had failed
predicateChainFailure := false
predicateChain.setFailCallback(func() {
    predicateChainFailure = true
})

value := newValue(predicateChain, ...)

ok := predicateFunc(key, value)

if !ok || predicateChainFailure {
    // predicate failed
} else {
    // predicate succeeded
}
```

See also #169, #166.

knry0329:
@gavv hi, can I pick this up?

gavv:
Sure, thanks!

PR #176, opened 2022-11-27
166 array and object every

angshumanHalder:
Issue: #166

### DIFF
--- a/array.go
+++ b/array.go
@@ -212,6 +212,61 @@ func (a *Array) Iter() []Value {
 	return ret
 }
 
+// Every runs the passed function on all the Elements
+// in the array and returns a new Array instance.
+//
+// If assertion inside function fails, the original Array is marked failed.
+//
+// Every will execute the function for all values in the array irrespective
+// of assertion failures for some values in the array.
+//
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	newArray := array.Every(func(idx int, val *httpexpect.Value) {
+//		value.String().NotEmpty()
+//	})
+func (a *Array) Every(fn func(index int, value *Value)) *Array {
+	a.chain.enter("Every()")
+	defer a.chain.leave()
+
+	if a.chain.failed() {
+		return a
+	}
+
+	if fn == nil {
+		a.chain.fail(AssertionFailure{
+			Type: AssertUsage,
+			Errors: []error{
+				errors.New("unexpected nil"),
+			},
+		})
+		return a
+	}
+
+	chainFailure := false
+
+	for idx, val := range a.value {
+		valueChain := a.chain.clone()
+		valueChain.enter("Every[%d]", val)
+
+		valueChain.setFatal(false)
+		valueChain.setFailCallback(func() {
+			chainFailure = true
+		})
+
+		fn(idx, newValue(valueChain, val))
+	}
+
+	if chainFailure {
+		a.chain.fail(AssertionFailure{
+			Type: AssertNotValid,
+		})
+	}
+
+	return a
+}
+
 // Empty succeeds if array is empty.
 //
 // Example:
@@ -860,30 +915,4 @@ func countElement(array []interface{}, element interface{}) int {
 		}
 	}
 	return count
-}
-
-// Every runs the passed function on all the Elements
-// in the array and returns a new Array instance.
-//
-// Example:
-//
-//	array := NewArray(t, []interface{}{"foo", "bar"})
-//	newArray := array.Every(func(idx int, val *httpexpect.Value) {
-//		value.String().NotEmpty()
-//	})
-func (a *Array) Every(fn func(index int, value *Value)) *Array {
-	newArrayValue := []interface{}{}
-
-	for idx, val := range a.value {
-		valueChain := a.chain.clone()
-		valueChain.enter("Every[%d]", val)
-
-		newValue := newValue(valueChain, val)
-
-		fn(idx, newValue)
-
-		newArrayValue = append(newArrayValue, newValue.value)
-	}
-
-	return newArray(a.chain, newArrayValue)
 }
--- a/array.go
+++ b/array.go
@@ -248,7 +248,7 @@ func (a *Array) Every(fn func(index int, value *Value)) *Array {
 
 	for idx, val := range a.value {
 		valueChain := a.chain.clone()
-		valueChain.enter("Every[%d]", val)
+		valueChain.enter("Every[%v]", val)
 
 		valueChain.setFatal(false)
 		valueChain.setFailCallback(func() {
--- a/array.go
+++ b/array.go
@@ -861,3 +861,29 @@ func countElement(array []interface{}, element interface{}) int {
 	}
 	return count
 }
+
+// Every runs the passed function on all the Elements
+// in the array and returns a new Array instance.
+//
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	newArray := array.Every(func(idx int, val *httpexpect.Value) {
+//		value.String().NotEmpty()
+//	})
+func (a *Array) Every(fn func(index int, value *Value)) *Array {
+	newArrayValue := []interface{}{}
+
+	for idx, val := range a.value {
+		valueChain := a.chain.clone()
+		valueChain.enter("Every[%d]", val)
+
+		newValue := newValue(valueChain, val)
+
+		fn(idx, newValue)
+
+		newArrayValue = append(newArrayValue, newValue.value)
+	}
+
+	return newArray(a.chain, newArrayValue)
+}
--- a/array.go
+++ b/array.go
@@ -212,8 +212,7 @@ func (a *Array) Iter() []Value {
 	return ret
 }
 
-// Every runs the passed function on all the Elements
-// in the array and returns a new Array instance.
+// Every runs the passed function on all the Elements in the array.
 //
 // If assertion inside function fails, the original Array is marked failed.
 //
@@ -223,7 +222,8 @@ func (a *Array) Iter() []Value {
 // Example:
 //
 //	array := NewArray(t, []interface{}{"foo", "bar"})
-//	newArray := array.Every(func(idx int, val *httpexpect.Value) {
+//
+//	array.Every(func(index int, value *httpexpect.Value) {
 //		value.String().NotEmpty()
 //	})
 func (a *Array) Every(fn func(index int, value *Value)) *Array {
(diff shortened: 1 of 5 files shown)


array.go
```diff
@@ -861,3 +861,29 @@ func countElement(array []interface{}, element interface{}) int {
...
+
+		newValue := newValue(valueChain, val)
+
+		fn(idx, newValue)
+
+		newArrayValue = append(newArrayValue, newValue.value)
+	}
+
+	return newArray(a.chain, newArrayValue)
+}
```
gavv, line 889: The idea in the issue was a bit different: apply some *check* to every element, and return the original array (unmodified).

So that when you write:

```go
array.Every( .... )
```

you're saying: every element of the array should satisfy given condition.

And this code implement Every as a transformation - it assumes that the function modifies value, and returns a new array with modified values, instead of original array.

This is a legit feature, but there are two issues:

* First, for such function a better name would be Transform or Map (better Transform to avoid confusion with go maps)

* Second, httpexpect structs like Array, Object, etc are immutable; they have no public methods for changing value (note that in your test you're modifying private fields, which is not possible outside of httpexpect package); so Transform's function argument would have different signature - since it can't modify value, it would return a new value instead

So, I think `Every` should be implemented according to original idea, and `Transform` should be a separate feature.

Next, look, we have this issue: #169

It describes another function, `Filter`. And I think instead of implementing both `Filter` and `Transform`, we can upgrade description for `Filter` and make it looking like this:

```go
newArray := array.Filter(func(index int, value *httpexpect.Value) (*httpexpect.Value, bool) {
    // make assertions
    value.String().NotEmpty()

    // return boolean
    return v, true
})
```

This way, it will combine both features in one method: filtering-out values (if desired) and transforming values (if desired).

What do you think?

array_test.go
```diff
@@ -748,3 +748,41 @@ func TestArrayConvertContains(t *testing.T) {
...
+	tableTests := []struct {
+		name     string
+		input    *Array
+		function func(idx int, val *Value)
+		output   *Array
+	}{
+		{
+			name:  "Non-empty array",
+			input: NewArray(reporter, []interface{}{1, 2}),
+			function: func(_ int, val *Value) {
```
gavv, line 764: Nitpick: can we also check that passed index is correct?

array_test.go
```diff
@@ -748,3 +748,41 @@ func TestArrayConvertContains(t *testing.T) {
 	value.chain.assertFailed(t)
 	value.chain.reset()
 }
+
+func TestArrayEvery(t *testing.T) {
```
gavv, line 752: We also need a couple of tests for failed assertion:

* for one of the element in the middle, function fails; we should check that in this case the original array's chain is marked as failed; we should also check that the function was still invoked for every element, instead of stopping after first failure

* assertion fails for multiple or maybe all elements; we should check that the function was still invoked for every element, instead of stopping after first failure

Same for TestObjectEvery.

array.go
gavv: In the issue description for `Every`, I forgot one important detail: if assertion inside function failed, the original object should be marked as failed as well. Sorry for inconvenience.

To achieve this, you can use `setFailCallback` as described in #168.

gavv: We should mention two important facts in documentation comments:

* if assertion inside function is failed, the original object is marked as failed
* if assertion inside function is failed, Every will anyway execute function for every value, instead of stopping after first failure

array_test.go
```diff
@@ -748,3 +748,41 @@ func TestArrayConvertContains(t *testing.T) {
...
+				if v, ok := val.value.(float64); ok {
+					val.value = int(v) * 3
+				}
+			},
+			output: NewArray(reporter, []interface{}{}),
+		},
+	}
+
+	for _, test := range tableTests {
+		t.Log("Running: " + test.name)
```
gavv, line 784: Nitpick: we can use subtests: `t.Run(test.name, ...`

Same for TestObjectEvery.

object.go
```diff
@@ -657,3 +657,30 @@ func checkSubset(outer, inner map[string]interface{}) bool {
 	}
 	return true
 }
+
+// Every runs the passed function for all the key value pairs in the object
+// value should be map[string]interface{} or struct.
```
gavv, line 662: I guess this line:

```
// value should be map[string]interface{} or struct.
```

should be removed

array.go
```diff
@@ -861,3 +861,29 @@ func countElement(array []interface{}, element interface{}) int {
...
+// Every runs the passed function on all the Elements
+// in the array and returns a new Array instance.
+//
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	newArray := array.Every(func(idx int, val *httpexpect.Value) {
+//		value.String().NotEmpty()
+//	})
+func (a *Array) Every(fn func(index int, value *Value)) *Array {
```
gavv, line 874: * Before any other checks, `Every` should check if the chain is failed already, and return in this case.

* The implementation will crash if `fn` is nil. Instead, if it's nil, we should report failure (AssertUsage) and return. See how we do it in Request.WithMatcher, for example.

   To make the assertion path in failure meaningful, you'll need to do it like this:

   ```go
   if fn == nil {
     a.chain.enter("Every()")
     defer a.chain.leave()
     a.chain.fail(...)
   }
   ```

object_test.go
```diff
@@ -753,3 +754,77 @@ func TestObjectConvertValueEqual(t *testing.T) {
 	value.chain.assertFailed(t)
 	value.chain.reset()
 }
+
+func TestObjectEvery(t *testing.T) {
```
gavv, line 758: Please also add `Every` invocation to TestObjectFailed and TestArrayFailed.

(pass non-nil func to that invocation to be sure that `Every` handled failed chain, not nil func).

gavv (commented):

array.go
gavv: Also I suggest to move `Every` upper in file: right before `Empty` method (for both array and objects).

array.go
angshumanHalder: Understood. This was a misunderstanding on my end. I think now that I have some context of what `Every` should do. The original Idea makes sense.

Although for `Filter` I think we can keep the original idea and create separate function `Transform` which allows users to manipulate the elements in the array/slice and since the httpexpect objects are immutable we can return a new array. 
```

newArray := array.Transform(func(index int, value *httpexpect.Value) *httpexpect.Value {
    // make assertions
    value.String().NotEmpty()

    // return boolean
    return v
})
```
Something like this. 

In this way we can clearly indicate that Filter function is only for filtering and cannot modify or manipulate the values
whereas Transform is just to modify the value. In this way the user can chain the function.

```
newArray := array.Filter().Transform()
```

Same can be done on the object. 

Does this seem like a good idea?

array_test.go
angshumanHalder: Sure. I will also modify the tests according to new change.

object.go
angshumanHalder: Sure.

array.go
gavv: Cool.
Agree, seems that separate Filter and Transform functions would be more intuitive.

array.go
angshumanHalder: Sure thing.

array.go
angshumanHalder: Hey @gavv would you mind looking into this new function. If everything's alright I will write the test cases for this.

```go
func (a *Array) Every(fn func(index int, value *Value)) *Array {
	a.chain.enter("Every()")
	defer a.chain.leave()

	if fn == nil {
		a.chain.fail(AssertionFailure{
			Type: AssertNotNil,
			Errors: []error{
				errors.New("expected function, got nil"),
			},
		})
		return a
	}

	chainFailure := false

	for idx, val := range a.value {
		valueChain := a.chain.clone()
		valueChain.enter("Every[%d]", val)

		valueChain.setFatal(false)
		valueChain.setFailCallback(func() {
			chainFailure = true
		})

		fn(idx, newValue(valueChain, val))
	}

	if chainFailure {
		a.chain.fail(AssertionFailure{
			Type: AssertNotValid,
		})
	}

	return a
}

```

angshumanHalder:
Hey @gavv updated the code according to our discussion. Do have a look

object_test.go
```diff
@@ -753,3 +756,84 @@ func TestObjectConvertValueEqual(t *testing.T) {
...
+	// 	{
+	// 		output: NewObject(reporter, map[string]interface{}{
+	// 			"foo": "123",
+	// 			"bar": []interface{}{"456", "789"},
+	// 			"baz": map[string]interface{}{
+	// 				"a": "b",
+	// 			},
+	// 		}),
+	// 	},
+	// }
```
gavv, line 776: Please remove commented code.

gavv (commented):
Thanks for update, see comments below

object.go
```diff
@@ -162,6 +162,60 @@ func (o *Object) Value(key string) *Value {
 	return newValue(o.chain, value)
 }
 
+// Every runs the passed function for all the key value pairs in the object
+//
+// in the array and returns a new Array instance.
+//
+// If assertion inside function fails, the original Array is marked failed.
+// Example:
```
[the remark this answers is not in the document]
gavv: `newObject := ` should re remove too (we don't create new object)

array.go
```diff
@@ -212,6 +212,61 @@ func (a *Array) Iter() []Value {
...
+	a.chain.enter("Every()")
+	defer a.chain.leave()
```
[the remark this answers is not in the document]
angshumanHalder: I assume this should be done for Object too?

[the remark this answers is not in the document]
gavv: Yep, sure.

gavv:
Thanks, LGTM!

angshumanHalder:
> @angshumanHalder, BTW, if you're interested to implement Transform() method that you suggested, you're welcome!
> 
> Note: I've just assigned @vtech6 is going to implement Filter: [#169 (comment)](https://github.com/gavv/httpexpect/issues/169#issuecomment-1329137741)

Sure.  I will take up the Transform method for both Array and objects

gavv:
I've "unresolved" some discussion that may be useful for future readers.

gavv:
> Sure. I will take up the Transform method for both Array and objects

Cool, appreciate that!

angshumanHalder:
> I've "unresolved" some discussion that may be useful for future readers.

No worries. BTW do I need to create a ticket for the Transform implementation or will you create one and assign it to me?

gavv:
> BTW do I need to create a ticket [...]

Yes please, would be awesome

gavv:

gavv:
Follow-up issue: #184

PR #178, opened 2022-11-29
[#169]Feat: add array and object filter

vtech6:
This PR adds the array and object filter method. It accepts a function that does a boolean check. A review will be highly appreciated.

### DIFF
--- a/array.go
+++ b/array.go
@@ -934,19 +934,31 @@ func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
 	var filteredArray []interface{}
 
 	chainFailure := false
+
 	for index, element := range a.value {
+
 		valueChain := a.chain.clone()
+		valueChain.setFatal(false)
 		valueChain.setFailCallback(func() {
 			chainFailure = true
 		})
+
 		if filter(index, newValue(valueChain, element)) {
 			filteredArray = append(filteredArray, element)
+		} else {
+			valueChain.replace("Filter[%v]", element)
 		}
+
 	}
 
 	if chainFailure {
-		a.chain.setFailed()
+		a.chain.fail(AssertionFailure{
+			Type: AssertUsage,
+			Errors: []error{
+				errors.New("predicate function failed"),
+			},
+		})
 	}
-	a.value = filteredArray
-	return a
+
+	return newArray(a.chain, filteredArray)
 }
--- a/array.go
+++ b/array.go
@@ -933,31 +933,22 @@ func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
 
 	var filteredArray []interface{}
 
-	chainFailure := false
-
 	for index, element := range a.value {
-
+		chainFailed := false
 		valueChain := a.chain.clone()
 		valueChain.setFatal(false)
+
 		valueChain.setFailCallback(func() {
-			chainFailure = true
+			chainFailed = true
+
 		})
 
-		if filter(index, newValue(valueChain, element)) {
+		if filter(index, newValue(valueChain, element)) && !chainFailed {
 			filteredArray = append(filteredArray, element)
 		} else {
 			valueChain.replace("Filter[%v]", element)
 		}
 
-	}
-
-	if chainFailure {
-		a.chain.fail(AssertionFailure{
-			Type: AssertUsage,
-			Errors: []error{
-				errors.New("predicate function failed"),
-			},
-		})
 	}
 
 	return newArray(a.chain, filteredArray)
--- a/array.go
+++ b/array.go
@@ -914,15 +914,15 @@ func countElement(array []interface{}, element interface{}) int {
 	return count
 }
 
-func (a *Array) Filter(filter func(index int, value *Value) bool) *Array{
+func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
 	a.chain.enter("Filter()")
 	defer a.chain.leave()
 
-	if a.chain.failed(){
+	if a.chain.failed() {
 		return a
 	}
 
-	if filter==nil{
+	if filter == nil {
 		a.chain.fail(AssertionFailure{
 			Type: AssertUsage,
 			Errors: []error{
@@ -932,11 +932,11 @@ func (a *Array) Filter(filter func(index int, value *Value) bool) *Array{
 	}
 
 	var filteredArray []interface{}
-	
-	chainFailure:=false
+
+	chainFailure := false
 	for index, element := range a.value {
 		valueChain := a.chain.clone()
-		valueChain.setFailCallback(func(){
+		valueChain.setFailCallback(func() {
 			chainFailure = true
 		})
 		if filter(index, newValue(valueChain, element)) {
@@ -944,7 +944,7 @@ func (a *Array) Filter(filter func(index int, value *Value) bool) *Array{
 		}
 	}
 
-	if chainFailure{
+	if chainFailure {
 		a.chain.setFailed()
 	}
 	a.value = filteredArray
--- a/array.go
+++ b/array.go
@@ -922,17 +922,16 @@ func countElement(array []interface{}, element interface{}) int {
 // is returned.
 //
 // If there are any failed assertions in the filtering function, the
-// item is omitted without causing chain failure.
+// item is omitted without causing test failure.
 //
 // Example:
 //
-// array := newArray(reporter, []interface{}{1, 2, "foo", "bar"}
-// filteredArray := array.Filter(func(_ int, value *Value) bool {
-//	value.String().NotEmpty()		//fails on 1 and 2
-//  return value != "bar"			//fails on "bar"
-// })
-//
-// assert.Equal(t, filteredArray, []interface{}{"foo"}) //succeeds
+//	array := NewArray([]interface{}{1, 2, "foo", "bar"}
+//	filteredArray := array.Filter(func(index int, value *Value) bool {
+//		value.String().NotEmpty()		//fails on 1 and 2
+//		return value.Raw() != "bar"		//fails on "bar"
+//	})
+//	filteredArray.Equal([]interface{}{"foo"})	//succeeds
 
 func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
 	a.chain.enter("Filter()")
@@ -960,7 +959,7 @@ func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
 		valueChain.setFailCallback(func() {
 			chainFailed = true
 		})
-		valueChain.replace("Filter[%v]", element)
+		valueChain.replace("Filter[%v]", index)
 		if filter(index, newValue(valueChain, element)) && !chainFailed {
 			filteredArray = append(filteredArray, element)
 		}
--- a/array.go
+++ b/array.go
@@ -928,11 +928,11 @@ func countElement(array []interface{}, element interface{}) int {
 //
 // object := make(map[string]interface{})
 // filteredObject := object.Filter(func(_ string, value *Value) bool {
-//	value.String().NotEmpty()                //fails on 1 and 2
-//  return value != "bar"                    //fails on "bar"
+//	value.String().NotEmpty()		//fails on 1 and 2
+//  return value != "bar"			//fails on "bar"
 // })
 //
-// filteredArray.Raw() == []interface{}{"foo"} //true
+// assert.Equal(t, filteredObject, []interface{}{"foo"}) //true
 
 func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
 	a.chain.enter("Filter()")
@@ -954,13 +954,11 @@ func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
 	filteredArray := []interface{}{}
 
 	for index, element := range a.value {
-		chainFailed := false
 		valueChain := a.chain.clone()
 		valueChain.setFatal(false)
-
+		chainFailed := false
 		valueChain.setFailCallback(func() {
 			chainFailed = true
-
 		})
 		valueChain.replace("Filter[%v]", element)
 		if filter(index, newValue(valueChain, element)) && !chainFailed {
--- a/array.go
+++ b/array.go
@@ -913,3 +913,40 @@ func countElement(array []interface{}, element interface{}) int {
 	}
 	return count
 }
+
+func (a *Array) Filter(filter func(index int, value *Value) bool) *Array{
+	a.chain.enter("Filter()")
+	defer a.chain.leave()
+
+	if a.chain.failed(){
+		return a
+	}
+
+	if filter==nil{
+		a.chain.fail(AssertionFailure{
+			Type: AssertUsage,
+			Errors: []error{
+				errors.New("unexpected nil function argument"),
+			},
+		})
+	}
+
+	var filteredArray []interface{}
+	
+	chainFailure:=false
+	for index, element := range a.value {
+		valueChain := a.chain.clone()
+		valueChain.setFailCallback(func(){
+			chainFailure = true
+		})
+		if filter(index, newValue(valueChain, element)) {
+			filteredArray = append(filteredArray, element)
+		}
+	}
+
+	if chainFailure{
+		a.chain.setFailed()
+	}
+	a.value = filteredArray
+	return a
+}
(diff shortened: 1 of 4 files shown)


gavv:
#169

array.go
```diff
@@ -913,3 +913,40 @@ func countElement(array []interface{}, element interface{}) int {
...
+				errors.New("unexpected nil function argument"),
+			},
+		})
+	}
+
+	var filteredArray []interface{}
+
+	chainFailure := false
+	for index, element := range a.value {
+		valueChain := a.chain.clone()
```
gavv, line 938: We need to use chain.replace as mentioned here: https://github.com/gavv/httpexpect/issues/169#issuecomment-1329202581

array.go
```diff
@@ -913,3 +913,40 @@ func countElement(array []interface{}, element interface{}) int {
...
+			chainFailure = true
+		})
+		if filter(index, newValue(valueChain, element)) {
+			filteredArray = append(filteredArray, element)
+		}
+	}
+
+	if chainFailure {
+		a.chain.setFailed()
+	}
```
gavv, line 949: In Filter(), in contrast to Every(), if the predicate fails, we should exclude the element from array, but we should not mark the array as failed. And we should use setFatal(false) to make failures inside the predicate non-fatal. See code snippets in #169.

And we'll to cover this behavior with tests.

array.go
```diff
@@ -913,3 +913,40 @@ func countElement(array []interface{}, element interface{}) int {
...
+	a.value = filteredArray
+	return a
```
gavv, line 951: Instead of modifying the original array, we need to return a new one, which content is filtered.

gavv (commented):
Hi, thanks for PR. See comments below

array.go
vtech6: I'm not sure how this should be done, do we have a similar example in the repo? Should I use the newArray method or did you have something else in mind?

array.go
vtech6: So if I understand correctly, we don't need to use the setFailCallback function, is that right?

gavv: Nope, we still need setFailCallback to detect that some assertion inside predicate failed. We only need to change our handling of this case: instead of marking original array as failed, we should just skip failed element.

array.go
gavv: Yep, you just need to construct a slice with filtered values and pass it to newArray.

vtech6: I'm sorry, but I can't figure it out - every time I use the newArray function it just copies the previous chain with its values unaltered and I have no idea how to make it work without overwriting the old values. Any hints?

array.go
vtech6: But in this case if we don't use the setFailed method, under what condition can the chain fail and what purpose does chainFailure serve?

gavv: Quote from the issue:

> Predicate is considered to succeed if BOTH conditions are met:
>
>    * assertions inside predicate did not fail (like value.String().NotEmpty() in the above example)
>    * predicate returned true

Imagine you have this call:

```go
object.Filter(func(key string, value *httpexpect.Value) bool {
    // make assertions
    value.String().NotEmpty()

    // return boolean
    return true
})
```

What Filter() should do for every value:

* check if any assertion failed **inside** predicate, in this case `value.String().NotEmpty()`
* check if predicate return true
* if at least one of the above is false, Filter() should skip this element

Important thing: if an assertion failed **inside predicate**, like `value.String().NotEmpty()` in this example, it does not mean that the test should fail. In this context failure means only "skip this element".

To achieve this, we need to do a few things:

* we need to clone the chain and pass this clone to predicate
* we need to setup the cloned chain as non-fatal (call setFalal(false)), so that assertions made in predicate would not cause failure of the whole test
* we need to be informed that the chain, or one of its children, failed inside the predicate; this can be done using setFailureCallback; if an assertion on value will fail inside predicate, this callback will be invoked and we will mark predicated as failed

No matter of predicate results, we should NOT call `a.chain.setFailed()` (where `a` is the original array). This will mark the whole array as failed and cause the whole test to fail, which we don't want.

array.go
gavv: TBH I don't quite understand the question. Maybe @angshumanHalder could show his code from early versions of #176? That PR had very similar code initially.

@vtech6 or you can show me specific code snippet and show when you have problematic / unexpected behavior.

angshumanHalder: @vtech6 you could do something like this 
```go
	array := make([]interface{}, 0)
	return newArray(a.chain, array)
```

where `a` is your original array

vtech6: @angshumanHalder when I do this, the values of the array stay exactly how they were before the filter, for example when I run the following test: 
```
	        array := NewArray(reporter, []interface{}{1, 2, 3, 4, 5, 6})
		array.Filter(func(index int, value *Value) bool {
			return value.Raw() != 2.0 && value.Raw() != 5.0
		})
		assert.Equal(t, []interface{}{1.0, 3.0, 4.0, 6.0}, array.Raw())
```
I get this error:
```
Not equal: 
        	            	expected: []interface {}{1, 3, 4, 6}
        	            	actual  : []interface {}{1, 2, 3, 4, 5, 6}
```
Here's the implementation of the filter:
```
var filteredArray = []interface{}

for index, element := range a.value {
		valueChain := a.chain.clone()
		valueChain.setFailCallback(func() {
			chainFailure = true
		})

		if filter(index, newValue(valueChain, element)) {
			filteredArray = append(filteredArray, element)
		} else {
			valueChain.replace("Filter[%v]", element)
		}

	}
...

return newArray(a.chain, filteredArray)
```
It works when I directly overwrite the value with `a.value = filteredArray` but it doesn't when I call the function above.

Also maybe you can help me verify if I use the `valueChain.replace` in the correct spot. Thanks for the feedback!

gavv: I guess that's because in the test above you're checking the original array, and instead you should check the array returned from Filter().

vtech6: Oh, you're absolutely right! Let me take a look.

vtech6: I confirm it works now :) I'll try to commit the issues tomorrow, just want to go through everything again polish it up a little.

gavv: Great, ping me when it's ready for review.

neverbeenthisweeb:
hi @vtech6 , maybe some suggestion

if you want to have this PR as draft, you can have the PR's state as `Draft`

later on, when this PR is _ready_, can change the state to `Ready for review`

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

vtech6:
@gavv I just pushed the changes, hopefully I managed to address all the comments :). Thanks again for the thorough feedback!

array.go
```diff
@@ -913,3 +913,43 @@ func countElement(array []interface{}, element interface{}) int {
...
+		valueChain.setFailCallback(func() {
+			chainFailed = true
+
+		})
+
+		if filter(index, newValue(valueChain, element)) && !chainFailed {
+			filteredArray = append(filteredArray, element)
+		} else {
+			valueChain.replace("Filter[%v]", element)
+		}
```
gavv, line 950: This piece looks strange:

* why do you call replace only if predicate failed?
* why do you call it after predicate? the idea is to replace chain path before passing chain to newValue, so that assertion made on value inside predicate would have replaced path

Instead, we can just write:

```go
valueChain.replace("Filter[%v]", element)

if filter(index, newValue(valueChain, element)) && !chainFailed {
    ...
```

array.go
```diff
@@ -913,3 +913,43 @@ func countElement(array []interface{}, element interface{}) int {
...
+
+	for index, element := range a.value {
+		chainFailed := false
+		valueChain := a.chain.clone()
+		valueChain.setFatal(false)
+
+		valueChain.setFailCallback(func() {
+			chainFailed = true
+
+		})
```
gavv, line 944: Nitpick: extra empty line

array.go
```diff
@@ -913,3 +913,43 @@ func countElement(array []interface{}, element interface{}) int {
 	}
 	return count
 }
+
+func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
```
gavv, line 917: Please add documentation comment with code example to Array.Filter and Object.Filter.

Don't forget to mention important details:

* that function returns new array
* when predicate is considered to be successful and failed
* what happens if assertion inside predicate fails

array_test.go
```diff
@@ -808,3 +812,37 @@ func TestArrayEvery(t *testing.T) {
 		array.chain.assertFailed(ts)
 	})
 }
+
+func TestArrayFilter(t *testing.T) {
```
gavv, line 816: Please also add tests for two corner cases:

* original array is empty
* original array is non-empty, but Filter() skips all elements and return empty array

gavv: Also, in each test, please also check that after Filter() call, original array was not modified (has the same value as when initialized).

gavv (commented):
Thanks for update! See comments below

gavv:

array_test.go
```diff
@@ -808,3 +812,66 @@ func TestArrayEvery(t *testing.T) {
...
+		})
+		assert.Equal(t, []interface{}{"foo", true, 1.0}, filteredArray.Raw())
+		assert.Equal(t, array.Raw(), []interface{}{"foo", "bar", true, 1.0})
+
+		filteredArray.chain.assertOK(t)
+	})
+
+	t.Run("Filter an array of different types and validate", func(ts *testing.T) {
+		reporter := newMockReporter(t)
+		array := NewArray(reporter, []interface{}{})
```
gavv, line 855: I guess the test name is wrong.

array.go
```diff
@@ -913,3 +913,60 @@ func countElement(array []interface{}, element interface{}) int {
...
+//
+// object := make(map[string]interface{})
+// filteredObject := object.Filter(func(_ string, value *Value) bool {
+//	value.String().NotEmpty()                //fails on 1 and 2
+//  return value != "bar"                    //fails on "bar"
+// })
+//
+// filteredArray.Raw() == []interface{}{"foo"} //true
+
+func (a *Array) Filter(filter func(index int, value *Value) bool) *Array {
```
gavv, line 937: A few problems here:

* example code should be indented with tab (see other methods in httpexpect)
* object variable should be httpexpect.Object, not map, I guess
* I think you forget to initialize map values
* this example should be for Array, not Object
* extra empty line between doc comment and method
* instead of `filteredArray.Raw() == []interface{}{"foo"}`, can you please provide an assertion using httpexpect? like, `filteredArray.Equal([]interface{}{"foo"})` - like we do in other documentation examples

gavv: Indentation inside example code is broken:

image
****

array.go
```diff
@@ -913,3 +913,60 @@ func countElement(array []interface{}, element interface{}) int {
...
+	}
+
+	filteredArray := []interface{}{}
+
+	for index, element := range a.value {
+		chainFailed := false
+		valueChain := a.chain.clone()
+		valueChain.setFatal(false)
+
+		valueChain.setFailCallback(func() {
```
gavv, line 961: Nitpick: it would be a bit more readble if you declare chainFailed right before setFailCallback call.

object.go
```diff
@@ -710,3 +710,59 @@ func checkSubset(outer, inner map[string]interface{}) bool {
...
+	filteredObject := make(map[string]interface{})
+
+	for key, element := range o.value {
+		chainFailed := false
+		valueChain := o.chain.clone()
+		valueChain.setFailCallback(func() {
+			chainFailed = true
+		})
+		valueChain.replace("Every[%v]", element)
+		if filter(key, newValue(valueChain, element)) && !chainFailed {
```
gavv, line 762: You forgot setFatal here. Please add one and cover it in tests - in every test, we need to check chain of the original array too, not only of filtered array.

array.go
gavv: This is still unresolved.

array_test.go
gavv: The second one is still unresolved.

object_test.go
```diff
@@ -826,3 +830,71 @@ func TestObjectEvery(t *testing.T) {
...
+		object := NewObject(reporter, map[string]interface{}{})
+		filteredObject := object.Filter(func(key string, value *Value) bool {
+			return false
+		})
+		assert.Equal(t, map[string]interface{}{}, filteredObject.Raw())
+
+		filteredObject.chain.assertOK(t)
+	})
+
+	t.Run("Filter should return an empty non-nil array if no items pass",
```
gavv, line 886: array -> object

object.go
```diff
@@ -710,3 +710,61 @@ func checkSubset(outer, inner map[string]interface{}) bool {
...
+	filteredObject := make(map[string]interface{})
+
+	for key, element := range o.value {
+		valueChain := o.chain.clone()
+		valueChain.setFatal(false)
+		chainFailed := false
+		valueChain.setFailCallback(func() {
+			chainFailed = true
+		})
+		valueChain.replace("Every[%v]", element)
```
gavv, line 763: Every -> Filter

array.go
```diff
@@ -913,3 +913,58 @@ func countElement(array []interface{}, element interface{}) int {
...
+
+// Filter accepts a function that returns a boolean. The function is ran
+// over the array items. If the function returns true, the item passes
+// the filter and is added to the new array of filtered items. If false,
+// the item is skipped (or in other words filtered out). After iterating
+// through all the items of the original array, the new filtered array
+// is returned.
+//
+// If there are any failed assertions in the filtering function, the
+// item is omitted without causing chain failure.
```
gavv, line 925: Let's replace "chain failure" with "test failure" because we don't expose chain to the user and they normally don't know what is it.

array.go
```diff
@@ -913,3 +913,58 @@ func countElement(array []interface{}, element interface{}) int {
...
+//
+// Example:
+//
+// array := newArray(reporter, []interface{}{1, 2, "foo", "bar"}
```
gavv, line 929: newArray -> NewArray
this is documentation example, so it should use exported functions

array.go
```diff
@@ -913,3 +913,58 @@ func countElement(array []interface{}, element interface{}) int {
...
+// filteredArray := array.Filter(func(_ int, value *Value) bool {
+//	value.String().NotEmpty()		//fails on 1 and 2
+//  return value != "bar"			//fails on "bar"
```
gavv, line 932: `value != "bar"` -> `value.Raw() != "bar"`

array.go
```diff
@@ -913,3 +913,58 @@ func countElement(array []interface{}, element interface{}) int {
...
+// through all the items of the original array, the new filtered array
+// is returned.
+//
+// If there are any failed assertions in the filtering function, the
+// item is omitted without causing chain failure.
+//
+// Example:
+//
+// array := newArray(reporter, []interface{}{1, 2, "foo", "bar"}
+// filteredArray := array.Filter(func(_ int, value *Value) bool {
```
gavv, line 930: Let's replace `_ int` with `index int` to give a hint for the meaning of that argument.

array.go
```diff
@@ -913,3 +913,58 @@ func countElement(array []interface{}, element interface{}) int {
...
+//	value.String().NotEmpty()		//fails on 1 and 2
+//  return value != "bar"			//fails on "bar"
+// })
+//
+// assert.Equal(t, filteredArray, []interface{}{"foo"}) //succeeds
```
gavv, line 935: In httpexpect docs we don't use testify (because not all users use it). So better replace this line with httpexpect assertions:

```
filteredArray.Equal([]interface{}{"foo"})
```

array.go
```diff
@@ -913,3 +913,58 @@ func countElement(array []interface{}, element interface{}) int {
...
+	filteredArray := []interface{}{}
+
+	for index, element := range a.value {
+		valueChain := a.chain.clone()
+		valueChain.setFatal(false)
+		chainFailed := false
+		valueChain.setFailCallback(func() {
+			chainFailed = true
+		})
+		valueChain.replace("Filter[%v]", element)
```
gavv, line 963: `"Filter[%v]", element` -> `"Filter[%v]", index`

gavv:
Thanks!!

gavv:
Small follow-up commit: a3e57670ac78b222fdc2311ae1e082890a288c4e

gavv:
One more: d7744177b9678b8a5df0f3d1abe8103414ea7830

gavv:
Follow-up issue: #184

Issue #177, opened 2022-11-28
Implement Array.Transform and Object.Transform


angshumanHalder:
Add new methods to Array and Object that allow to run function for every element:

```go
func (a *Array) Transform(fn func(index int, value *httpexpect.Value) (*httpexpect.Value, error)) *Array
...
array.Transform(func(index int, value *httpexpect.Value) (*httpexpect.Value, error) {
    // do something with value, for example:
    return httpexpect.NewValue(httpexpect.Reporter, value), nil
})
```

```go
func (a *Object) Transform(fn func(key string, value *httpexpect.Value) (*httpexpect.Value, nil)) *Object
...
object.Transform(func(key string, value *httpexpect.Value) *httpexpect.Value {
    // do something with value, for example:
    return httpexpect.NewValue(httpexpect.Reporter, value)
})
```
New methods should have documentation comment with example and be covered by unit tests.

When array or object is constructing value to be passed to function, it should clone chain and append new path entry to it.

```go
valueChain := a.chain.clone()
valueChain.replace("Transform[%q]", key)

value := newValue(valueChain, o.value[key])
```
Upon returning error or nil from the passed function Transform would use the original value in the new Array/Object

angshumanHalder:
Hey @gavv can you assign this one to I just created it. I don't think I have the permission to assign it to myself. 😅

gavv:
BTW I think there is no need to return error from callback function. If it wants to fail, it can use testing suite to do so (t.Fail, or testify).

gavv:
I think, if callback returned nil, Transform should fail and return empty array/object. It should be so no matter if we return error or not.

And for error, since httpexpect is used in tests, and tests are already capable of reporting errors (using e.g. t.Errorf or testify assertions), then there is no need to introduce additional way to report them.

I mean, this:

```go
array.Transform(func(index int, value *httpexpect.Value) *httpexpect.Value) {
    if ... {
        t.Errorf("something happened")
        return nil
    }
    return <new value>
})
```

seems a bit more natural **in tests**, compared to this:

```go
array.Transform(func(index int, value *httpexpect.Value) (*httpexpect.Value, error)) {
    if ... {
        return nil, errors.New("something happened")
    }
    return <new value>, nil
})
```

What do you think?

gavv:
BTW, instead of `httpexpect.NewValue()` and friends, user will most likely use `Expect.Value()` and friends, because it will return value with the same config. Note that the latter does not return error.

[...]

PR #183, opened 2022-12-01
177 array and object transform

angshumanHalder:
Issue #177 

Hey @gavv checkout this PR. Let me know if something needs to be changed in it.

### DIFF
--- a/array.go
+++ b/array.go
@@ -969,14 +969,17 @@ func countElement(array []interface{}, element interface{}) int {
 }
 
 // Transform runs the passed function on all the Elements in the array
-// returns a new array without effeecting original array.
+// and returns a new array without effeecting original array.
 //
 // Example:
 //
 //	array := NewArray(t, []interface{}{"foo", "bar"})
 //	transformedArray := array.Transform(
 //		func(index int, value interface{}) interface{} {
-//			return transformedValue.(interface{})
+//		  if val, ok := value.(float64); ok {
+//		    return int(val) * int(val)
+//		  }
+//	    return nil
 //		},
 //	)
 //	transformedArray.Equals([]interface{}{....})
--- a/array.go
+++ b/array.go
@@ -913,3 +913,66 @@ func countElement(array []interface{}, element interface{}) int {
 	}
 	return count
 }
+
+// Transform runs the passed function on all the Elements in the array.
+//
+// If transformation inside function fails, the original Array is marked failed.
+//
+// Transform will immediately stop execution of the function upon first transformation failures
+// or if return value is nil.
+//
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	transformedArray := array.Transform(func(index int, value *httpexpect.Value) *httpexpect.Value {
+//	  *httpexpect.NewValue(...)
+//	})
+func (a *Array) Transform(fn func(index int, value *Value) *Value) *Array {
+	a.chain.enter("Transform()")
+	defer a.chain.leave()
+
+	if a.chain.failed() {
+		return a
+	}
+
+	if fn == nil {
+		a.chain.fail(AssertionFailure{
+			Type: AssertUsage,
+			Errors: []error{
+				errors.New("unexpected nil function argument"),
+			},
+		})
+		return a
+	}
+
+	chainFailure := false
+
+	array := make([]interface{}, 0)
+
+	for index, val := range a.value {
+		valueChain := a.chain.clone()
+		valueChain.replace("Transform[%d]", index)
+
+		valueChain.setFailCallback(func() {
+			chainFailure = true
+		})
+
+		transformedValue := fn(index, newValue(valueChain, val))
+		if transformedValue == nil {
+			a.chain.fail(AssertionFailure{
+				Type: AssertNil,
+				Errors: []error{
+					errors.New("unexpected nil return value"),
+				},
+			})
+			return a
+		}
+		array = append(array, transformedValue)
+	}
+
+	if chainFailure {
+		a.chain.setFailed()
+	}
+
+	return newArray(a.chain, array)
+}
--- a/array.go
+++ b/array.go
@@ -945,17 +945,11 @@ func (a *Array) Transform(fn func(index int, value *Value) *Value) *Array {
 		return a
 	}
 
-	chainFailure := false
-
-	array := make([]interface{}, 0)
+	array := []interface{}{}
 
 	for index, val := range a.value {
 		valueChain := a.chain.clone()
 		valueChain.replace("Transform[%d]", index)
-
-		valueChain.setFailCallback(func() {
-			chainFailure = true
-		})
 
 		transformedValue := fn(index, newValue(valueChain, val))
 		if transformedValue == nil {
@@ -967,12 +961,9 @@ func (a *Array) Transform(fn func(index int, value *Value) *Value) *Array {
 			})
 			return a
 		}
-		array = append(array, transformedValue)
+		array = append(array, transformedValue.value)
 	}
 
-	if chainFailure {
-		a.chain.setFailed()
-	}
-
-	return newArray(a.chain, array)
+	newArr := newArray(a.chain, array)
+	return newArr
 }
--- a/array.go
+++ b/array.go
@@ -967,28 +967,24 @@ func countElement(array []interface{}, element interface{}) int {
 	return count
 }
 
-// Transform runs the passed function on all the Elements in the array.
-//
-// If transformation inside function fails, the original Array is marked failed.
-//
-// Transform will immediately stop execution of the function
-// upon first transformation failures
-// or if return value is nil.
+// Transform runs the passed function on all the Elements in the array
+// returns a new array without effeecting original array.
 //
 // Example:
 //
 //	array := NewArray(t, []interface{}{"foo", "bar"})
 //	transformedArray := array.Transform(
-//		func(index int, value *httpexpect.Value) *httpexpect.Value {
-//			return *httpexpect.NewValue(...)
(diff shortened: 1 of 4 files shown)


array.go
```diff
@@ -913,3 +913,59 @@ func countElement(array []interface{}, element interface{}) int {
 	}
 	return count
 }
+
+// Transform runs the passed function on all the Elements in the array.
+//
```
gavv, line 918: Let's also mention that it return a new array, and original one is not affected.

array.go
```diff
@@ -913,3 +913,59 @@ func countElement(array []interface{}, element interface{}) int {
...
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	transformedArray := array.Transform(
+//		func(index int, value *httpexpect.Value) *httpexpect.Value {
+//			return *httpexpect.NewValue(...)
+//		},
+//	)
+func (a *Array) Transform(fn func(index int, value *Value) *Value) *Array {
```
gavv, line 933: Given it a second thought, and after looking at real usage example in your tests, it seems that it would be more logical and convenient if transform function operated with raw values, instead of httpexpect.Value.

Why:

* passing httpexpect.Value to user func looks artificial: most times, user will just call `.Raw()` and don't use httpexpect.Value itself, because the point of Transform is to replace the underlying value, not to perform assertions on it

* returning httpexpect.Value, again, looks artificial: user func wants to return underlying value; however it has to wrap it with httpexpect.Value, and then we immediately unwrap it back because we also need only underlying value

So it seems this would a better API:

```go
func (a *Array) Transform(fn func(index int, value interface{}) interface{}) *Array
```

(Note that in this case nil return value is not a failure.)

What do you think?

array.go
```diff
@@ -913,3 +913,59 @@ func countElement(array []interface{}, element interface{}) int {
...
+	a.chain.enter("Transform()")
+	defer a.chain.leave()
+
+	if a.chain.failed() {
+		return a
+	}
```
gavv, line 939: Let's be consistent and ensure that Transform() always returns a new array, no matter if it failed or succeeded:

```
return newArray(a.chain, nil)
```

Same in other places.

array.go
```diff
@@ -913,3 +913,59 @@ func countElement(array []interface{}, element interface{}) int {
...
+// or if return value is nil.
+//
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	transformedArray := array.Transform(
+//		func(index int, value *httpexpect.Value) *httpexpect.Value {
+//			return *httpexpect.NewValue(...)
+//		},
+//	)
```
gavv, line 932: Let's provide specific implementation of func in the example, and add one more line that demonstrated what is the result:

```
transformedArray.Equals([]interface{}{....})
```

array.go
```diff
@@ -913,3 +913,59 @@ func countElement(array []interface{}, element interface{}) int {
...
+		valueChain.replace("Transform[%d]", index)
+
+		transformedValue := fn(index, newValue(valueChain, val))
+		if transformedValue == nil {
+			a.chain.fail(AssertionFailure{
+				Type: AssertNil,
+				Errors: []error{
+					errors.New("unexpected nil return value"),
+				},
+			})
```
gavv, line 964: AssertNil -> AssertNotNil (we assert that value should not be nil)

gavv (commented):
Thanks!

array.go
angshumanHalder: I see what you mean. We can unwrap the underlying value and pass it to the function and then wrap the return value and add it to the new Array/Object. However in that case how do the Transform function handle failures? Should we check if the chain's failbit is true and just return? If so then what do we return the copy of original array or the half transformed array?

gavv: > We can unwrap the underlying value and pass it to the function

Actually the underlying array holds unwrapped values.

> and then wrap the return value and add it to the new Array/Object.

Actually our filteredSlice also holds unwrapped values.

So no need for any wraps / unwraps at all (unlike current implementation).

> However in that case how do the Transform function handle failures?

httpexpect is for testing, so we can assume that transform function is executing inside a test. Thus, it can use regular facilities to report failures in tests - t.Errorf, t.Fatalf, testify.

We don't need to bother I think. Transform function will return something, maybe nil, maybe not. If it will decide to fail the test also, it's not our business, our help is not needed for that.

The reason why we're currently checking for nil return value is just because we can't do anything meaningful if we get nil httpexpect.Value. And returning nil is not the only way to fail, currently transform func can call t.Errorf to mark test failed, and we're not trying to handle it (and don't have to).

angshumanHalder: Hey @gavv I think we don't need 

> If transformation inside function fails, the original Array is marked failed.

This behaviour. We can just let the user handle whether the function failure should be fatal or just a error using t.Fatal/t.Error because in case of t.Error the passed function would still run for all the values. The reason I am saying this is because inside the `Transform` function would not know if the transformation failed.

```go
func (a *Array) Transform(fn func(index int, value interface{}) interface{}) *Array {
	a.chain.enter("Transform()")
	defer a.chain.leave()

	if a.chain.failed() {
		return a
	}

	if fn == nil {
		a.chain.fail(AssertionFailure{
			Type: AssertUsage,
			Errors: []error{
				errors.New("unexpected nil function argument"),
			},
		})
		return newArray(a.chain, nil)
	}

	array := []interface{}{}

	for index, val := range a.value {
		transformedValue := fn(index, val)
		array = append(array, transformedValue)
	}

	return newArray(a.chain, array)
}
```

Notice I removed the chain creation for each value since only raw values will be sent to the function. So the `Transform`
function would not know if there is any error. We can just leave the error handling to the user and let them decide do they want to fail the entire transformation operation or log the error and continue on. 

Does this make sense? What do you think?

gavv: Absolutely. Looks good

angshumanHalder: Awesome.

angshumanHalder: Hey @gavv I am done with changes. Have a look and lemme know if something else requires improvement.
Cheers.

array.go
```diff
@@ -967,3 +967,43 @@ func countElement(array []interface{}, element interface{}) int {
 	}
 	return count
 }
+
+// Transform runs the passed function on all the Elements in the array
+// returns a new array without effeecting original array.
```
gavv, line 972: `returns` -> `and returns`? Or add point after first sentence? Same for Object

array.go
```diff
@@ -967,3 +967,43 @@ func countElement(array []interface{}, element interface{}) int {
...
+//
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	transformedArray := array.Transform(
+//		func(index int, value interface{}) interface{} {
+//			return transformedValue.(interface{})
+//		},
+//	)
+//	transformedArray.Equals([]interface{}{....})
```
gavv, line 982: Could you please make this example a bit more complete? I mean, perform some real transformation in func (e.g., apply strings.ToUpper) and check for real values in transformedArray.Equals.

gavv (commented):
Thanks for update

gavv:
> [Coverage Status](https://coveralls.io/builds/54841247)
> 
> Coverage decreased (-0.2%) to 92.5% when pulling **[ab8886f](https://github.com/gavv/httpexpect/commit/ab8886fd63304610dfabd79c1a2de7e4338afff1) on angshumanHalder:177-array-and-object-Transform** into **[67cdcd4](https://github.com/gavv/httpexpect/commit/67cdcd4bfa69ab1c707405ad475e8175cd5eea2b) on gavv:master**.

@angshumanHalder this comment indicates that test coverage decreased in this PR.

If you follow the link to coveralls, you'll see that in array.go and object.go there are two new branches which are not covered (in Transform).

Could you cover those branches in unit tests?

array.go
angshumanHalder: Understood.

angshumanHalder:
> > [Coverage Status](https://coveralls.io/builds/54841247)
> > Coverage decreased (-0.2%) to 92.5% when pulling **[ab8886f](https://github.com/gavv/httpexpect/commit/ab8886fd63304610dfabd79c1a2de7e4338afff1) on angshumanHalder:177-array-and-object-Transform** into **[67cdcd4](https://github.com/gavv/httpexpect/commit/67cdcd4bfa69ab1c707405ad475e8175cd5eea2b) on gavv:master**.
> 
> @angshumanHalder this comment indicates that test coverage decreased in this PR.
> 
> If you follow the link to coveralls, you'll see that in array.go and object.go there are two new branches which are not covered (in Transform).
> 
> Could you cover those branches in unit tests?

Sure. I see there is one branch in `Every` which isn't covered. Will do that too.

gavv:
> Sure. I see there is one branch in Every which isn't covered. Will do that too.

There are actually more of them, see #184.

If you would like, you could add those tests in separate PR.

angshumanHalder:
> > Sure. I see there is one branch in Every which isn't covered. Will do that too.
> 
> There are actually more of them, see #184.
> 
> If you would like, you could add those tests in separate PR.

Umm.. how urgent is that? Maybe I can do them at my own pace. Is that okay?

angshumanHalder:
> > > > Sure. I see there is one branch in Every which isn't covered. Will do that too.
> > > 
> > > 
> > > There are actually more of them, see #184.
> > > If you would like, you could add those tests in separate PR.
> > 
> > 
> > Umm.. how urgent is that? Maybe I can do them at my own pace. Is that okay?
> 
> Of course, I just mean that those tests for Every() which you mentioned _can_ be done later in #184. You can do them in this PR, you can postpone them, you can pick up #184, or ignore it, completely up to you and thanks for the help :)

Awesome. BTW I have updated the coverage for the transform functions along with the doc comment changes.

PR #189, opened 2022-12-04
185 fix err handle and add unit test of string and number

### LINKED ISSUE #185: Improve test coverage of Number and String
See coverage reports here:

* https://coveralls.io/builds/54781134/source?filename=number.go
* https://coveralls.io/builds/54781134/source?filename=string.go

There are uncovered branches in Number.NotInRange and String.AsNumber. We should cover them in unit tests.

knry0329:
fix #185

### DIFF
--- a/string.go
+++ b/string.go
@@ -563,7 +563,7 @@ func (s *String) AsNumber(base ...int) *Number {
 		inum, err = strconv.ParseInt(s.value, base[0], 64)
 		num = float64(inum)
 
-		if err == strconv.ErrRange {
+		if errors.Is(err, strconv.ErrRange) {
 			var unum uint64
 
 			unum, err = strconv.ParseUint(s.value, base[0], 64)
--- a/string_test.go
+++ b/string_test.go
@@ -342,6 +342,14 @@ func TestStringAsNumber(t *testing.T) {
 		num2.chain.assertFailed(t)
 		assert.Equal(t, float64(0), num2.Raw())
 	})
+
+	t.Run("multiple_base", func(t *testing.T) {
+		value1 := NewString(reporter, "100")
+		num1 := value1.AsNumber(10, 16)
+		value1.chain.assertFailed(t)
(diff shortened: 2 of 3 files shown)


string.go
```diff
@@ -563,7 +563,7 @@ func (s *String) AsNumber(base ...int) *Number {
 		inum, err = strconv.ParseInt(s.value, base[0], 64)
 		num = float64(inum)
 
-		if err == strconv.ErrRange {
+		if errors.Is(err, strconv.ErrRange) {
```
knry0329, line 566: I suppose `errors.Is` can handle error properly

string_test.go
```diff
@@ -341,6 +341,20 @@ func TestStringAsNumber(t *testing.T) {
 		value2.chain.assertFailed(t)
 		num2.chain.assertFailed(t)
 		assert.Equal(t, float64(0), num2.Raw())
+
+		value3 := NewString(reporter, "8000000000000000")
+		num3 := value3.AsNumber(16)
+		value3.chain.assertOK(t)
+		num3.chain.assertOK(t)
+		assert.Equal(t, float64(9223372036854775808), num3.Raw())
```
knry0329, line 349: This test case may not be the expected output.
Converting `8000000000000000` from hexadecimal to decimal will be `9223372036854775808`,
but `num3.Raw()` is `9223372036854776000`.
This may be due to rounding by float64.

gavv: Good catch!

I've pushed follow-up commit: 67cdcd4bfa69ab1c707405ad475e8175cd5eea2b

I've replaced 8000000000000000 with 4000000000000000 (2^62), the value from which float64 starts loosing precision for integers.

I also added precision loss checks to AsNumber, so that it fails if the string is an integer that can not be represented as float64 without loss, and added tests for that.

gavv: I've created a new issue #191 for a more appropriate fix for this problem.

Issue #186, opened 2022-12-02
 Improve test coverage of Response methods


gavv:
Add unit tests to improve coverage of response.go
No need to cover panics, but we should cover all assertions
https://coveralls.io/builds/54781134/source?filename=response.go

Rohith-Raju:
Hey @gavv I'd  like to work on this issue

gavv:
Merged!

PR #193, opened 2022-12-07
Improving test coverage for Response methods

Rohith-Raju:
#186 
Signed-off-by: Rohith-Raju <rohithraju488@gmail.com>


Rohith-Raju:
@gavv I'm using the latest version of go and there seems to be a problem. Let me downgrade

gavv:
Hi, you can use any version of ho locally, but the code should be compatible with go 1.14 because we still support it.

Rohith-Raju:
@gavv is this approach acceptable?

gavv:
The test itself looks fine. We need to change two things though:

* test NewResponse instead of newResponse
* extract it into separate test, because TestResponseFailed is for testing behavior of Response when it's created with a failed chain

Let me know when it will be ready for review.

gavv:
Please rebase on fresh `master`.

Rohith-Raju:
@gavv  I couldn't use the NewResponse because it doesn't accept an array of rtt  args

gavv:
> @gavv I couldn't use the NewResponse because it doesn't accept an array of rtt .

Actually it does:

```
func NewResponse(
	reporter Reporter, response *http.Response, rtt ...time.Duration,
) *Response {
```

Rohith-Raju:
> 

Ok sure, let me retry

Rohith-Raju:
> > @gavv I couldn't use the NewResponse because it doesn't accept an array of rtt .
> 
> Actually it does:
> 
> ```
> func NewResponse(
> 	reporter Reporter, response *http.Response, rtt ...time.Duration,
> ) *Response {
> ```
Ok sure let me retry

Rohith-Raju:
@gavv is my approach fine ? should I add all the `AssertUsage` tests in one pr?

response_test.go
```diff
@@ -85,6 +85,18 @@ func TestResponseFailed(t *testing.T) {
 	})
 }
 
+func TestMultipleRttArgs(t *testing.T) {
+	check := func(resp *Response) {
+		resp.chain.assertFailed(t)
+	}
```
gavv, line 91: I think this check() function is unnecessary, we can inline it.

gavv:
Thanks for update.

> @gavv is my approach fine ?

Yep.

> should I add all the AssertUsage tests in one pr?

I think yes, why not.

gavv:
Also I think it makes sense to merge all usage tests into one test with sub-tests.

response_test.go
Rohith-Raju: Ok sure @gavv

Rohith-Raju:
> Also I think it makes sense to merge all usage tests into one test with sub-tests.

On it!!

Rohith-Raju:
@gavv done with the changes.

gavv:
Thanks!

Small follow-up commit: 82cc2f171b292c49fb73c4dcb109b8647ae8ea75

PR #194, opened 2022-12-08
Implement String.IsAscii

### LINKED ISSUE #154: Implement String.IsAscii
Implement new assertion `String.IsAscii`, which succeeds if all string characters are ASCII.

Also add counterpart `String.NotIsAscii`.

Failure message for `IsAscii` should contains character that does not belong to ASCII.

Assertions should have documentation comment with example and be covered by unit tests.

See also #153.

knry0329:
close #154

### DIFF
--- a/string.go
+++ b/string.go
@@ -752,8 +752,8 @@ func (s *String) MatchAll(re string) []Match {
 //
 // Example:
 //
-// str := NewString(t, "Hello")
-// str.IsASCII()
+//	str := NewString(t, "Hello")
+//	str.IsASCII()
 func (s *String) IsASCII() *String {
 	s.chain.enter("IsASCII()")
 	defer s.chain.leave()
@@ -772,7 +772,7 @@ func (s *String) IsASCII() *String {
 
 	if !isASCII {
 		s.chain.fail(AssertionFailure{
-			Type:   AssertContainsSubset,
+			Type:   AssertValid,
 			Actual: &AssertionValue{s.value},
 			Errors: []error{
 				errors.New("expected: all string characters are ascii"),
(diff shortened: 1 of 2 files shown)


string.go
```diff
@@ -747,6 +748,76 @@ func (s *String) MatchAll(re string) []Match {
...
+	}
+
+	if !isASCII {
+		s.chain.fail(AssertionFailure{
+			Type:   AssertContainsSubset,
+			Actual: &AssertionValue{s.value},
+			Errors: []error{
+				errors.New("expected: all string characters are ascii"),
+			},
+		})
```
gavv, line 780: AssertContainsSubset requires Actual and Expect options, see comment to that constant:

```
	// Check expression: [Actual] contains subset [Expected]
	AssertContainsSubset
	AssertNotContainsSubset
```

Since here we don't have any expected value,I think we can instead use more generic AssertValid type, which requires only Actual option.

string.go
```diff
@@ -747,6 +748,76 @@ func (s *String) MatchAll(re string) []Match {
...
+}
+
+// IsNotASCII succeeds if all string characters are not ASCII
+//
+// Example:
+//
+// str := NewString(t, "こんにちは")
+// str.IsNotASCII()
+func (s *String) IsNotASCII() *String {
+	s.chain.enter("IsNotASCII()")
```
gavv, line 793: IsNotASCII -> NotIsASCII, according to convention in this lib

string.go
```diff
@@ -747,6 +748,76 @@ func (s *String) MatchAll(re string) []Match {
...
+		return s
+	}
+
+	isASCII := false
+	for _, c := range s.value {
+		if c <= unicode.MaxASCII {
+			isASCII = true
+			break
+		}
+	}
```
gavv, line 806: Here we check that all characters are non-ASCII.

However, in practice we would usually call a string non-ASCII if it has at least one non-ASCII character. For example, we would call "Straße" non-ASCII string. Also, logical negation of IsASCII() would give us the same result.

So I suggest to fix the algorithm, error message, and comment.

string.go
```diff
@@ -747,6 +748,76 @@ func (s *String) MatchAll(re string) []Match {
 	return ret
 }
 
+// IsASCII succeeds if all string characters are ASCII
+//
+// Example:
+//
+// str := NewString(t, "Hello")
+// str.IsASCII()
```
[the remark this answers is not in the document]
gavv: Same for second function.

string.go
gavv: Same for second function.

gavv (commented):
Thanks!

knry0329:
Thank you for your review, @gavv !

string.go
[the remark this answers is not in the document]
knry0329: I fixed ed4c868

string.go
knry0329: I fixed ed4c868

string.go
knry0329: I rename ed4c868

string.go
knry0329: I fix algorithm ed4c868

gavv:
LGTM!

Issue #167, opened 2022-11-23
Implement Object.Iter


gavv:
Implement method similar to `Array.Iter()`, but for Object:

```go
func (o *Object) Iter() map[string]Value
```

Method should have documentation comment with example and be covered by unit tests.

PR #195, opened 2022-12-08
Create iter function for object

VacariGabriel:
Issue: https://github.com/gavv/httpexpect/issues/167

This is my first contact with the project and I am a beginner in Go.

I hope I can help :)

### DIFF
--- a/object.go
+++ b/object.go
@@ -762,14 +762,15 @@ func (o *Object) ValueNotEqual(key string, value interface{}) *Object {
 // Example:
 //
 // object := newObject(chain, map[string]interface{})
-// for n, val := range object.Iter() {
-//	    val.Equal(object[n])
-//	}
+//
+//	for n, val := range object.Iter() {
+//		    val.Equal(object[n])
+//		}
 func (o *Object) Iter() map[string]Value {
 	o.chain.enter("Iter")
 	defer o.chain.leave()
-	
-	if o.chain.failed() { 
+
+	if o.chain.failed() {
 		return map[string]Value{}
 	}
 
--- a/object.go
+++ b/object.go
@@ -757,6 +757,34 @@ func (o *Object) ValueNotEqual(key string, value interface{}) *Object {
 	return o.NotValueEqual(key, value)
 }
 
+// Iter returns a new map of Value
+//
+// Example:
+//
+// object := newObject(chain, map[string]interface{})
+// for n, val := range object.Iter() {
+//	    val.Equal(object[n])
+//	}
+func (o *Object) Iter() map[string]Value {
+	o.chain.enter("Iter")
+	defer o.chain.leave()
+	
+	if o.chain.failed() { 
+		return map[string]Value{}
+	}
+
+	obj := make(map[string]Value)
+
+	for k, v := range o.value {
+		valueChain := o.chain.clone()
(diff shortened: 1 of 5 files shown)


object.go
```diff
@@ -757,6 +757,35 @@ func (o *Object) ValueNotEqual(key string, value interface{}) *Object {
 	return o.NotValueEqual(key, value)
 }
 
+// Iter returns a new map of Value
```
gavv, line 760: Let's expand comment a bit:

```
Iter returns a new map of Values attached to object elements.
```

object.go
```diff
@@ -757,6 +757,35 @@ func (o *Object) ValueNotEqual(key string, value interface{}) *Object {
...
+//
+// Example:
+//
+// object := newObject(chain, map[string]interface{})
+//
+//	for n, val := range object.Iter() {
+//		    val.Equal(object[n])
+//		}
```
gavv, line 768: * Please fix indentation
* Let's rename `n` to `key` and `val` to `value`
* Let's provide a valid and complete example similar to one in `Array.Iter`

object.go
```diff
@@ -757,6 +757,35 @@ func (o *Object) ValueNotEqual(key string, value interface{}) *Object {
...
+func (o *Object) Iter() map[string]Value {
+	o.chain.enter("Iter")
+	defer o.chain.leave()
```
gavv, line 771: `"Iter"` -> `"Iter()"`, according to convention in this lib

object.go
```diff
@@ -757,6 +757,35 @@ func (o *Object) ValueNotEqual(key string, value interface{}) *Object {
...
+
+	if o.chain.failed() {
+		return map[string]Value{}
+	}
+
+	obj := make(map[string]Value)
+
+	for k, v := range o.value {
+		valueChain := o.chain.clone()
+		valueChain.enter("Iter[%d]", v)
```
gavv, line 781: * Let's change `enter` to `replace`; this will substitute last part of path from `Iter()` to `Iter[...]`
* Instead of `v`, we should use `k` (our assertion paths include keys, not values)
* Instead of `%d`, we should use `%q` (we use this format for keys)

object_test.go
```diff
@@ -130,6 +131,12 @@ func TestObjectGetters(t *testing.T) {
 	assert.Equal(t, nil, value.Value("BAZ").Raw())
 	value.chain.assertFailed(t)
 	value.chain.reset()
+
+	it := value.Iter()
+	assert.Equal(t, it["foo"].value, value.Value("foo").Raw())
+	assert.Equal(t, it["bar"].value, value.Value("bar").Raw())
```
gavv, line 137: Let's also check then `it` has expected length.

gavv (commented):
Thanks! See a few comments below

gavv:
I've pushed small refactoring of `Array.Iter`: d9fe9eea2093b453e3d962eb9ced84d497a77456

gavv:
Thanks. Here is a follow-up commit: 5265dda306f5775a3b0b7ae57c58806c2e9fd9cb

Changes:

* fixed indentation of code example
* made code example complete
* added chain checks to to the test

PS. please don't resolve discussions by yourself, usually reviewer does it.

PR #198, opened 2022-12-10
Add retry sleep with context

neverbeenthisweeb:
Issue: https://github.com/gavv/httpexpect/issues/163

Changes:

- Update `sleepFn` to return `<-chan time.Time`
- Update unit test
- Update e2e test

### DIFF
--- a/e2e_context_test.go
+++ b/e2e_context_test.go
@@ -501,7 +501,7 @@ func TestContextPerRequestRetry(t *testing.T) {
 	t.Run("cancelled on second retry attempt", func(t *testing.T) {
 		var callCount int
 		var isCtxCancelled bool
-		ctxCancellation := make(chan bool, 1) // To cancel context after frist retry attempt
+		ctxCancellation := make(chan bool, 1) // To cancel context after first retry attempt
 		retryDelayDuration := 5 * time.Second
 
 		ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
--- a/e2e_context_test.go
+++ b/e2e_context_test.go
@@ -393,3 +393,45 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
 	// expected error should occur
 	assert.True(t, suppressor.expErrorOccurred)
 }
+
+func TestContextPerRequestWithRetryCancelledByContext(t *testing.T) {
+	var callCount int
+	var isConfigCtxCancelled bool
+	delayDuration := TimeOutDuration / 2
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		callCount++
+		w.WriteHeader(http.StatusInternalServerError)
+	}))
+	defer ts.Close()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func(fn context.CancelFunc) {
+		time.Sleep(delayDuration / 2)
+		fn()
+		isConfigCtxCancelled = true
+	}(cancel)
+
+	// Config with context cancelled error
+	suppressor := newExpErrorSuppressor(t,
+		func(err error) bool {
+			return strings.Contains(err.Error(), context.Canceled.Error())
+		})
+
+	e := WithConfig(Config{
+		BaseURL:          ts.URL,
+		AssertionHandler: suppressor,
+	})
+
+	e.GET("/").
+		WithContext(ctx).
+		WithMaxRetries(1).
+		WithRetryPolicy(RetryAllErrors).
+		WithRetryDelay(delayDuration, TimeOutDuration).
+		Expect()
+
+	assert.Equal(t, 1, callCount)
+	assert.True(t, isConfigCtxCancelled)
+	assert.True(t, suppressor.expErrorOccurred)
+}
--- a/e2e_context_test.go
+++ b/e2e_context_test.go
@@ -394,7 +394,7 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
 	assert.True(t, suppressor.expErrorOccurred)
 }
 
-func TestContextPerRequestWithRetryCancelledByContext(t *testing.T) {
+func TestContextPerRequestRetryCancelledByContext(t *testing.T) {
 	var callCount int
 	var isConfigCtxCancelled bool
 	delayDuration := TimeOutDuration / 2
--- a/e2e_context_test.go
+++ b/e2e_context_test.go
@@ -420,13 +420,13 @@ func TestContextPerRequestRetry(t *testing.T) {
 
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		go func(fn context.CancelFunc) {
+		go func() {
 			time.Sleep(ctxCancellationSleepDuration)
-			fn()
+			cancel()
 			m.Lock()
 			defer m.Unlock()
 			isCtxCancelled = true
-		}(cancel)
+		}()
 
 		// Config with all error
 		suppressor := newExpErrorSuppressor(t,
@@ -467,13 +467,13 @@ func TestContextPerRequestRetry(t *testing.T) {
 
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		go func(fn context.CancelFunc) {
+		go func() {
 			time.Sleep(ctxCancellationSleepDuration)
-			fn()
+			cancel()
 			m.Lock()
 			defer m.Unlock()
 			isCtxCancelled = true
-		}(cancel)
+		}()
 
 		// Config with context cancelled error
 		suppressor := newExpErrorSuppressor(t,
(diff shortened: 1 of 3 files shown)


neverbeenthisweeb:
hi @gavv wdyt about this? will proceed to e2e test when impl and unit test are OK

request_test.go
```diff
@@ -2861,4 +2866,39 @@ func TestRequestRetry(t *testing.T) {
 			assert.Equal(t, int64(100+200+300), totalSleepTime.Milliseconds())
 		})
 	})
+
+	t.Run("cancelled retries", func(t *testing.T) {
```
gavv, line 2870: Ideally we should also check that sleep function was NOT called.

Also would be nice to add a test when we call cancel after first retry, before next try.

e2e_context_test.go
```diff
@@ -393,3 +393,45 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
...
 	assert.True(t, suppressor.expErrorOccurred)
 }
+
+func TestContextPerRequestRetryCancelledByContext(t *testing.T) {
+	var callCount int
+	var isConfigCtxCancelled bool
+	delayDuration := TimeOutDuration / 2
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		callCount++
```
gavv, line 403: Since handler is invoked from another goroutine, we should use atomic for call count.

e2e_context_test.go
```diff
@@ -393,3 +393,45 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
...
+		w.WriteHeader(http.StatusInternalServerError)
+	}))
+	defer ts.Close()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func(fn context.CancelFunc) {
+		time.Sleep(delayDuration / 2)
+		fn()
+		isConfigCtxCancelled = true
```
gavv, line 413: We should use atomic for isConfigCtxCancelled as well.

e2e_context_test.go
```diff
@@ -393,3 +393,45 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
...
+	e := WithConfig(Config{
+		BaseURL:          ts.URL,
+		AssertionHandler: suppressor,
+	})
+
+	e.GET("/").
+		WithContext(ctx).
+		WithMaxRetries(1).
+		WithRetryPolicy(RetryAllErrors).
+		WithRetryDelay(delayDuration, TimeOutDuration).
```
gavv, line 431: So here is the timing:

* Request.Expect sends request, receives error
* Request.Expect sleeps for 500ms
* during this sleep, after 250ms, we cancel context
* we assume that Request.Expect will see the cancelled context before timeout expires (i.e. in less then 250ms)

This makes sense and is usually OK, but sometimes on slow CI workers there are big delays. So usually it's better to avoid tests that depend on short timeouts.

I think here we can just use some very large timeout like a minute or 10 minutes, because the test anyway does not wait for the full timeout, right? This way we can be sure that the test will work even when CI is very slow.

If we set very large timeout, e.g. 10 minutes, we can also add a check that the actual wait time was significantly smaller (e.g. less then 5 minutes).

gavv (commented):
Thanks! Here are a few comments

e2e_context_test.go
neverbeenthisweeb: > This makes sense and is usually OK, but sometimes on slow CI workers there are big delays. So usually it's better to avoid tests that depend on short timeouts.

hi, sry but could you elaborate more about this big delays? where will this big delays happen? 

and what will it cause? (could it _somehow_ makes the 500ms comes earlier than 200ms)

because I think at the end 250ms will come earlier than 500ms right

many thanks!

gavv: In general, if you run two goroutines:

```go
go func() {
    time.Sleep(time.Second)
    println("g1")
}()

go func() {
    time.Sleep(time.Second * 2)
    println("g2")
}()
```

The following is guaranteed:

* g1 will be printed *at least* after 1 second
* g2 will be printed *at least* after 2 seconds

Nothing more. Here are some examples of what is NOT guaranteed:

* g1 will be printed exactly after 1 second
* g1 will be printed before g2

Normally, the two assertions above are true. However, on a loaded system, when there are more ready goroutines than available CPUs, both prints may be delayed and even reordered, depending on how overloaded are goroutine queues and how they will be scheduled.

So it's better to avoid relying on real time when possible and don't assume that everything will be scheduled in time and in order.

neverbeenthisweeb: got it

will increase the time

thanks for the explanation

neverbeenthisweeb:
updated, ready for review

request_test.go
neverbeenthisweeb: > Ideally we should also check that sleep function was NOT called.

added

btw i can't mock the sleepFn then sum the duration, because (it looks like) the sleepFn is still called in the select statement (on this line [request.go](https://github.com/gavv/httpexpect/pull/198/files#diff-3c2feffcf57ffb6c596dc5608186c518252092b7017be5189d3dc12aa3c45cecR1931)). So, i measure the elapsed time instead.

> Also would be nice to add a test when we call cancel after first retry, before next try.

this is already covered in the e2e test. would that be enough? thanks

e2e_context_test.go
```diff
@@ -393,3 +393,160 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
...
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		go func() {
+			time.Sleep(ctxCancellationSleepDuration)
+			cancel()
+			m.Lock()
+			defer m.Unlock()
+			isCtxCancelled = true
+		}()
```
gavv, line 429: This sleep is always longer then the whole test, so why do we need it at all? I think we can just remove this goroutine from "not cancelled" test.

e2e_context_test.go
```diff
@@ -393,3 +393,160 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
...
+
+		// Config with all error
+		suppressor := newExpErrorSuppressor(t,
+			func(err error) bool {
+				return true
+			})
```
gavv, line 435: It seems suppressor is not needed in "not cancelled" test as well.

e2e_context_test.go
```diff
@@ -393,3 +393,160 @@ func TestContextPerRequestWithTimeoutCancelledByContext(t *testing.T) {
...
+		e.GET("/").
+			WithContext(ctx).
+			WithMaxRetries(1).
+			WithRetryPolicy(RetryAllErrors).
+			WithRetryDelay(retryDelayDuration, TimeOutDuration).
+			Expect()
+
+		assert.Equal(t, 1, callCount)
+		assert.True(t, isCtxCancelled)
+		assert.True(t, suppressor.expErrorOccurred)
```
gavv, line 498: Here you assume that cancel() will be called after first retry and before second retry. This will usually work, however it's not guaranteed, and this test is not deterministic. Examples:

* if httptest goroutine is scheduled lately, cancel() may be called before first try
* if cancellation goroutine is scheduled lately after sleep expires, cancel() may be called after second try

And so on. You can't rely on specific scheduling order.

To keep things simple and deterministic at the same time, I suggest the following:

* merge "cancelled on first retry attempt" and "cancelled on second retry attempt" into one test
* don't assume that cancel is called after first and before second try; only assume that it's called after first try
* to guarantee that it's called only after first try, use ctxCancellation channel like you did in "cancelled on second retry attempt"
* set WithMaxRetries to large value (don't limit retries)
* set min and max in WithRetryDelay to very small values (a few ms) to keep tests fast

request_test.go
gavv: Oh, I see.

Yes, e2e tests are good, let's keep this unit test as is.

I think this check:

```
assert.Equal(t, time.Duration(0), elapsed.Round(req.minRetryDelay))
```

is not really useful, and also I'd prefer to avoid relying on private fields in tests when possible. So I suggest to remove it.

Also I think we should set retry delay to some large value to ensure that it wont happen before cancellation.

neverbeenthisweeb: got it

e2e_context_test.go
neverbeenthisweeb: got it will do the update

neverbeenthisweeb:
updated, ready for review

gavv:
Thanks!

Here is a follow-up commit: b5e8ea9fc21e0bb3a3810aa8e0dabd640f06d98e

It reduces delays and fixes race described in https://github.com/gavv/httpexpect/pull/198#discussion_r1045293081

gavv:
Related: 0aa1e0cdd7b6b7ed7cd50ea9994a866c54d859cc

neverbeenthisweeb:
> Thanks!
> 
> Here is a follow-up commit: [b5e8ea9](https://github.com/gavv/httpexpect/commit/b5e8ea9fc21e0bb3a3810aa8e0dabd640f06d98e)
> 
> It reduces delays and fixes race described in [#198 (comment)](https://github.com/gavv/httpexpect/pull/198#discussion_r1045293081)

ok, thanks for the follow up commit

Issue #164, opened 2022-11-23
Improve test coverage of Request methods


gavv:
Add unit tests to improve coverage of request.go
No need to cover panics, but we should cover all assertions
https://coveralls.io/builds/54507988/source?filename=request.go

labala-sravani:
@gavv , I have added few test cases for the above request.go file in the [PR 199](https://github.com/gavv/httpexpect/pull/199). Please review.

gavv:
Thanks, will take a look tomorrow!

gavv:
Merged!

PR #199, opened 2022-12-10
Add test cases for request.go file

labala-sravani:
Add test cases for panics and validation failures in request.go file. This increases the test coverage as well. This is for [Issue 164](https://github.com/gavv/httpexpect/issues/164)

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -2197,17 +2197,16 @@ func TestValidationFailures(t *testing.T) {
 
 	t.Run("setupRedirect - redirect policy", func(t *testing.T) {
 		req := NewRequest(config, "METHOD", "/")
-		req.config.Client = nil
-		req.redirectPolicy = FollowAllRedirects
+		req.config.Client = &mockClient{}
+		req.WithRedirectPolicy(FollowAllRedirects)
 		req.Expect() // calls setupRedirect indirectly
 		req.chain.assertFailed(t)
 	})
 
 	t.Run("setupRedirect - maxRedirects is not -1", func(t *testing.T) {
 		req := NewRequest(config, "METHOD", "/")
-		req.config.Client = nil
-		req.redirectPolicy = defaultRedirectPolicy
-		req.maxRedirects = 0
+		req.config.Client = &mockClient{}
+		req.WithMaxRedirects(0)
 		req.Expect() // calls setupRedirect indirectly
 		req.chain.assertFailed(t)
 	})
--- a/request_test.go
+++ b/request_test.go
@@ -2129,16 +2129,9 @@ func TestRequestRedirect(t *testing.T) {
 }
 
 func TestValidationFailures(t *testing.T) {
-	factory := DefaultRequestFactory{}
-
-	client := &mockClient{}
-
-	reporter := newMockReporter(t)
-
 	config := Config{
-		RequestFactory: factory,
-		Client:         client,
-		Reporter:       reporter,
+		Reporter: newMockReporter(t),
+		Client:   &mockClient{},
 	}
 
 	t.Run("WithMatcher", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


gavv:
Nice, thanks for PR!

Some comments:

* It would be nice to avoid testing internal methods directly (like setupRedirects, encodeWebsocketRequest, setBody, etc). This internals are subject to change and the way how they are used from Request may change as well, making these tests less relevant. It would be better to do all tests via Request public interface, with rare exceptions like checking request `chain`.

* To test panics, we can use assert.Panics

* I think there is no need to hard-code specific panic messages in tests, so that we don't have to fix tests if we change messages a bit

labala-sravani:
Thanks for the review comments,
As per the second and third comment, I have modified the panic test cases to use assert.Panics and also removed the panic message hardcoding.

Will be checking the first comment. Currently, I have removed the internal req methods direct testing (except newRequest function). Please have a look in the mean time.

request_test.go
```diff
@@ -2128,6 +2128,149 @@ func TestRequestRedirect(t *testing.T) {
...
+
+	t.Run("WithFile", func(t *testing.T) {
+		req := NewRequest(config, "METHOD", "/")
+		req.WithFile("test-key", "test-path", nil, nil)
+		req.chain.assertFailed(t)
+	})
+
+	t.Run("setupRedirect - redirect policy", func(t *testing.T) {
+		req := NewRequest(config, "METHOD", "/")
+		req.config.Client = nil
```
gavv, line 2200: Client can't be nil, we're lucky that in this case request does not panic because of it, but it can change in future.

I guess instead, we should set Config.Client to any implementation which is not http.Client, e.g. we can use mockClient.

request_test.go
```diff
@@ -2128,6 +2128,149 @@ func TestRequestRedirect(t *testing.T) {
...
+		req.redirectPolicy = FollowAllRedirects
```
gavv, line 2201: Instead of setting private field directly, we can use WithRedirectPolicy

request_test.go
```diff
@@ -2128,6 +2128,149 @@ func TestRequestRedirect(t *testing.T) {
...
+		req.Expect() // calls setupRedirect indirectly
+		req.chain.assertFailed(t)
+	})
+
+	t.Run("setupRedirect - maxRedirects is not -1", func(t *testing.T) {
+		req := NewRequest(config, "METHOD", "/")
+		req.config.Client = nil
+		req.redirectPolicy = defaultRedirectPolicy
+		req.maxRedirects = 0
```
gavv, line 2210: Instead of setting private field directly, we can use WithMaxRedirects

gavv (commented):
Thanks, let me know when the rest will be ready. Here are a few minor comments.

gavv:
FYI: welcome to join discord chat if you prefer such format https://discord.gg/5SCPCuCWA9

gavv:
Please rebase on fresh `master`.

labala-sravani:
For now, I have fixed the minor comments. Rest, will check over the weekend. Thanks.

request_test.go
```diff
@@ -2128,6 +2128,148 @@ func TestRequestRedirect(t *testing.T) {
...
+		assert.Panics(t, func() { newRequest(nil, config, "", "") })
+	})
+
+	t.Run("newRequest - AssertionHandler is nil", func(t *testing.T) {
+		config := Config{
+			RequestFactory: factory,
+			Client:         client,
+			Reporter:       reporter,
+		}
+		config.AssertionHandler = nil
```
gavv, line 2267: I guess in these tests, we need to set AssertionHandler, not Reporter. Request does not use reporter.

request_test.go
gavv: Note: better set config.Client before creating request instead of setting private field after creating it.

gavv:
Thanks, let me know when it's ready for review.

labala-sravani:
@gavv, I have fixed your comments.
And I tried to add test cases for missing one’s as well, but I couldn’t add. Let me know if you have any suggestions. Otherwise, please review the changes in the MR.

gavv:
Thanks!

gavv:
Small follow-up commit: bbea5832ac61548a3e727fac1f2ba01ab1c6854d

Changes:

* create config with bad client only in tests then need it
* use AssertionHandler instead of Reporter
* remove TestWithRetryDelay (seems that test does not make much sense)
* cosmetic fixes

gavv:
I've created a follow-up issue for remaining uncovered paths: #200

Issue #204, opened 2022-12-19
Implement additional constructors


gavv:
For each matcher type (Value, Object, Array, etc) that have constructor in form `NewFoo(reporter, value)`, implement an additional constructor in form `NewFooC(config, value)`.

New constructors should use newChainWithConfig() and config.withDefaults().

For example, for Array it would look like this:

```go
func NewArray(reporter Reporter, value []interface{}) *Array {
    return newArray(newChainWithDefaults("Array()", reporter), value)
}

func NewArrayC(config Config, value []interface{}) *Array {
    return newArray(newChainWithConfig("Array()", config.withDefaults()), value)
}

func newArray(parent *chain, val []interface{}) *Array {
   ...
```

New constructors should have documentation comment with code example and be covered by unit tests.

New constructors are NOT needed for Request and Websocket structs.

Rohith-Raju:
@gavv can I give this a shot ?

gavv:
Sure!

PR #207, opened 2022-12-22
Additional constructors for matcher type  (Value, Object, Array, etc)

Rohith-Raju:
Signed-off-by: Rohith-Raju <rohithraju488@gmail.com>
Added new constructors for the following :

1. Array
2. Boolean
3. Cookie
4. Datetime
5. Duration
6. Match
7. Number
8. Object
9. String
10. Value

Closes: #204

### DIFF
--- a/array.go
+++ b/array.go
@@ -27,12 +27,11 @@ func NewArray(reporter Reporter, value []interface{}) *Array {
 
 // NewArrayC returns a new Array instance with config
 //
-// Both config and value should not be nil. If value is nil, failure is
-// reported.
+// Requirements for config are same as for WithConfig function.
 //
 // Example:
 //
-// array := NewArrayC(conf,[]interface{}{"foo",123})
+//	array := NewArrayC(config, []interface{}{"foo",123})
 func NewArrayC(config Config, value []interface{}) *Array {
 	return newArray(newChainWithConfig("Array()", config.withDefaults()), value)
 
--- a/array.go
+++ b/array.go
@@ -25,6 +25,19 @@ func NewArray(reporter Reporter, value []interface{}) *Array {
 	return newArray(newChainWithDefaults("Array()", reporter), value)
 }
 
+// NewArrayC returns a new Array instance with config
+//
+// Both config and value should not be nil. If value is nil, failure is
+// reported.
+//
+// Example:
+//
+// array := NewArrayC(conf,[]interface{}{"foo",123})
+func NewArrayC(config Config, value []interface{}) *Array {
+	return newArray(newChainWithConfig("Array()", config.withDefaults()), value)
+
+}
+
 func newArray(parent *chain, val []interface{}) *Array {
 	a := &Array{parent.clone(), nil}
 
--- a/cookie.go
+++ b/cookie.go
@@ -28,10 +28,11 @@ func NewCookie(reporter Reporter, value *http.Cookie) *Cookie {
 
 // NewCookieC returns a new Cookie instance with config
 //
-// config and value should not be nil.
+// Requirements for config are same as for WithConfig function.
 //
 // Example:
-// cookie := NewCookieC(conf,&http.Cookie{...})
+//
+//	cookie := NewCookieC(config, &http.Cookie{...})
 //
 //	cookie.Domain().Equal("example.com")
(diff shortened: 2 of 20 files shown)


Rohith-Raju:
@gavv I'll be going forward with the test cases if everything here looks fine.

gavv:
Thanks, looks good.

You forgot to call `withDefaults` in some places.

array.go
```diff
@@ -25,6 +25,19 @@ func NewArray(reporter Reporter, value []interface{}) *Array {
...
 }
 
+// NewArrayC returns a new Array instance with config
+//
+// Both config and value should not be nil. If value is nil, failure is
+// reported.
+//
+// Example:
+//
+// array := NewArrayC(conf,[]interface{}{"foo",123})
```
gavv, line 35: Please ensure that all code example are correctly indented.

There should be a tab (not a space and not a few spaces) right after `//`.

There should be also an empty line right after `Example:`.

E.g., this one is GOOD:

image

Also, there should NOT be an empty line between comment and `func`, like here:

image

Unfortunately we don't have a linter for this, so we have to review formatting manually.

cookie.go
```diff
@@ -26,6 +26,20 @@ func NewCookie(reporter Reporter, value *http.Cookie) *Cookie {
 	return newCookie(newChainWithDefaults("Cookie()", reporter), value)
 }
 
+// NewCookieC returns a new Cookie instance with config
+//
+// config and value should not be nil.
+//
+// Example:
+// cookie := NewCookieC(conf,&http.Cookie{...})
```
gavv, line 34: Here and in other examples:

* let's rename `conf` to `config`, like in other places in documentation
* don't forget spaces between arguments: `conf,&http.Cookie{...}` -> `conf, &http.Cookie{...}`

duration.go
```diff
@@ -23,6 +23,18 @@ func NewDuration(reporter Reporter, value time.Duration) *Duration {
 	return newDuration(newChainWithDefaults("Duration()", reporter), &value)
 }
 
+// NewDurationC returns a new Duration instance with config.
+//
+// config should not be nil.
```
gavv, line 28: Here and in other places.

> config should not be nil.

This comment doesn't make sense, config is not a pointer and you can't pass nil.

Instead, we have the following requirement:

> Either Reporter or AssertionHandler should not be nil.

However, instead of copy-pasting it, it's better to add a comment like this:

> Requirements for config are same as for WithConfig function.

gavv:
I've just created #208. Please adjust formatting according to examples from that issue.

Rohith-Raju:
Ok Sure !!

array.go
Rohith-Raju: @gavv When I add tab space golang linter (vscode) reverting back to single space. Also the above lines have single space too.

gavv: Interesting. Well, I suppose there should be a way to configure vscode to disable this behavior. 

BTW this kind of formatting was not invented for me, this is how godoc wants to see code examples, so we have to use it.

Rohith-Raju: > Interesting. Well, I suppose there should be a way to configure vscode to disable this behavior.

Sure, yea I'll do that. 

> BTW this kind of formatting was not invented by me, this is how godoc wants to see code examples, so we have to use it.

Sure, ill make them tab space.

Rohith-Raju:
@gavv For the test case, I was thinking something like this. 
```
func TestNewArrayC(t *testing.T) {
	testValue := []interface{}{"foo", 123}
	arr := NewArrayC(Config{
		Reporter: reporter,
	}, testValue)
	arr.chain.assertNotFailed(t)
	arr.Equal(testValue)
}

```
Is this sufficient? Or should transfer this into a subtest?

gavv:
Ideally we could add new test TestArrayConstructors with two subtests, with and without config, each one similar to what you showed above.

Same for other structs.

gavv:
Thank you

gavv:
Follow-up commit: c94170818fdd6fe56b40f9daa4b3fa92b6c9602d

gavv:
One more: 6cf31eeba3019e55ea6ac19b3ca04231aec5775e

PR #213, opened 2022-12-26
Added methods to enhance datetime class.

kaviranjan88:
- New Methods for datetime class
- Added unit test

### DIFF
--- a/datetime.go
+++ b/datetime.go
@@ -370,7 +370,8 @@ func (dt *DateTime) WeekDay() *Number {
 	return newNumber(dt.chain, float64(dt.value.Weekday()))
 }
 
-// YearDay returns the day of the year specified by datetime, in the range [1,365] for non-leap years,
+// YearDay returns the day of the year specified by datetime,
+// in the range [1,365] for non-leap years,
 // and [1,366] in leap years..
 //
 // Example:
@@ -388,7 +389,8 @@ func (dt *DateTime) YearDay() *Number {
 	return newNumber(dt.chain, float64(dt.value.YearDay()))
 }
 
-// Hour returns the hour within the day specified by datetime, in the range [0, 23]..
+// Hour returns the hour within the day specified by datetime,
+// in the range [0, 23]..
 //
 // Example:
 //
@@ -405,7 +407,8 @@ func (dt *DateTime) Hour() *Number {
 	return newNumber(dt.chain, float64(dt.value.Hour()))
 }
 
-// Minute returns the minute offset within the hour specified by datetime, in the range [0, 59]..
+// Minute returns the minute offset within the hour specified by datetime,
+// in the range [0, 59]..
 //
 // Example:
 //
@@ -422,7 +425,8 @@ func (dt *DateTime) Minute() *Number {
 	return newNumber(dt.chain, float64(dt.value.Minute()))
 }
 
-// Second returns the second offset within the minute specified by datetime, in the range [0, 59].
+// Second returns the second offset within the minute specified by datetime,
+// in the range [0, 59].
 //
 // Example:
 //
--- a/datetime.go
+++ b/datetime.go
@@ -288,8 +288,10 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
 //
 // Example:
 //
-//	dt := NewDateTime(t, time.Unix(0, 10))
-//	dt.Zone().Equal("/foo")
+//	tm, _ := time.Parse("Jan 2, 2006 at 3:04pm (MST)",
+//					"Dec 30, 2022 at 3:04pm (IST)")
+//	dt := NewDateTime(t, tm)
+//	dt.Zone().Equal("IST")
 func (dt *DateTime) Zone() *String {
 	dt.chain.enter("Zone()")
 	defer dt.chain.leave()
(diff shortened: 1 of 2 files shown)


gavv:
Hi, thanks for pr!

A couple of comments.

gavv:
For each method, could you please add a short comment describing the return value?

For example, for Month we should mention the range of returned integer (is it 0-11 or 1-12).

You can base comments on comments from time.Time.

gavv:
And another one. 

For the test, could you use a date-time which fields are all nonzero and all have different numeric values?

This is needed to be sure that the methods work properly.

gavv:
Let me know when it's ready for review.

kaviranjan88:
@gavv Hi, I have updated code as per your comments. Please have a look.

datetime.go
```diff
@@ -283,3 +283,214 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
...
+	defer dt.chain.leave()
+
+	if dt.chain.failed() {
+		return newNumber(dt.chain, float64(0))
+	}
+
+	return newNumber(dt.chain, float64(dt.value.Year()))
+}
+
+// Month returns the month of the year specified by datetime.
```
gavv, line 322: Let's also specify the range of returned value.

datetime.go
```diff
@@ -283,3 +283,214 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
...
+	defer dt.chain.leave()
+
+	if dt.chain.failed() {
+		return newNumber(dt.chain, float64(0))
+	}
+
+	return newNumber(dt.chain, float64(dt.value.Month()))
+}
+
+// Day returns the day of the month specified datetime.
```
gavv, line 339: Let's also specify the range of returned value.

datetime.go
```diff
@@ -283,3 +283,214 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
...
+	defer dt.chain.leave()
+
+	if dt.chain.failed() {
+		return newNumber(dt.chain, float64(0))
+	}
+
+	return newNumber(dt.chain, float64(dt.value.Day()))
+}
+
+// Weekday returns the day of the week specified by datetime.
```
gavv, line 356: Let's also specify the range of returned value.

datetime.go
```diff
@@ -283,3 +283,214 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
...
+
+	if dt.chain.failed() {
+		return newNumber(dt.chain, float64(0))
+	}
+
+	return newNumber(dt.chain, float64(dt.value.YearDay()))
+}
+
+// Hour returns the hour within the day specified by datetime,
+// in the range [0, 23]..
```
gavv, line 393: Extra dot

datetime.go
```diff
@@ -283,3 +283,214 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
...
+
+	if dt.chain.failed() {
+		return newNumber(dt.chain, float64(0))
+	}
+
+	return newNumber(dt.chain, float64(dt.value.Hour()))
+}
+
+// Minute returns the minute offset within the hour specified by datetime,
+// in the range [0, 59]..
```
gavv, line 411: Extra dot

gavv (commented):
Thanks for update. A few more comments

gavv:
Regarding code examples in comments:

* instead of unix(0,0), I suggest to use time.Parse with a date with distinct fileds, to make examples more convenient 
* when calling Equal, cast to float64 is not needed

datetime.go
kaviranjan88: Updated

datetime.go
kaviranjan88: Done

datetime.go
kaviranjan88: Done

datetime.go
kaviranjan88: Done

datetime.go
kaviranjan88: Done

kaviranjan88:
@gavv Hi, I have updated code as per your comments. Casting to float64 as time returns **int** while number class has return type as **float64**. Please suggest in case I am not doing it correct.

gavv:
Sorry, I was talking about code on examples in documentation comments - I suggested to use time.Parse and casts there, not in the test.

Regarding casts. Yes, they are needed in test. But seems they are not needed in those code examples because they use httpexpect's Equal, which allow numbers of different types (fliat and int).

kaviranjan88:
@gavv Hi, Updated examples as per your comments. Please have a look.

datetime.go
```diff
@@ -283,3 +283,230 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
 
 	return dt
 }
+
+// Zone returns a new String instance with datetime zone.
+//
+// Example:
+//
+//	tm, _ := time.Parse(time.UnixDate, "FRI Dec 30 15:04:05 IST 2022")
```
gavv, line 291: Could we replace "FRI" with "Fri" in examples? Usually it's how days are spelled.

datetime.go
```diff
@@ -283,3 +283,230 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
...
+
+	if dt.chain.failed() {
+		return newNumber(dt.chain, float64(0))
+	}
+
+	return newNumber(dt.chain, float64(dt.value.Day()))
+}
+
+// Weekday returns the day of the week specified by datetime,
+// in the range [0, 6].
```
gavv, line 364: We should mention in comment that 0 corresponds to Sunday.

datetime.go
```diff
@@ -283,3 +283,230 @@ func (dt *DateTime) NotInRange(min, max time.Time) *DateTime {
...
+//
+// Example:
+//
+//	tm, _ := time.Parse(time.UnixDate, "FRI Dec 30 15:04:05 IST 2022")
+//	dt := NewDateTime(t, tm)
+//	dt.WeekDay().Equal(5)
```
gavv, line 370: Instead of 5, let's use time.Friday constant

gavv:
Thanks, LGTM!

Issue #190, opened 2022-12-05
Add DefaultFormatter.EnableScientific


gavv:
Add flag EnableScientific to DefaultFormatter that enables printing floats in scientific form. Without this flag, floats should be always printed in non-scientific form.

See also https://github.com/gavv/httpexpect/issues/160 and https://github.com/gavv/httpexpect/issues/159.

neverbeenthisweeb:
hi @gavv , can I work on this? thanks

gavv:
Sure!

neverbeenthisweeb:
thanks

will make progress within this week

neverbeenthisweeb:
hi @gavv 

in httpexpect we have this `formatFloat` function, which is called by `filDelta`:

https://github.com/gavv/httpexpect/blob/0bdec41c7b776760239e0c3a4dfdfb80eef3207c/formatter.go#L394

question: is the scope of this change only within "`fillDelta` that calls `formatFloat`"? or it could be bigger?

because it looks like `formatRange` and `isNumber` also involves `float64` type

thanks

gavv:
Good question.

formatFloat is used only for printing delta when use use assertions like EqualDelta. This feature is definitely broader than just printing deltas, it should apply to actual and expected values in all assertion types.

Actually I don't quite understand why I added formatFloat. It seems misleading, because most times it's not used for formatting floats..

I think we should remove formatFloat at all and instead use formatValue and teach formatValue to print floats as formatFloat does it.

Also, as you correctly noticed, numbers may be printed via formatRange. Let's rework formatRange so that it will use formatValue to format min and max; this way all float formatting logic will be in one place - in formatValue.

Similarly, we should rework formatTyped to use formatValue when value is a number.

gavv:
Also, after more thought, it seems that it's better to reverse the option: add DisableScientific instead of EnableScientific.

It would work as follows:

* if DisableScientific is false (default), use `%g` to format floats (which will automatically use scientific notation for large exponents); currently behavior is the same for everything except deltas (which use formatFloat) - go automatically uses `%g` when you use `%v` with floats

[...]

PR #218, opened 2022-12-31
Update formatXXX to be methods of DefaultFormatter

neverbeenthisweeb:
Issue: (part of) https://github.com/gavv/httpexpect/issues/190

Tasks:

- [X] Update formatXXX to be methods of `DefaultFormatter`

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -229,11 +229,11 @@ func (f *DefaultFormatter) fillActual(
 
 	case AssertType, AssertNotType:
 		data.HaveActual = true
-		data.Actual = formatTyped(failure.Actual.Value)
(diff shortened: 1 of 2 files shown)


gavv:
I think it makes sense to split refactoring (making format funcs to be methods) and adding new flag into two prs. It would be easie to review them separately. If you don't mind this approach, let me know when the first part is ready for review.

gavv:
Let's rename formatTemplate to something without format, to make it distinct from formatXxx methods. E.g. applyTemplate.

neverbeenthisweeb:
got it, will do the change

neverbeenthisweeb:
@gavv ready for review

gavv:
Thanks!

PR #220, opened 2023-01-02
Add DefaultFormatter.DisableScientific

neverbeenthisweeb:
Issue: https://github.com/gavv/httpexpect/issues/190

Tasks:

- [x] Add DisableScientific flag in DefaultFormatter
- [x] Add/update _sample_ unit test (for early review)
- [ ] ~~Add/update remaining unit test (if _sample_ is ok)~~
- [X] Refine comments
- [ ] ~~Complete unit test with other assertion types (e.g. `formatting of AssertionList with number` as mentioned in https://github.com/gavv/httpexpect/pull/220#issuecomment-1399282552)~~


gavv:
I'd suggest to extract disable scientific ckecks into a separate test. This way it's easy to track what is covered for this flag.

Also, I'd double each check: one check with flag, one check without flag, both checks with the same number. This way it's easy to track how the flag changes behavior.

formatter.go
```diff
@@ -42,6 +43,9 @@ type DefaultFormatter struct {
 	// Exclude diff from failure report.
 	DisableDiffs bool
 
+	// Disable printing floats in scientific form.
```
gavv, line 46: We should mention what happens if the flag is not set.

gavv:
Thanks for PR! See a few comments above.

formatter.go
neverbeenthisweeb: ok, will update the comment

neverbeenthisweeb:
> I'd suggest to extract disable scientific ckecks into a separate test. This way it's easy to track what is covered for this flag.
> 
> Also, I'd double each check: one check with flag, one check without flag, both checks with the same number. This way it's easy to track how the flag changes behavior.

got it, will separate the unit tests

formatter.go
neverbeenthisweeb: how about?

```suggestion
	// Disable printing floats in scientific form. If not set will print floats in decimal notation.
```

neverbeenthisweeb: updated https://github.com/gavv/httpexpect/pull/220/commits/a8c1e6492c6083ccc1a3b858b16d711b6c43ec17

neverbeenthisweeb:
> > I'd suggest to extract disable scientific ckecks into a separate test. This way it's easy to track what is covered for this flag.
> > Also, I'd double each check: one check with flag, one check without flag, both checks with the same number. This way it's easy to track how the flag changes behavior.
> 
> got it, will separate the unit tests

updated:

- https://github.com/gavv/httpexpect/pull/220/commits/1bfe1aaf40662aa33a5d9c2504b2d4635ea6aad6
- https://github.com/gavv/httpexpect/pull/220/commits/cc14c1efc4dd9b6729b2ab2d5afe9532f5575ca3

wdyt @gavv ? I check the template string instead of `formatData` here because:

- The `AssertionFailure` varies between cases of expected, actual, reference, and delta. Checking the `formatData` is less _generic_ because each "template value" is stored in different field (`.Expected`, `.Actual`, `.Reference`, and `.Delta`). So, I check the template string instead 

if u are OK, i will continue to test the actual, reference, and delta cases of `DisabledScientific`

thanks

formatter.go
gavv: > If not set, will print floats in decimal.

I think this is not true. If it's not set, the behavior will be to automatically switch between scientific and decimal forms, like in %g. See details in fmt docs.

formatter.go
```diff
@@ -404,6 +408,15 @@ func (f *DefaultFormatter) formatTyped(value interface{}) string {
 
 func (f *DefaultFormatter) formatValue(value interface{}) string {
 	if isNumber(value) {
+		if f.DisableScientific && isFloat(value) {
+			var ff float64
+			if v, ok := value.(float32); ok {
+				ff = float64(v)
+			} else {
+				ff = value.(float64)
+			}
```
gavv, line 417: Let's create extractFloat() function for this.

formatter.go
gavv: BTW if I'm right let's cover this in tests.

gavv:
> The `AssertionFailure` varies between cases of expected, actual, reference, and delta. Checking the `formatData` is less _generic_ because each "template value" is stored in different field (`.Expected`, `.Actual`, `.Reference`, and `.Delta`). So, I check the template string instead

But we can easily add wantExpected, wantActual, etc? In every test case, we need what field to expect because every test case manually set Expected, Actual, etc.

E.g. instead of:

```

**Comment by neverbeenthisweeb** [comment]
> > The `AssertionFailure` varies between cases of expected, actual, reference, and delta. Checking the `formatData` is less _generic_ because each "template value" is stored in different field (`.Expected`, `.Actual`, `.Reference`, and `.Delta`). So, I check the template string instead
> 
> But we can easily add wantExpected, wantActual, etc? In every test case, we already know what field to expect because every test case manually set Expected, Actual, etc.
> 
> E.g. instead of:
> 
> ```
> 			assertionFailure: AssertionFailure{
> 				Type: AssertInRange,
> 				Expected: &AssertionValue{
> 					Value: AssertionRange{
> 						Min: float32(-1234567.89),
> 						Max: float32(1234567.89),
> 					},
> 				},
> 			},
> 			wantTpl: "[-1234567.875; 1234567.875]",
> ```
> 
> we could write:
> 
> ```
> 			assertionFailure: AssertionFailure{
> 				Type: AssertInRange,
> 				Expected: &AssertionValue{
> 					Value: AssertionRange{
> 						Min: float32(-1234567.89),
> 						Max: float32(1234567.89),
> 					},
> 				},
> 			},
> 			wantExpected: "[-1234567.875; 1234567.875]",
> ```

got it

so, if we want to assert with `wantExpected`, `wantActual`, and `wantReference`, there will be 3 table driven tests (one for each _wantXXX_?

File: formatter.go (hunk 1 above)
Comment (gavv) [reply formatter.go]: Looks legit.

Currently we use %v, which uses %g, which have this in docs:

```
%g	%e for large exponents, %f otherwise. Precision is discussed below.
```

But please re-check and cover in unit test to be 100% sure :)

In our comment, we can say something like this:

* without DisableScientific, we use %g, which uses %e (scientific form) for large exponents and %f (decimal form) otherwise (see documentation for fmt package)
* with DisableScientific, we use %f (decimal form)

Comment (gavv) [reply formatter.go]: BTW, after thinking about this again... :)

It seems even better and clearer API would be to replace boolean with a enum with three values:

* Auto (corresponds to %g)
* Decimal (corresponds to %f)
* Scientific (corresponds to %e)

Where Auto is equal to `0` and thus is used by default.

**Comment by neverbeenthisweeb** [comment]
hi @gavv , just FYI, sry i am a bit slow resp here. Perhaps, will progress on the weekend. Cheers 🥂

**Comment by gavv** [comment]
Cool, no worries

**Comment by neverbeenthisweeb** [comment]
@gavv 

updates:
- Refactor unit test (merge all assertion into one test
- Add extractFloat to extract `float64` from `interface{}`

File: formatter.go [hunk 3]
```diff
@@ -425,6 +432,14 @@ func (f *DefaultFormatter) formatValue(value interface{}) string {
 	return sq.Sdump(value)
 }
 
+func extractFloat(value interface{}) float64 {
+	if v, ok := value.(float32); ok {
+		return float64(v)
+	}
+
+	return value.(float64)
+}
```
gavv, line 441: * Please move it below, near other `extract*` funcs.
* Other extract funcs have default branch, let's add it here too.
* Other extract funcs are switches, could we make this one too?

gavv:
The updated test looks good.

Also it would be nice to cover these cases:

* formatting of AssertionList with number
* formatting of Expected with number that is not AssertionList or AssertionRange
* formatting of Actual with number
* formatting of Reference with number

What else? I guess only https://github.com/gavv/httpexpect/pull/220#discussion_r1068125302 ?

neverbeenthisweeb:
> The updated test looks good.
> 
> Also it would be nice to cover these cases:
> 
> * formatting of AssertionList with number
> * formatting of Expected with number that is not AssertionList or AssertionRange
> * formatting of Actual with number
> * formatting of Reference with number
> 
> What else? I guess only [#220 (comment)](https://github.com/gavv/httpexpect/pull/220#discussion_r1068125302) ?

yes agree

the update is more into an example, will complete the rest afterwards

neverbeenthisweeb:
ah yes, sry i forgot the https://github.com/gavv/httpexpect/pull/220#discussion_r1068125302

i think that idea looks fine

we can have that implemented in this PR

formatter.go
neverbeenthisweeb: ok, will do the change in this PR

neverbeenthisweeb: updated https://github.com/gavv/httpexpect/pull/220/commits/d418ccf01e2e52ddd1893ecf57fef1fe888badac

neverbeenthisweeb:
@gavv 

Updates:

- Replace `DisableScientific` with `FloatFormat` to support decimal, scientific, and auto float formatting (i'm still unsure of the naming, LMK if you have suggestions)
- Update unit test. We tests:
  - auto with small exponent
  - auto with large exponent
  - decimal with large exponent
  - scientific with small exponent

TODO:

- [ ] Complete unit test with other assertion types (e.g. `formatting of AssertionList with number` as mentioned in https://github.com/gavv/httpexpect/pull/220#issuecomment-1399282552)

neverbeenthisweeb:
btw @gavv, for float in decimals why did we prefer:

```
strconv.FormatFloat(*fl, 'f', -1, 64)
```

to?:

```
fmt.Sprintf("%f", value)
```

sry i kinda forget about this one

gavv:
Because strconv variant shows full precision, unlike fmt: https://go.dev/play/p/qFly0dwuIYn

gavv:
Using `%f`, you can specify arbitrary precision, e.g. `%.6f`. However there is now way to ask "use minimum precision needed to represent fractional part full without loss". And strconv.FormatFloat has this option - we can use precision `-1` for that.

formatter.go
```diff
@@ -21,6 +21,20 @@ type Formatter interface {
...
+const (
+	// Print float in scientific notation for large exponent,
+	// otherwise print as decimal
+	FloatFormatAuto FloatFormatMode = iota
+
+	// Print float in decimal
+	FloatFormatDecimal
+
+	// Print float in scientific notation
+	FloatFormatScientific
```
gavv, line 35: Let's refine comments a bit:

```
	// Print floats in scientific notation for large exponents,
	// otherwise print in decimal notation.
	// Similar to %g format.
	FloatFormatAuto FloatFormatMode = iota

	// Always print floats in decimal notation.
	// Similar to %f format.
	FloatFormatDecimal

	// Always print floats in scientific notation.
	// Similar to %e format.
	FloatFormatScientific
```

formatter.go
```diff
@@ -42,6 +56,11 @@ type DefaultFormatter struct {
 	// Exclude diff from failure report.
 	DisableDiffs bool
 
+	// Float printing format.
+	// If not set, will print small exponent in decimal and
+	// large exponent in scientific notation.
+	FloatFormat FloatFormatMode
```
gavv, line 62: Let's refine comment a bit:

```

	// Float printing format.
	// Default is FloatFormatAuto.
	FloatFormat FloatFormatMode
```

neverbeenthisweeb:
> Using `%f`, you can specify arbitrary precision, e.g. `%.6f`. However there is now way to ask "use minimum precision needed to represent fractional part fully without loss". And strconv.FormatFloat has this option - we can use precision `-1` for that.

ok, so is it expected that we use:

- `%g` for auto
- `strconv.FormatFloat` for decimal
- `%e` for scientific?

because it seems a bit misleading, since `%g` fallbacks to `%f` for decimal

but i think it is OK, considering your comments here

https://github.com/gavv/httpexpect/pull/220#discussion_r1085286292

where it says _similar to_

gavv:
Good point. Yes, probably not critical, but I wonder if there is a sane way to implement auto using strconv instead of %f for small exponents.

gavv:
> Complete unit test with other assertion types (e.g. formatting of AssertionList with number as mentioned in Add DefaultFormatter.DisableScientific

If you wish, we can create a separate issue for this. Let me know if you want to create one.

gavv:
> The default precision for %e, %f and %#g is 6; for %g it is the smallest number of digits necessary to identify the value uniquely.

Oh, actually %g already have desired behavior.

So we can just mention in comment for Decimal:

> Similar to %f with precision set to the smallest number of digits necessary to identify the value uniquely

neverbeenthisweeb:
> If you wish, we can create a separate issue for this. Let me know if you want to create one.

@gavv ok, let's separate the issue for this one

i will write a note in the #190 (flexible if u think it should be better with a new issue)

neverbeenthisweeb:
> So we can keep it as is, just mention in comment for Decimal:

ok, will update the comment

neverbeenthisweeb:
this MR is now ready @gavv

gavv:
Thanks!!

I've pushed a few follow-up commits:

* bcef9e54b93aea71597883c993d6d1f43704f471
* 28795cc2b2e2491b67e3011bd8a5a6383c2fa5f7
* d8def20f6e125d1e5a408d0dc1a3c76408e6edbc
* 76350720fd2ee13b97363924fe26a6314ba83d6e
* e7c4cd8c8bfb78f7fcba1ee7897dddd4664e42f4

PR #221, opened 2023-01-02
Extend fruits server example

kairemor:
Extend example add new test for matching functions

- Array.Iter
- Array.Every
- Array.ContainsAny
- Object.ContainsKey
- Object.ContainsValue
- Object.ContainsSubset
- String.HasPrefix and String.HasSuffix
- String.IsASCII

### DIFF
--- a/_examples/fruits_test.go
+++ b/_examples/fruits_test.go
@@ -46,7 +46,7 @@ func TestFruits(t *testing.T) {
 	})
 	fruits.ContainsAny("orange", "melon")
 	fruits.ContainsOnly("orange", "apple")
-	
+
 	e.GET("/fruits/orange").
 		Expect().
 		Status(http.StatusOK).JSON().Object().Equal(orange).NotEqual(apple)
@@ -75,7 +75,7 @@ func TestFruits(t *testing.T) {
 		Expect().
 		Status(http.StatusNotFound)
 
-	apple["image"] = []map[string]string{{"id":" 1", "url":"http://example.com","type": "fruit"}, {"id": "2", "url":"http://example2.com", "type": "fruit"}}
+	apple["image"] = []map[string]string{{"id": " 1", "url": "http://example.com", "type": "fruit"}, {"id": "2", "url": "http://example2.com", "type": "fruit"}}
 
 	e.PUT("/fruits/apple").WithJSON(apple).
 		Expect().
--- a/_examples/fruits_test.go
+++ b/_examples/fruits_test.go
@@ -33,7 +33,6 @@ func TestFruits(t *testing.T) {
 		"weight": 200,
 	}
 
-	
 	e.PUT("/fruits/apple").WithJSON(apple).
 		Expect().
 		Status(http.StatusNoContent).NoContent()
@@ -41,9 +40,9 @@ func TestFruits(t *testing.T) {
 	fruitsResult := e.GET("/fruits").
 		Expect().
 		Status(http.StatusOK).JSON().Array()
-	
+
 	fruitsResult.ContainsOnly("orange", "apple")
-	fruitsResult.ContainsAny("orange","melon")
+	fruitsResult.ContainsAny("orange", "melon")
 	fruitsResult.Every(func(index int, value *httpexpect.Value) {
 		value.String().NotEmpty()
 	})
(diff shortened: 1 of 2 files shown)


_examples/fruits_test.go
```diff
@@ -37,9 +37,15 @@ func TestFruits(t *testing.T) {
 		Expect().
 		Status(http.StatusNoContent).NoContent()
 
-	e.GET("/fruits").
+	fruitsResult := e.GET("/fruits").
```
gavv, line 40: Let's rename this just to "fruits" for clarity.

_examples/fruits_test.go
```diff
@@ -37,9 +37,15 @@ func TestFruits(t *testing.T) {
...
 		Expect().
-		Status(http.StatusOK).JSON().Array().ContainsOnly("orange", "apple")
+		Status(http.StatusOK).JSON().Array()
+
+	fruitsResult.ContainsOnly("orange", "apple")
+	fruitsResult.ContainsAny("orange", "melon")
+	fruitsResult.Every(func(index int, value *httpexpect.Value) {
+		value.String().NotEmpty()
+	})
```
gavv, line 48: I'd reorder this: first call Every (a more generic check - all elements are non-empty strings), then call ContainsXxx (more specific checks).

_examples/fruits_test.go
```diff
@@ -59,10 +65,31 @@ func TestFruits(t *testing.T) {
...
+
+	var fruits fruitMapList
+	apple["type"] = "fruit"
+	orange["type"] = "fruit"
+	fruits = append(fruits, apple)
+	fruits = append(fruits, orange)
+
+	fruitsData := e.GET("/fruits/with-data").WithJSON(fruits).
+		Expect().
+		Status(http.StatusOK).JSON().Array()
```
gavv, line 86: The original fruits server has get/put logic: you put some data and then you can get it.

The new endpoint instead has echo server logic: you send data in body and receive it back.

To keep example simpler, let's use the same logic everywhere. Instead of adding echo server, let's put all necessary data at the beginning of the test, as we already do.

I think we don't need to add new endpoint at all, to get a map. We already have one: `/fruit/<xxx>` will return a map and we can put whatever map we want. So let's just reuse it and put a more complicated fruit and inspect it here.

gavv (commented):
Thanks for pr

_examples/fruits_test.go
kairemor: 👍🏽

_examples/fruits_test.go
gavv: But we can put array of maps into one of the fruits, can't we?

kairemor: ah ok I see you're right will do that

_examples/fruits_test.go
```diff
@@ -59,10 +65,31 @@ func TestFruits(t *testing.T) {
...
 
 	e.GET("/fruits/melon").
 		Expect().
 		Status(http.StatusNotFound)
+
+	apple["image"] = []map[string]string{{"id": " 1", "url": "http://example.com", "type": "fruit"}, {"id": "2", "url": "http://example2.com", "type": "fruit"}}
+
+	e.PUT("/fruits/apple").WithJSON(apple).
+		Expect().
+		Status(http.StatusNoContent).NoContent()
```
gavv, line 82: We do all PUTs in the beginning of the test. Let's move this one there too?

Also, it doesn't need to be a separate PUT. You can modify `apple` declaration at the beginning of the test and include `image` there.

Also, second GET wont be needed, we already have `obj` with retrieved apple.

kairemor: 👍🏽

_examples/fruits_test.go
```diff
@@ -31,15 +31,22 @@ func TestFruits(t *testing.T) {
 	apple := map[string]interface{}{
 		"colors": []interface{}{"green", "red"},
 		"weight": 200,
+		"image":  []map[string]string{{"id": " 1", "url": "http://example.com", "type": "fruit"}, {"id": "2", "url": "http://example2.com", "type": "fruit"}},
```
gavv, line 34: Please use multiline formatting for more readability

_examples/fruits_test.go
```diff
@@ -54,15 +61,27 @@ func TestFruits(t *testing.T) {
...
+		element.Object().ContainsValue("fruit")
+		element.Object().ContainsSubset(map[string]interface{}{
+			"type": "fruit",
+		})
+	}
+
 	e.GET("/fruits/melon").
 		Expect().
 		Status(http.StatusNotFound)
+
```
gavv, line 86: Accidental newline

gavv:
Thanks!

PR #222, opened 2023-01-05
Decode methods

Rohith-Raju:
Signed-off-by: Rohith Raju <55309324+Rohith-Raju@users.noreply.github.com>


Rohith-Raju:
Will redo in a new branch !!

Issue #192, opened 2022-12-05
Implement Decode method


gavv:
Currently almost every struct has Raw() method, that returns the underlying value.

It would be nice to add to each such struct a complementary Decode() method, that will json-decode underlying value into the argument.

Example:

```go
var m map[string]interface{}

m = e.GET("/user").
	Expect().
	Status(http.StatusOK).
	JSON().
	Object().
	Raw()
```

Is same as:

```go
var m map[string]interface{}

e.GET("/user").
	Expect().
	Status(http.StatusOK).
	JSON().
	Object().
	Decode(&m)
```

This method becomes useful when you want to use specific type for decoding, e.g. struct:

```go
var user User

e.GET("/fruits/orange").
	Expect().
	Status(http.StatusOK).
	JSON().
	Object().
	Decode(&user)
```

BTW, this should give the same result:

```go
var user User

e.GET("/fruits/orange").
	Expect().
	Status(http.StatusOK).
	JSON().
	Decode(&user)
```

If argument is nil, the method should report failure (AssertUsage). If decoding fails, the method should also report failure (AssertValid).

The new method should be documented and covered with unit tests. Also it should be covered in e2e test (e2e_basic_test.go).

Also we should add example to README.

Rohith-Raju:
@gavv I'd like to give this a shot

Rohith-Raju:
Hey @gavv, I'll be getting back to this after 5th cause od my exams

gavv:
Great, thanks, no hurry

gavv:
Landed.

PR #223, opened 2023-01-05
Decode methods

Rohith-Raju:
Signed-off-by: Rohith Raju <rohithraju488@gmail.com>
Closes: #192 
- Added canonDecode Method
### Decode for the following structs:
- [x] Array
- [x] Boolean
- [x] Number
- [x] Object
- [x] String
- [ ] Value

### DIFF
--- a/array.go
+++ b/array.go
@@ -67,13 +67,11 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
-func (a *Array) Decode(target *[]interface{}) {
-	a.chain.enter("Decode")
+func (a *Array) Decode(target interface{}) *Array {
+	a.chain.enter("Decode()")
 	defer a.chain.leave()
-
-	if decode := canonDecode(a.chain, a.value, target); !decode {
-		a.chain.failed()
-	}
+	canonDecode(a.chain, a.value, target)
+	return a
 }
 
 // Path is similar to Value.Path.
--- a/array.go
+++ b/array.go
@@ -70,11 +70,9 @@ func (a *Array) Raw() []interface{} {
 // Decode unmarshals the underlying value attached to the Array to a target variable.
 // target should be one of these:
 //
-// 1) pointer to empty interface
+// - pointer to an empty interface
 //
-// 2) pointer to slice of empty interfaces
-//
-// 3) pointer to slice of structs
+// - pointer to slice of any type
 //
 // Example:
 //
--- a/array.go
+++ b/array.go
@@ -67,6 +67,15 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
+func (a *Array) Decode(target *[]interface{}) {
+	a.chain.enter("Decode")
+	defer a.chain.leave()
+
+	if decode := canonDecode(a.chain, a.value, target); !decode {
+		a.chain.failed()
+	}
+}
+
 // Path is similar to Value.Path.
 func (a *Array) Path(path string) *Value {
 	a.chain.enter("Path(%q)", path)
--- a/array.go
+++ b/array.go
@@ -89,7 +89,7 @@ func (a *Array) Raw() []interface{} {
 //				"foo": 456,
 //			},
 //	}
-//	array:=NewArray(t,value)
+//	array := NewArray(t,value)
 //
 //	var target []S
 //	arr.Decode(&target)
--- a/array.go
+++ b/array.go
@@ -90,8 +90,10 @@ func (a *Array) Raw() []interface{} {
 //			},
 //	}
 //	array:=NewArray(t,value)
+//
 //	var target []S
 //	arr.Decode(&target)
+//
 //	assert.Equal(t,[]S{{123},{456}},target)
 func (a *Array) Decode(target interface{}) *Array {
 	a.chain.enter("Decode()")
--- a/array.go
+++ b/array.go
@@ -67,6 +67,15 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
+// Decode binds the underlying value attached to the Array to a Target variable.
+//
+// Example:
+//
+//	var target interface{}
+//	value := []interface{}{"foo",123}
+//	array := NewArray(t,value)
+//	array.Decode(&target)
+//	assert.Equal(t,value,target)
 func (a *Array) Decode(target interface{}) *Array {
 	a.chain.enter("Decode()")
 	defer a.chain.leave()
--- a/array.go
+++ b/array.go
@@ -87,12 +87,12 @@ func (a *Array) Raw() []interface{} {
 //				"foo": 456,
 //			},
 //	}
-//	array := NewArray(t,value)
+//	array := NewArray(t, value)
 //
 //	var target []S
 //	arr.Decode(&target)
 //
-//	assert.Equal(t,[]S{{123},{456}},target)
+//	assert.Equal(t, []S{{123},{456}}, target)
 func (a *Array) Decode(target interface{}) *Array {
 	a.chain.enter("Decode()")
 	defer a.chain.leave()
--- a/array.go
+++ b/array.go
@@ -67,7 +67,7 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
-// Decode binds the underlying value attached to the Array to a Target variable.
+// Decode unmarshals the underlying value attached to the Array to a Target variable.
 //
 // Example:
 //
@@ -79,6 +79,11 @@ func (a *Array) Raw() []interface{} {
 func (a *Array) Decode(target interface{}) *Array {
 	a.chain.enter("Decode()")
 	defer a.chain.leave()
+
+	if a.chain.failed() {
+		return a
+	}
+
 	canonDecode(a.chain, a.value, target)
 	return a
 }
--- a/array.go
+++ b/array.go
@@ -67,15 +67,32 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
-// Decode unmarshals the underlying value attached to the Array to a Target variable.
+// Decode unmarshals the underlying value attached to the Array to a target variable.
+// target should be one of these:
+//
+// 1) pointer to empty interface
+//
+// 2) pointer to slice of empty interfaces
+//
+// 3) pointer to slice of structs
 //
 // Example:
 //
-//	var target interface{}
-//	value := []interface{}{"foo",123}
-//	array := NewArray(t,value)
-//	array.Decode(&target)
-//	assert.Equal(t,value,target)
+//	type S struct{
+//		Foo int `json:foo`
+//	}
+//	value := []interface{}{
+//		map[string]interface{}{
+//				"foo": 123,
+//			},
+//		map[string]interface{}{
+//				"foo": 456,
+//			},
+//	}
+//	array:=NewArray(t,value)
+//	var target []S
+//	arr.Decode(&target)
+//	assert.Equal(t,[]S{{123},{456}},target)
 func (a *Array) Decode(target interface{}) *Array {
 	a.chain.enter("Decode()")
 	defer a.chain.leave()
--- a/array_test.go
+++ b/array_test.go
@@ -117,13 +117,13 @@ func TestArray_Constructors(t *testing.T) {
 
 func TestArray_Decode(t *testing.T) {
 	t.Run("Decode with slice of interface", func(t *testing.T) {
-		target := []interface{}{}
+		var target []interface{}
 		testValue := []interface{}{"Foo", 123.0}
 		reporter := newMockReporter(t)
 		arr := NewArray(reporter, testValue)
 		arr.Decode(&target)
-		arr.chain.assertNotFailed(reporter)
-		arr.Equal(target)
+		arr.chain.assertNotFailed(t)
+		assert.Equal(t, testValue, target)
 	})
 	t.Run("Decode with slice of struct", func(t *testing.T) {
 		reporter := newMockReporter(t)
@@ -142,8 +142,17 @@ func TestArray_Decode(t *testing.T) {
 		var target []S
 		actualStruct := []S{{123}, {456}}
 		arr.Decode(&target)
-		arr.chain.assertFailed(reporter)
-		assert.Equal(reporter, actualStruct, target)
+		arr.chain.assertNotFailed(t)
+		assert.Equal(t, actualStruct, target)
+	})
+	t.Run("Empty interface", func(t *testing.T) {
+		reporter := newMockReporter(t)
+		testValue := []interface{}{"Foo", 123.0}
+		var target interface{}
+		arr := NewArray(reporter, testValue)
+		arr.Decode(&target)
+		arr.chain.assertNotFailed(t)
+		assert.Equal(t, testValue, target)
 	})
 	t.Run("Passing unmarshable value", func(t *testing.T) {
 		reporter := newMockReporter(t)
@@ -166,15 +175,7 @@ func TestArray_Decode(t *testing.T) {
 		arr.Decode(nil)
 		arr.chain.assertFailed(t)
 	})
-	t.Run("Empty interface", func(t *testing.T) {
-		reporter := newMockReporter(t)
-		testValue := []interface{}{"Foo", 123.0}
-		var target interface{}
-		arr := NewArray(reporter, testValue)
-		arr.Decode(&target)
-		arr.chain.assertNotFailed(reporter)
-		assert.Equal(reporter, testValue, target)
-	})
+
 }
 
 func TestArray_Getters(t *testing.T) {
--- a/array_test.go
+++ b/array_test.go
@@ -116,12 +116,58 @@ func TestArray_Constructors(t *testing.T) {
 }
 
 func TestArray_Decode(t *testing.T) {
-	target := []interface{}{}
-	testValue := []interface{}{"Foo", 123.0}
-	reporter := newMockReporter(t)
-	arr := NewArray(reporter, testValue)
-	arr.Decode(&target)
-	assert.Equal(reporter, testValue, target)
+	t.Run("Decode with slice of interface", func(t *testing.T) {
+		target := []interface{}{}
+		testValue := []interface{}{"Foo", 123.0}
+		reporter := newMockReporter(t)
+		arr := NewArray(reporter, testValue)
+		arr.Decode(&target)
+		arr.Equal(target)
+	})
+	t.Run("Decode with slice of struct", func(t *testing.T) {
+		reporter := newMockReporter(t)
+		type S struct {
+			Foo int `json:"foo"`
+		}
+		testValue := []interface{}{
+			map[string]interface{}{
+				"foo": 123,
+			},
+			map[string]interface{}{
+				"foo": 456,
+			},
+		}
+		arr := NewArray(reporter, testValue)
+		target := new([]S)
+		arr.Decode(target)
+		arr.Equal(target)
+	})
+	t.Run("Passing unmarshable value", func(t *testing.T) {
+		reporter := newMockReporter(t)
+		type S struct {
+			Foo int `json:"foo"`
+		}
+		testValue := []interface{}{
+			map[string]interface{}{
+				"foo": 123,
+			},
+			map[string]interface{}{
+				"foo": 456,
+			},
+		}
+		arr := NewArray(reporter, testValue)
+		target := []S{}
+		arr.Decode(target)
+		arr.chain.assertFailed(t)
+	})
+	t.Run("Target is nil", func(t *testing.T) {
+		reporter := newMockReporter(t)
+		testValue := []interface{}{"Foo", 123.0}
+		arr := NewArray(reporter, testValue)
+		arr.Decode(nil)
+		arr.chain.assertFailed(t)
+	})
+
 }
 
 func TestArray_Getters(t *testing.T) {
--- a/array_test.go
+++ b/array_test.go
@@ -115,6 +115,15 @@ func TestArray_Constructors(t *testing.T) {
 	})
 }
 
+func TestArray_Decode(t *testing.T) {
+	target := []interface{}{}
+	testValue := []interface{}{"Foo", 123.0}
+	reporter := newMockReporter(t)
+	arr := NewArray(reporter, testValue)
+	arr.Decode(&target)
+	assert.Equal(reporter, testValue, target)
+}
+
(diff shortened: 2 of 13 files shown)


Rohith-Raju:
@gavv Just to check my bearings, how does this look?

array.go
```diff
@@ -66,6 +66,15 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
+func (a *Array) Decode(target *[]interface{}) {
+	a.chain.enter("Decode")
```
gavv, line 70: Decode -> Decode()

array.go
```diff
@@ -66,6 +66,15 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
+func (a *Array) Decode(target *[]interface{}) {
```
gavv, line 69: * let's return receiver from this method; all httpexpect methods are chainable
* as I mentioned in chat, let's replace pointer to slice of interfaces with just interface; this will allow to pass pointer to slice of any objects, including structs, and current signature allows only slice of interfaces

array.go
```diff
@@ -66,6 +66,15 @@ func (a *Array) Raw() []interface{} {
...
+	a.chain.enter("Decode")
+	defer a.chain.leave()
+
+	if decode := canonDecode(a.chain, a.value, target); !decode {
+		a.chain.failed()
```
gavv, line 74: No need for this call (check its source code). Actually seems we don't need to check return value of canonDecode at all, and it doesn't need to return anything

array_test.go
```diff
@@ -108,6 +108,15 @@ func TestArray_Constructors(t *testing.T) {
 	})
 }
 
+func TestArray_Decode(t *testing.T) {
```
gavv, line 111: It's also important to add a test for decoding into sluce of structs

canon.go
```diff
@@ -91,3 +91,28 @@ func canonValue(chain *chain, in interface{}) (interface{}, bool) {
 
 	return out, true
 }
+
+func canonDecode(chain *chain, value interface{}, target interface{}) bool {
```
gavv, line 95: * we should check if chain is already failed and return in this case
* we should check if target is nil and fail with AssertUsage in this case

gavv (commented):
Thanks, see comments below

array.go
Rohith-Raju: > * as I mentioned in chat, let's replace pointer to slice of interfaces with just interface; this will allow to pass pointer to slice of any objects, including structs, and current signature allows only slice of interfaces
Oh, yes I remember. I did  that you couldn't pass a reference for the target variable with just slice of interface, now changed it to just interface

Rohith-Raju: > let's return receiver from this method; all httpexpect methods are chainable
I don't quite understand this. receiver meaning target?

canon.go
Rohith-Raju: Ok,sure

array.go
gavv: Yes, exactly.

array_test.go
```diff
@@ -108,6 +108,61 @@ func TestArray_Constructors(t *testing.T) {
...
+		testValue := []interface{}{
+			map[string]interface{}{
+				"foo": 123,
+			},
+			map[string]interface{}{
+				"foo": 456,
+			},
+		}
+		arr := NewArray(reporter, testValue)
+		target := new([]S)
```
Rohith-Raju, line 134: @gavv is this fine or use the slice denotation as I did below? I presume we should use one of those to stay consistent. Which one tho?

gavv: You mean whether to use new([]s) or []s{} ?

I guess there is no difference, just a matter of tatste, so up to you.

array_test.go
```diff
@@ -108,6 +108,61 @@ func TestArray_Constructors(t *testing.T) {
...
+		arr.Decode(target)
+		arr.Equal(target)
```
gavv, line 136: Since we use mock reporter, arr.Equal check is not real (never marks test as failed).

Instead, we should use assert.Equal.

Also we should add arr.chain.assertNotFailed right after Decode call, to ensure that Decode succeeded.

array_test.go
gavv: Maybe this way it would be more readable:

var target []S
arr.Decode(&target)

This way the type of target is clear, as well it's clear that we're passing pointer.

Same for the first test:

var target []interface{}
arr.Decode(&target)

Btw we also need third test:

var target interface{}
arr.Decode(&target)

array_test.go
```diff
@@ -108,6 +108,61 @@ func TestArray_Constructors(t *testing.T) {
...
+			map[string]interface{}{
+				"foo": 123,
+			},
+			map[string]interface{}{
+				"foo": 456,
+			},
+		}
+		arr := NewArray(reporter, testValue)
+		target := []S{}
+		arr.Decode(target)
```
gavv, line 153: Decide will fail because target is not a pointer. It's not important whether it's using struct or not.

Hence, to make this test easier to understand, let's not use struct here. We can just do something minimal like:

arr.Decode(123)

array_test.go
Rohith-Raju: > Since we use mock reporter, arr.Equal check is not real (never marks test as failed).
   Instead, we should use assert.Equal. 

Yes, makes Sense. I actually tried this and wanted to clarify somethings..

So if we use assert.Equal , the types doesn't match because _value_ is map and _target_ is struct. So I tried comparing the values. I extracted the value.element(0).string().raw() data, which works fine. How do I extract the target data? Or is there another way?

gavv: You could just create a variable of needed struct type snd value and compare target with that.

Rohith-Raju:
@gavv looking good?

canon.go
```diff
@@ -91,3 +91,38 @@ func canonValue(chain *chain, in interface{}) (interface{}, bool) {
...
+
+	if target == nil {
+		chain.fail(AssertionFailure{
+			Type:   AssertNil,
+			Actual: &AssertionValue{target},
+			Errors: []error{
+				errors.New("expected: target should not be nil"),
+			},
+		})
+	}
```
gavv, line 108: Missing return here

canon.go
```diff
@@ -91,3 +91,38 @@ func canonValue(chain *chain, in interface{}) (interface{}, bool) {
...
+	if chain.failed() {
+		return
+	}
+
+	if target == nil {
+		chain.fail(AssertionFailure{
+			Type:   AssertNil,
+			Actual: &AssertionValue{target},
+			Errors: []error{
+				errors.New("expected: target should not be nil"),
```
gavv, line 105: Let's reword this: `expected: target should not be nil` -> `unexpected nil target argument` (we use same wording in other places)

canon.go
```diff
@@ -91,3 +91,38 @@ func canonValue(chain *chain, in interface{}) (interface{}, bool) {
...
 }
+
+func canonDecode(chain *chain, value interface{}, target interface{}) {
+	if chain.failed() {
+		return
+	}
+
+	if target == nil {
+		chain.fail(AssertionFailure{
+			Type:   AssertNil,
```
gavv, line 102: AssertNil -> AssertUsage (we use this kind of assertion for cases when the lib is used incorrectly, like here)

canon.go
```diff
@@ -91,3 +91,38 @@ func canonValue(chain *chain, in interface{}) (interface{}, bool) {
...
+		})
+	}
+	if err := json.Unmarshal(b, target); err != nil {
+		chain.fail(AssertionFailure{
+			Type:   AssertValid,
+			Actual: &AssertionValue{target},
+			Errors: []error{
+				errors.New("expected: unmarshable value"),
+			},
+		})
```
gavv, line 126: Let's reword this: `expected: value can be unmarshaled into target argument`. Instead of AssertValid, let's use AssertType, it will include target type in error message.

array_test.go
```diff
@@ -115,6 +115,68 @@ func TestArray_Constructors(t *testing.T) {
 	})
 }
 
+func TestArray_Decode(t *testing.T) {
+	t.Run("Decode with slice of interface", func(t *testing.T) {
+		target := []interface{}{}
```
gavv, line 120: Nitpick: for consistency let's use `var target []interface{}`

array_test.go
```diff
@@ -115,6 +115,68 @@ func TestArray_Constructors(t *testing.T) {
 	})
 }
 
+func TestArray_Decode(t *testing.T) {
+	t.Run("Decode with slice of interface", func(t *testing.T) {
```
gavv, line 119: Nitpick: let's reorder subtests a bit: first all successful checks, then all failure checks.

array_test.go
```diff
@@ -115,6 +115,68 @@ func TestArray_Constructors(t *testing.T) {
...
+		target := []interface{}{}
+		testValue := []interface{}{"Foo", 123.0}
+		reporter := newMockReporter(t)
+		arr := NewArray(reporter, testValue)
+		arr.Decode(&target)
+		arr.chain.assertNotFailed(reporter)
+		arr.Equal(target)
```
gavv, line 126: We should use assert.Equal here.

array_test.go
```diff
@@ -115,6 +115,68 @@ func TestArray_Constructors(t *testing.T) {
...
+			map[string]interface{}{
+				"foo": 456,
+			},
+		}
+		arr := NewArray(reporter, testValue)
+		var target []S
+		actualStruct := []S{{123}, {456}}
+		arr.Decode(&target)
+		arr.chain.assertFailed(reporter)
+		assert.Equal(reporter, actualStruct, target)
```
gavv, line 146: It should be `assert.Equal(t`, not `assert.Equal(reporter`, because reporter is mock.

array_test.go
```diff
@@ -115,6 +115,68 @@ func TestArray_Constructors(t *testing.T) {
...
+		arr.chain.assertFailed(t)
+	})
+	t.Run("Empty interface", func(t *testing.T) {
+		reporter := newMockReporter(t)
+		testValue := []interface{}{"Foo", 123.0}
+		var target interface{}
+		arr := NewArray(reporter, testValue)
+		arr.Decode(&target)
+		arr.chain.assertNotFailed(reporter)
+		assert.Equal(reporter, testValue, target)
```
gavv, line 176: It should be `assert.Equal(t`, not `assert.Equal(reporter`, because reporter is mock.

canon.go
Rohith-Raju: Oh ok, when do we use AssertNil then?

array_test.go
Rohith-Raju: oops, I was experimenting with something and forgot to change it back.

canon.go
Rohith-Raju:
```
FAIL: TestArray_Decode/Target_is_nil (0.00s)
panic: AssertionFailure of type AssertUsage can't have Actual field [recovered]
        panic: AssertionFailure of type AssertUsage can't have Actual field
```

I'm guessing we can't use AssetUsage without an actual value, which in our case is nil.

Rohith-Raju:
@gavv if good, I'll go ahead with adding docs and moving to other methods

canon.go
gavv: Usually we use them like this:

* if user asks that value should be nil: AssertNil
* if user asks that value should not be nil: AssertNotNil
* if user incorrectly uses the library: AssertUsage

In this case, the *user* does not want to check `target` for null, the user just requests some operation that involves `target`. And it's *our* requirement that the `target` is not null, and if it is, we want to tell the user that the library is being used incorrectly and we can't fulfill the user request.

gavv: This is of course just a convention and AssertNotNil would work too. But use AssertUsage in similar places when the user provides some incorrect arguments, and it's handy that all such places can be easily identified by AssertUsage.

array_test.go
gavv: Yes, exactly, good catch.

array_test.go
```diff
@@ -115,6 +115,69 @@ func TestArray_Constructors(t *testing.T) {
 	})
 }
 
+func TestArray_Decode(t *testing.T) {
+	t.Run("Decode with slice of interface", func(t *testing.T) {
```
gavv, line 119: Let's rename subtests for more consistency & readability:

```
Decode into slice of empty interfaces
Decode into slice of structs
Decode into empty interface
Target is unmarshable
Target is nil
```

canon.go
Rohith-Raju: Ahh, I see!!.

array_test.go
```diff
@@ -115,6 +115,69 @@ func TestArray_Constructors(t *testing.T) {
...
+		var target []interface{}
+		testValue := []interface{}{"Foo", 123.0}
+		reporter := newMockReporter(t)
+		arr := NewArray(reporter, testValue)
+		arr.Decode(&target)
+		arr.chain.assertNotFailed(t)
+		assert.Equal(t, testValue, target)
```
gavv, line 126: For readability, let's use the same order in all subtest, and separate them into a few blocks, e.g.:

```go

		testValue := []interface{}{"Foo", 123.0}
		arr := NewArray(reporter, testValue)

		var target []interface{}
		arr.Decode(&target)

		arr.chain.assertNotFailed(t)
		assert.Equal(t, testValue, target)
```

the blocks are:

* declare reporter
* declare test value and array
* declare target and decode into it
* make assertions

canon.go
```diff
@@ -91,3 +91,38 @@ func canonValue(chain *chain, in interface{}) (interface{}, bool) {
...
+	if err := json.Unmarshal(b, target); err != nil {
+		chain.fail(AssertionFailure{
+			Type:   AssertValid,
+			Actual: &AssertionValue{target},
+			Errors: []error{
+				errors.New("expected: value can be unmarshaled into target argument"),
+			},
+		})
+	}
+}
```
gavv, line 128: Please also add return after each chain.fail call.

Also, I'd add a newline before json.Marshal, to separate initial checks from the rest of the body.

gavv (commented):
Thanks, a couple more comments. Feel free to proceed with docs and other structs.

gavv:
LGTM

gavv:
Regarding docs, ideally code example should be with struct, since it's the main motivation for Decode.

Also, instead of "binds" I'd use term "unmarshals" because we rely on json.Unmarsal here and in particular user has to use json tags in structs.

Rohith-Raju:
@gavv Yes, understood.

gavv:
One more nitpick.. Usually chain.failed checks are in struct methods, not in canon funcs. Let's use the same approach for Decode too.

That's because all struct methods has similar preamble: call to chain.enrer/leave, and check for chain.failed.

Rohith-Raju:
I'll have to change the docs to use struct

Rohith-Raju:
> Signed-off-by: Rohith Raju [rohithraju488@gmail.com](mailto:rohithraju488@gmail.com) Closes: #192
> 
> * Added canonDecode Method
> 
> ### Decode for the following structs:
> * [x]  Array
> * [ ]  Boolean
> * [ ]  Cookie
> * [ ]  DateTime
> * [ ]  Duration
> * [ ]  Number
> * [ ]  Object
> * [ ]  String
> * [ ]  Value

@gavv  btw does it make sense to add decode for boolean ?

canon.go
```diff
@@ -91,3 +91,39 @@ func canonValue(chain *chain, in interface{}) (interface{}, bool) {
 
 	return out, true
 }
+
+func canonDecode(chain *chain, value interface{}, target interface{}) {
+
+	if target == nil {
```
gavv, line 97: Nitpick: extra newline.

gavv:
> btw does it make sense to add decode for boolean

I think yes. For consistency ideally we should add Decode to all json types.

> also am I missing some structs?

Nope.

But it seems we should NOT add Decode to non-json structs: cookie, datetime, duration. Unmarshaling into a struct does not make sense for them.

canon.go
Rohith-Raju: oops, my bad

gavv:
Also please add Decode call to TestArray_Failed (same for other structs).

gavv:
Also please add test for canonDecode to canon_test.go

Rohith-Raju:
> Also please add Decode call to TestArray_Failed (same for other structs).
```
t.Run("failed_chain_Decode", func(t *testing.T) {
		chain := newMockChain(t)
		chain.fail(mockFailure())

		value := []interface{}{123}
		arr := newArray(chain, value)

		var target []interface{}

		check(arr.Decode(target))
	})
```
@gavv Is this fine?

Rohith-Raju:
> Also please add test for canonDecode to canon_test.go

@gavv what tests should I add for canonDecode? Pretty much the same tests I added for arrayDecode right?

array.go
```diff
@@ -67,6 +67,44 @@ func (a *Array) Raw() []interface{} {
...
 }
 
+// Decode unmarshals the underlying value attached to the Array to a target variable.
+// target should be one of these:
+//
+// 1) pointer to empty interface
+//
+// 2) pointer to slice of empty interfaces
+//
+// 3) pointer to slice of structs
```
gavv, line 77: Actually this list is is incomplete. It can be slice of any type into which array values can be unmrashalled. E.g. https://go.dev/play/p/p64my6HNdXi

gavv:
> > Also please add Decode call to TestArray_Failed (same for other structs).
> 
> ```
> t.Run("failed_chain_Decode", func(t *testing.T) {
> 		chain := newMockChain(t)
> 		chain.fail(mockFailure())
> 
> 		value := []interface{}{123}
> 		arr := newArray(chain, value)
> 
> 		var target []interface{}
> 
> 		check(arr.Decode(target))
> 	})
> ```
> 
> @gavv Is this fine?

No. The new check should be integrated into `check` function, because the idea of that unit test it to run similar checks several times of objects prepared differently.

gavv:
> @gavv what tests should I add for canonDecode? Pretty much the same tests I added for arrayDecode right?

Oh, you're right, it will be complete duplication of Array tests. Let's not do it then.

array.go
```diff
@@ -67,6 +67,44 @@ func (a *Array) Raw() []interface{} {
...
+//				"foo": 123,
+//			},
+//		map[string]interface{}{
+//				"foo": 456,
+//			},
+//	}
+//	array:=NewArray(t,value)
+//	var target []S
+//	arr.Decode(&target)
+//	assert.Equal(t,[]S{{123},{456}},target)
```
gavv, line 95: Nitpick: missing some spaces

array_test.go
```diff
@@ -115,6 +127,78 @@ func TestArray_Constructors(t *testing.T) {
...
+
+		testValue := []interface{}{"Foo", 123.0}
+		arr := NewArray(reporter, testValue)
+
+		arr.Decode(nil)
+
+		arr.chain.assertFailed(t)
+	})
+
+}
```
gavv, line 200: Nitpick: extra newline.

array.go
Rohith-Raju: Yes, that makes sense. So now, do I just add "pointer to slice of any type"?

gavv: Also, could please change list formatting a bit, here and in other methods? To match the rest of the lib.

```
// target should be one of these:
//  - pointer to an empty interface
//  - pointer to a slice of any type
```

boolean.go
```diff
@@ -49,6 +49,33 @@ func (b *Boolean) Raw() bool {
 	return b.value
 }
 
+// Decode unmarshals the underlying value attached to the Boolean to a target variable.
+// target should be one of these:
+//
+// 1) pointer to a empty interface.
```
gavv, line 55: `a empty` -> `an empty`

number.go
```diff
@@ -50,6 +50,35 @@ func (n *Number) Raw() float64 {
...
 }
 
+// Decode unmarshals the underlying value attached to the Number to a target variable.
+// target should be one of these:
+//
+// 1) pointer to empty interface.
+//
+// 2) pointer to int variable.
+//
+// 3) pointer to float64 variable.
```
gavv, line 60: It can be any integer or floating type: https://go.dev/play/p/_GNh1u40BQX

number_test.go
```diff
@@ -53,6 +53,63 @@ func TestNumber_Constructors(t *testing.T) {
 	})
 }
 
+func TestNumber_Decode(t *testing.T) {
```
gavv, line 56: Don't forget to add Decode call to TestNumber_Failed and TestBoolean_Failed.

gavv:
Coverage decreased a bit because we added new uncovered branch in canonDecode: https://coveralls.io/builds/56167786/source?filename=canon.go

This specific branch can't be covered by tests for Array/Object/etc because they always have correct value.

So let's indeed add small test for canonDecode, only for error handling. I'd cover three errors: target is nil, value is not marshallable, value is not unmrashallable into target.

Rohith-Raju:
> Coverage decreased a bit because we added new uncovered branch in canonDecode: https://coveralls.io/builds/56167786/source?filename=canon.go
> 
> This specific branch can't be covered by tests for Array/Object/etc because they always have correct value.
> 
> So let's indeed add small test for canonDecode, only for error handling. I'd cover three errors: target is nil, value is not marshallable, value is not unmrashallable into target.

Yes, sure.

Rohith-Raju:
While I'm at it do I cover it for Cannon value too ?
Screen Shot 2023-01-23 at 1 44 07 PM

gavv:
> While I'm at it, do I cover it for cannonValue too?

It would be awesome if you do it, though I'm not sure it's possible :)

Rohith-Raju:
> > While I'm at it, do I cover it for cannonValue too?
> 
> It would be awesome if you do it, though I'm not sure it's possible :)

Oh yes, Not quite possible 😅

gavv:
Hi, I need to merge some commits into master which rework chain interface and affect almost every usage of it. I don't want to bother you with resolving conflicts, so I'll merge this PR first, as it's already good in its current state.

After merging this PR, I'll merge my changes too, resolve conflicts, and then you can create a separate PR for Value.Decode.

gavv:
Done.

I've pushed two follow-up commits:

* 1623932e79bafe5a3e6d87623a14611851606b4d - minor fixes in formatting
* 79c152975ec143b91bfc22eaa61868334d983740 - use new chain interface in Decode() 

Please use the same approach when you'll be adding Value.Decode.

gavv:
BTW it would be awesome if you also can add usage example for Decode to README.

Rohith-Raju:
> BTW it would be awesome if you also can add usage example for Decode to README.

Yes, will Do!!

PR #224, opened 2023-01-05
Add unit tests for chain cloning

kairemor:
Adding unit tests for constructors of all matcher structs (newValue, newArray, etc).

- chain field of constructed object point to different address
- chain of constructed object has same path


gavv:
Thanks for pr!

Instead of NotEqual with unsafe, I think we can just use NotSame?

gavv:
Thanks for update. Could you please downgrade testify version to 1.5.0? It's the first version that has NotSame, according  to https://pkg.go.dev/github.com/stretchr/testify@v1.8.1/assert#NotSame

In this lib, I'm trying to keep dependency requirements as low as possible, because users may rely on older versions in some cases.

gavv:
LGTM!

PR #225, opened 2023-01-09
Add unit tests for validateTraits() function

kairemor:
unit test for validateTraits

### DIFF
--- a/assertion_test.go
+++ b/assertion_test.go
@@ -195,7 +195,205 @@ func TestAssertion_HandlerPanics(t *testing.T) {
 	})
 }
 
-func TestAssertion_Validation(t *testing.T) {
+func TestAssertion_ValidateTraits(t *testing.T) {
+	tests := []struct {
+		testName          string
+		errorContainsText string
+		failure           AssertionFailure
+		traits            fieldTraits
+	}{
+		{
+			testName:          "bad Type",
+			errorContainsText: "AssertionType",
+			failure: AssertionFailure{
+				Type: AssertionType(9999),
+			},
+			traits: fieldTraits{
+				Actual: fieldRequired,
+			},
+		},
+		{
+			testName:          "required Actual",
+			errorContainsText: "Actual",
+			failure: AssertionFailure{
+				Actual: nil,
+			},
+			traits: fieldTraits{
+				Actual: fieldRequired,
+			},
+		},
+		{
+			testName:          "denied Actual",
+			errorContainsText: "Actual",
+			failure: AssertionFailure{
+				Actual: &AssertionValue{},
+			},
+			traits: fieldTraits{
+				Actual: fieldDenied,
+			},
+		},
+		{
+			testName:          "required Expected",
+			errorContainsText: "Expected",
+			failure: AssertionFailure{
+				Expected: nil,
+			},
+			traits: fieldTraits{
+				Expected: fieldRequired,
+			},
+		},
+		{
+			testName:          "denied Expected",
+			errorContainsText: "Expected",
+			failure: AssertionFailure{
+				Expected: &AssertionValue{},
+			},
+			traits: fieldTraits{
+				Expected: fieldDenied,
+			},
+		},
+		{
+			testName:          "required on Expected",
+			errorContainsText: "",
+			failure: AssertionFailure{
+				Expected: nil,
+			},
+			traits: fieldTraits{
+				Range: fieldRequired,
+			},
+		},
+		{
+			testName:          "AssertionRange should not be pointer",
+			errorContainsText: "AssertionRange",
+			failure: AssertionFailure{
+				Expected: &AssertionValue{
+					Value: &AssertionRange{},
+				},
+			},
+			traits: fieldTraits{
+				Range: fieldRequired,
+			},
+		},
+		{
+			testName:          "Min should be non-nil",
+			errorContainsText: "AssertionRange",
+			failure: AssertionFailure{
+				Expected: &AssertionValue{
+					Value: AssertionRange{
+						Max: 1,
+					},
+				},
+			},
+			traits: fieldTraits{
+				Range: fieldRequired,
+			},
+		},
+		{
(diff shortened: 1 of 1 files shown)


kairemor:
ready for review

assertion_test.go
```diff
@@ -477,3 +477,183 @@ func TestAssertion_Strings(t *testing.T) {
 		}
 	})
 }
+
+func TestAssertion_ValidateTraits(t *testing.T) {
```
gavv, line 481: Let's move this test right above TestAssertion_Validation. And also let's rename TestAssertion_Validation to TestAssertion_ValidateAssertion.

assertion_test.go
```diff
@@ -477,3 +477,183 @@ func TestAssertion_Strings(t *testing.T) {
...
+			failure: AssertionFailure{
+				Type: AssertionType(9999),
+			},
+			traits: fieldTraits{
+				Actual: fieldRequired,
+			},
+		},
+		{
+			testName:          "required Actual",
+			errorContainsText: "should have Actual",
```
gavv, line 500: Here and below, let's relax requirements for error messages as much as possible. Here I suggest to just check that the error message contains "Actual".

Otherwise we'll have to fix all tests whenever we change wording of errors.

assertion_test.go
```diff
@@ -477,3 +477,183 @@ func TestAssertion_Strings(t *testing.T) {
...
+			errorContainsText: "can't have Expected",
+			failure: AssertionFailure{
+				Expected: &AssertionValue{},
+			},
+			traits: fieldTraits{
+				Expected: fieldDenied,
+			},
+		},
+		{
+			testName:          "Required value on Expected to be pointer to AssertionRange",
```
gavv, line 539: Here and below, let's simplify test name a bit, e.g. we can call this one just "AssertionRange should not be pointer", and the next one could be just "Min should be non-nil", etc.

assertion_test.go
```diff
@@ -477,3 +477,183 @@ func TestAssertion_Strings(t *testing.T) {
...
+				Expected: &AssertionValue{
+					Value: AssertionList{},
+				},
+			},
+			traits: fieldTraits{
+				List: fieldRequired,
+			},
+		},
+		{
+			testName:          "Required Expected AssertionList with values not being a slice",
```
gavv, line 615: Let's rename this to "Expected can't be slice of single element".

assertion_test.go
```diff
@@ -477,3 +477,183 @@ func TestAssertion_Strings(t *testing.T) {
...
+				Expected: &AssertionValue{
+					Value: AssertionList{[]int{1}},
+				},
+			},
+			traits: fieldTraits{
+				List: fieldRequired,
+			},
+		},
+		{
+			testName:          "Required Expected",
```
gavv, line 627: Let's rename this to "required List", because we already have "required Expected" above.

assertion_test.go
```diff
@@ -477,3 +477,183 @@ func TestAssertion_Strings(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.testName, func(t *testing.T) {
+			err := validateTraits(&test.failure, test.traits)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), test.errorContainsText)
+		})
+	}
+
+	t.Run("nil Errors", func(t *testing.T) {
```
gavv, line 646: Let's rename this to "no error" to avoid confusion with "nil Errors" test in TestAssertion_Validation.

gavv (commented):
Thanks! Sorry for delay.

See a few comments below.

gavv:
We still have two uncovered branches: https://coveralls.io/builds/55772691/source?filename=assertion_validation.go

(see coveralls comment above)

Could you take a look, maybe you could cover them too?

assertion_test.go
```diff
@@ -195,7 +195,205 @@ func TestAssertion_HandlerPanics(t *testing.T) {
...
+			failure: AssertionFailure{
+				Expected: &AssertionValue{},
+			},
+			traits: fieldTraits{
+				Expected: fieldDenied,
+			},
+		},
+		{
+			testName:          "required on Expected",
+			errorContainsText: "",
```
gavv, line 257: Let's rename this to `required AssertionRange`.

Also errorContainsText can be set to "Expected".

assertion_test.go
```diff
@@ -195,7 +195,205 @@ func TestAssertion_HandlerPanics(t *testing.T) {
...
+			errorContainsText: "AssertionList",
+			failure: AssertionFailure{
+				Expected: &AssertionValue{
+					Value: &AssertionList{},
+				},
+			},
+			traits: fieldTraits{
+				List: fieldRequired,
+			},
+		},
```
gavv, line 328: Seems this test duplicates previous one and can be removed. Also its name is misleading.

assertion_test.go
```diff
@@ -195,7 +195,205 @@ func TestAssertion_HandlerPanics(t *testing.T) {
...
+					Value: AssertionList{},
+				},
+			},
+			traits: fieldTraits{
+				List: fieldRequired,
+			},
+		},
+		{
+			testName:          "Expected can't be slice of single element",
+			errorContainsText: "",
```
gavv, line 343: Let's rename this to `AssertionList can't be slice of single element` for consistency.

Also, errorContainsText can be set to "AssertionList".

assertion_test.go
```diff
@@ -195,7 +195,205 @@ func TestAssertion_HandlerPanics(t *testing.T) {
...
+					Value: AssertionList{[]int{1}},
+				},
+			},
+			traits: fieldTraits{
+				List: fieldRequired,
+			},
+		},
+		{
+			testName:          "required List",
+			errorContainsText: "",
```
gavv, line 355: Let's rename this to `required AssertionList` for consistency.

Also, errorContainsText can be set to "Expected".

gavv:
Thanks!

PR #226, opened 2023-01-09
Add noCopy protection 

kairemor:
add no copy struct and as first field of matcher and where we have chain as field like Expect, Request, structs (Value, Object, Array, etc..

### DIFF
--- a/array.go
+++ b/array.go
@@ -9,7 +9,7 @@ import (
 // Array provides methods to inspect attached []interface{} object
 // (Go representation of JSON array).
 type Array struct {
-	noCopy *noCopy
+	noCopy noCopy
 	chain  *chain
 	value  []interface{}
 }
@@ -39,7 +39,7 @@ func NewArrayC(config Config, value []interface{}) *Array {
 }
 
 func newArray(parent *chain, val []interface{}) *Array {
-	a := &Array{noCopy: &noCopy{}, chain: parent.clone(), value: nil}
+	a := &Array{noCopy: noCopy{}, chain: parent.clone(), value: nil}
 	if val == nil {
 		a.chain.fail(AssertionFailure{
 			Type:   AssertNotNil,
--- a/array.go
+++ b/array.go
@@ -39,7 +39,8 @@ func NewArrayC(config Config, value []interface{}) *Array {
 }
 
 func newArray(parent *chain, val []interface{}) *Array {
-	a := &Array{noCopy: noCopy{}, chain: parent.clone(), value: nil}
+	a := &Array{chain: parent.clone(), value: nil}
+
 	if val == nil {
 		a.chain.fail(AssertionFailure{
 			Type:   AssertNotNil,
--- a/array.go
+++ b/array.go
@@ -39,8 +39,7 @@ func NewArrayC(config Config, value []interface{}) *Array {
 }
 
 func newArray(parent *chain, val []interface{}) *Array {
-	a := &Array{parent.clone(), nil}
-
+	a := &Array{noCopy: &noCopy{}, chain: parent.clone(), value: nil}
 	if val == nil {
 		a.chain.fail(AssertionFailure{
 			Type:   AssertNotNil,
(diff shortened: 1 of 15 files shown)


gavv:
Adding a pointer to noCopy does not make sense actually. See details on how it works here:

* https://github.com/valyala/fasthttp/blob/master/nocopy.go
* https://github.com/golang/go/issues/8005#issuecomment-190753527
* https://stackoverflow.com/questions/52494458/nocopy-minimal-example

gavv:
Also, I just realized that there are several places where we return `[]Value` slice, which may involve copying of `Value` structs. Ideally we should replace it with `[]*Value`, however this will break compatibility. So, for now let's add noCopy to everything except Value, and fix Value in 3.0, when we'll be able to break compatibility.

gavv:
Also we need a good comment for noCopy, explaining how it works. It can be based on comment from fasthttp.

gavv:
FindAll was not included in tagged releases yet, so I fixed signature for it: acd1cefa26ef6af06be140512ad70dff763c52a3

Another remaining place is Array.Iter / Object.Iter, which we will fix in v3.

kairemor:
ready for review

array.go
```diff
@@ -38,8 +39,7 @@ func NewArrayC(config Config, value []interface{}) *Array {
 }
 
 func newArray(parent *chain, val []interface{}) *Array {
-	a := &Array{parent.clone(), nil}
-
+	a := &Array{noCopy: noCopy{}, chain: parent.clone(), value: nil}
```
gavv, line 42: Nitpick: a newline was accidentally removed I guess.

boolean.go
```diff
@@ -34,7 +35,7 @@ func NewBooleanC(config Config, value bool) *Boolean {
 }
 
 func newBoolean(parent *chain, val bool) *Boolean {
-	return &Boolean{parent.clone(), val}
+	return &Boolean{noCopy: noCopy{}, chain: parent.clone(), value: val}
```
gavv, line 38: We don't need this `noCopy: noCopy{}` everywhere, I suggest to remove it.

cookie.go
```diff
@@ -8,8 +8,9 @@ import (
 
 // Cookie provides methods to inspect attached http.Cookie value.
 type Cookie struct {
-	chain *chain
-	value *http.Cookie
+	chain  *chain
+	value  *http.Cookie
+	noCopy noCopy
```
gavv, line 13: Let's make noCopy to be the first field in all structs, for consistency.

expect.go
```diff
@@ -88,6 +88,7 @@ import (
 // Expect is a toplevel object that contains user Config and allows
 // to construct Request objects.
 type Expect struct {
+	noCopy   *noCopy
```
gavv, line 91: It should be a pointer.

nocopy.go
```diff
@@ -0,0 +1,13 @@
+package httpexpect
+
+// noCopy struct is a special type that is used to prevent a value from being copied.
+// so `go vet` gives a warning if this struct is copied.
+// It has two methods, `Lock` and `Unlock`, that do not do anything.
+// The purpose of the `noCopy` struct is to prevent a value that has
+// it as a field from being copied.
+// Every matcher struct, e.g. Value, Object, Array, etc. contains a chain instance.
```
gavv, line 8: Let's cleanup this comment a bit:

```
// noCopy struct is a special type that is used to prevent a value from being copied.
// `go vet` gives a warning if it finds that a struct with a field of this type is being copied.
// To enable this behavior, this struct provides two methods `Lock` and `Unlock, that do not do anything.
// See more details here:
//   https://github.com/golang/go/issues/8005
//   https://stackoverflow.com/questions/52494458
```

expect.go
kairemor: here there is somewhere the value of the Expect is copied go in expect.go line 435
What can we do ?

gavv: Oh, good catch.
This is a legit use though, because it does deep copy there. So you can just add `// no lint` to that line.

kairemor: done

gavv:
Thanks!

Issue #171, opened 2022-11-24
Add support for aliases


gavv:
httpexpect is able to print assertion chain path on failure. For example, when you write:

```go
e.GET("/endpoint").
  Expect().
  JSON().Object().Value("foo").
  Array().Element(0).Number()
```

and assertion fails, you will see:

```
        expected: value is number
        
        assertion:
          Request("GET").Expect().JSON().Object().Value("foo").Array().
            Element(0).Number()
```

Often it is convenient to create variables when making assertions:

```go
foo := e.GET("/endpoint").
  Expect().
  JSON().Object().Value("foo").Array()

foo.Element(0).Number()
foo.Element(1).Number()
...
```

In this case, it would be nice if variables names were reflected in the failure message.

This issue suggest to implement support for **aliases**, which will work like this:

```go
foo := e.GET("/endpoint").
  Expect().
  JSON().Object().Value("foo").Array().Alias("foo")

foo.Element(0).Number()
```

```
        expected: value is number
        
        assertion:
          foo.Element(0).Number()
```

To make it possible, the following changes are needed:

* Add `AssertionContext.AliasedPath`. It will be similar to `AssertionContext.Path`, but will take into account aliases.
* `chain.enter()` and `chain.leave()` should update both `Path` and `AliasedPath`
* `chain.clone()` should clone both `Path` and `AliasedPath`
* `chain` will provide new `setAlias(name)` method that will replace `AliasedPath` with a new slice with single element `name`, and keep `Path` untouched
* To every inspector struct (Array, Object, Number, etc), add `Alias()` method that accepts a string name and returns receiver; it will just call `chain.setAlias`.

After this changes, `AssertionContext.AliasedPath` will automatically contain the value we're interested in. Now we just need to update `DefaultFormatter`:

* add `DisableAliases` flag (zero by default), which disables new feature (enabled by default)
* if aliases are enabled, fill `FormatData.AssertPath` using `AliasedPath` instead of `Path`

And that's it.

We also need to add unit tests:

* comprehensive tests for alias support in `chain`
* short unit test (or maybe new assertion in existing test) for every inspector object (Array, Object, etc)
* e2e test for aliases in reports (see #165)

Finally we should add documentation:

* added public methods should be documented
* we should add a section to README that explains the feature

k4n4ry:
Hi @gavv, can I pick this up?

gavv:
You're welcome, thanks!

PR #227, opened 2023-01-15
implement alias support

k4n4ry:
close https://github.com/gavv/httpexpect/issues/171

* Inspector structs which added Alias() are follows:
  * Array
  * Boolean
  * Cookie
  * DateTime
  * Duration
  * Match
  * Number
  * Object
  * Response
  * String
  * Value
  * WebSocketMessage
  * WebSocket

### DIFF
--- a/array.go
+++ b/array.go
@@ -70,6 +70,15 @@ func (a *Array) Raw() []interface{} {
 	return a.value
 }
 
+// Alias is similar to Value.Alias.
+func (a *Array) Alias(name string) *Array {
+	opChain := a.chain.enter("Alias(%s)", name)
+	defer opChain.leave()
+
+	a.chain.setAlias(name)
+	return a
+}
+
 // Decode unmarshals the underlying value attached to the Array to a target variable.
 // target should be one of these:
 //
--- a/chain.go
+++ b/chain.go
@@ -113,8 +113,10 @@ func newChainWithConfig(name string, config Config) *chain {
 
 	if name != "" {
 		c.context.Path = []string{name}
+		c.context.AliasedPath = []string{name}
 	} else {
 		c.context.Path = []string{}
+		c.context.AliasedPath = []string{}
 	}
 
 	if config.Environment != nil {
@@ -143,8 +145,10 @@ func newChainWithDefaults(name string, reporter Reporter) *chain {
 
 	if name != "" {
 		c.context.Path = []string{name}
+		c.context.AliasedPath = []string{name}
 	} else {
 		c.context.Path = []string{}
+		c.context.AliasedPath = []string{}
 	}
 
 	c.context.Environment = newEnvironment(c)
@@ -248,6 +252,7 @@ func (c *chain) clone() *chain {
 
 	contextCopy := c.context
 	contextCopy.Path = append(([]string)(nil), contextCopy.Path...)
+	contextCopy.AliasedPath = append(([]string)(nil), c.context.AliasedPath...)
 
 	return &chain{
 		parent: c,
@@ -270,6 +275,8 @@ func (c *chain) enter(name string, args ...interface{}) *chain {
 	chainCopy.state = stateEntered
 	if name != "" {
 		chainCopy.context.Path = append(chainCopy.context.Path, fmt.Sprintf(name, args...))
+		chainCopy.context.AliasedPath =
+			append(c.context.AliasedPath, fmt.Sprintf(name, args...))
 	}
 
 	return chainCopy
@@ -289,6 +296,9 @@ func (c *chain) replace(name string, args ...interface{}) *chain {
 			if len(c.context.Path) == 0 {
 				panic("replace allowed only if path is non-empty")
 			}
+			if len(c.context.AliasedPath) == 0 {
+				panic("replace allowed only if aliased path is non-empty")
+			}
 		}()
 	}
 
@@ -296,7 +306,12 @@ func (c *chain) replace(name string, args ...interface{}) *chain {
 
 	chainCopy.state = stateEntered
 	if len(chainCopy.context.Path) != 0 {
-		chainCopy.context.Path[len(chainCopy.context.Path)-1] = fmt.Sprintf(name, args...)
+		last := len(chainCopy.context.Path) - 1
+		chainCopy.context.Path[last] = fmt.Sprintf(name, args...)
+	}
+	if len(chainCopy.context.AliasedPath) != 0 {
+		last := len(chainCopy.context.AliasedPath) - 1
+		chainCopy.context.AliasedPath[last] = fmt.Sprintf(name, args...)
 	}
 
 	return chainCopy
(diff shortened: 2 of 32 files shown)


gavv:

array.go
```diff
@@ -75,6 +75,15 @@ func (a *Array) Path(path string) *Value {
 	return jsonPath(a.chain, a.value, path)
 }
 
+// Alias is similar to Value.Alias.
+func (a *Array) Alias(name string) *Array {
```
gavv, line 79: Here and in other structs: let's place Alias before assertion methods, right after Raw.

array.go
```diff
@@ -75,6 +75,15 @@ func (a *Array) Path(path string) *Value {
...
+	a.chain.enter("Alias(%s)", name)
+	defer a.chain.leave()
+
+	a.chain.setAlias(name)
+	return newArray(a.chain, a.value)
+}
```
gavv, line 85: Here and in other structs: instead of returning new value, we can just return receiver (`a` in this case).

In httpexpect, methods usually return new value if they return something unrelated (e.g. element of array), otherwise they just return receiver.

This will affect tests:

* new value will be the same as original (we can use assert.Same)
* Alias() wont be present in path

gavv (commented):
Thanks for PR!

array.go
k4n4ry: Thanks for review.
For example, for `Array`, should I modify following?
```go
// Alias is similar to Value.Alias.
func (a *Array) Alias(name string) *Array {
	a.chain.setAlias(name)
	return a
}
```

gavv: Yes, but

```
a.chain.enter("Alias(%s)", name)
defer a.chain.leave()
```

is still needed because it will log successful operation (if a logger is used).

k4n4ry: @gavv
Thanks.
As shown below, if `chain.enter()` is called before `setAlias()`, the set Alias will be deleted by `chain.leave()`.
```go
func (a *Array) Alias(name string) *Array {
	a.chain.enter("Alias(%s)", name)
	defer a.chain.leave()

	a.chain.setAlias(name)
	return a
}
```
```go
	value1 := NewArray(reporter, []interface{}{1, 2}).Alias("foo")
	assert.Equal(t, []string{"Array()"}, value1.chain.context.Path)
	assert.Equal(t, []string{"foo"}, value1.chain.context.AliasedPath) // failed, AliasedPath equal []string{}

```

This may be avoided by changing the order of calls as shown below. 
```go
func (a *Array) Alias(name string) *Array {
	a.chain.setAlias(name)

	a.chain.enter("Alias(%s)", name)
	defer a.chain.leave()
	return a
}
```

```go
	value1 := NewArray(reporter, []interface{}{1, 2}).Alias("foo")
	assert.Equal(t, []string{"Array()"}, value1.chain.context.Path)
	assert.Equal(t, []string{"foo"}, value1.chain.context.AliasedPath) // success
```
Is this policy correct?

gavv:

Instead of:

```go
	r.chain.enter("XXX()")
	defer r.leave()

	if r.failed() {
		return r
	}
```

we should now use:

```go
	defer opChain.leave()

	if opChain.failed() {
		return r
	}
```

Related commits:

gavv:
The following changes in chain affect this PR:

* enter does not modify the chain now; instead it creates a clone, modifies this clone, and returns it
* leave does not modify the chain now; since enter returns a clone, we can just drop this clone
* fail is allowed only between enter/leave
* it's not allowed to call enter two times in a row; after first enter you should make a clone, and you can call enter again only on that clone
* replace also returns a clone, just like enter

array.go
gavv: Oh, I see. Yes, this is a problem. However, it seems that it should be solved with the new chain interface that I just pushed and described here:

* https://github.com/gavv/httpexpect/pull/227#issuecomment-1401846252
* https://github.com/gavv/httpexpect/pull/227#issuecomment-1401852700

With new interface, it'll look like:

```go
func (a *Array) Alias(name string) *Array {
	opChain := a.chain.enter("Alias(%s)", name)
	defer opChain.leave()

	a.chain.setAlias(name)
	return a
}
```

Note that leave is called on opChain, but setAlias is still called on a.chain, since we want to modify array's chain, not operation's chain.

chain.go
```diff
@@ -145,13 +149,15 @@ func (c *chain) setResponse(resp *Response) {
...
 	copy.context.Path = append(([]string)(nil), c.context.Path...)
+	copy.context.AliasedPath = append(([]string)(nil), c.context.AliasedPath...)
 
 	return &copy
 }
 
 // Append string to chain path.
 func (c *chain) enter(name string, args ...interface{}) {
 	c.context.Path = append(c.context.Path, fmt.Sprintf(name, args...))
+	c.context.AliasedPath = append(c.context.AliasedPath, fmt.Sprintf(name, args...))
```
gavv, line 160: In addition to enter(), please also update replace().

chain.go
```diff
@@ -166,14 +172,25 @@ func (c *chain) replace(name string, args ...interface{}) {
...
 	}
 
 	c.context.Path = c.context.Path[:len(c.context.Path)-1]
+	c.context.AliasedPath = c.context.AliasedPath[:len(c.context.AliasedPath)-1]
+}
+
+// Initialize and set name to aliased path.
+func (c *chain) setAlias(name string) {
+	c.context.AliasedPath = nil
+	c.context.AliasedPath = append(c.context.AliasedPath, name)
```
gavv, line 193: Nitpick: this can be simplified to:

```
c.context.AliasedPath = []string{name}
```

gavv: Also, chain now has a mutex, so please add lock/unlock to this method too.

gavv:
Sorry for creating conflicts. I was waiting with my changes for a while, in hope that I'll find a time window when there are no active PRs and I can merge them without creating conflicts. However it seems that currently new PRs are coming too often and there is no such window :)

array.go
k4n4ry: I fixed on 534ab72 .

array.go
k4n4ry: Thank you for commits. I fixed on 534ab72 using opChain.

chain.go
k4n4ry: I fixed on 534ab72 .

chain.go
k4n4ry: I fixed on 534ab72 .

chain.go
```diff
@@ -289,14 +296,22 @@ func (c *chain) replace(name string, args ...interface{}) *chain {
...
 
 	chainCopy.state = stateEntered
 	if len(chainCopy.context.Path) != 0 {
-		chainCopy.context.Path[len(chainCopy.context.Path)-1] = fmt.Sprintf(name, args...)
+		last := len(chainCopy.context.Path) - 1
+		chainCopy.context.Path[last] = fmt.Sprintf(name, args...)
+	}
+	if len(chainCopy.context.AliasedPath) != 0 {
+		last := len(chainCopy.context.AliasedPath) - 1
+		chainCopy.context.AliasedPath[last] = fmt.Sprintf(name, args...)
```
k4n4ry, line 314: This re-assignment is to resolve lint lll.

k4n4ry:
@gavv  I rebased and fixed, so please check when you have time. Thank you.

gavv:
Thank you! Solid work!

gavv:
Two small follow-up commits:

* 6223d1fdf880dee0d47be7dd3bc8f6d338921d0b
* 565be649f2be28ed77848cc34e72970a29157310

Issue #187, opened 2022-12-02
Implement FatalReporter


gavv:
In reporter.go, we already have AssertReporter (uses testify/assert) and RequireReporter (uses testify/require), for non-fatal and fatal assertions, respectively.

If you don't use testify, you can directly use testing.T as reporter. httpexpect will call its Errorf method, so assertions will be non-fatal.

If you don't use testify, and want fatal assertions, you have to write your own reporter that invokes t.Fatalf. It would be nice to fill this gap and provide FatalReporter that does that for you - provides Errorf method that just invokes t.Fatalf.

Note: FatalReporter should use testing.TB interface, not testing.T struct.

Note: we should also update comments for Reporter interface and Config.Reporter field. Those comments list available reporter implementations.

kairemor:
Can I check this?

gavv:
You're welcome!

PR #228, opened 2023-01-16
FatalReporter implementation

kairemor:
Implementation of FatalReporter that raises an exception when a test fails, which will stop the execution of the test suite immediately and alert you of a failure

### DIFF
--- a/expect.go
+++ b/expect.go
@@ -158,10 +158,11 @@ type Config struct {
 	// constructed when AssertionHandler is nil.
 	//
 	// You can use:
-	// - AssertReporter / RequireReporter (non-fatal / fatal failures using testify package)
-	// - testing.T / FatalReporter (non-fatal / fatal failures
-	//   using standard testing package)
-	// - custom implementation
+	//  - AssertReporter / RequireReporter
+	//    (non-fatal / fatal failures using testify package)
+	//  - testing.T / FatalReporter
+	//    (non-fatal / fatal failures using standard testing package)
+	//  - custom implementation
 	Reporter Reporter
 
 	// Formatter is used to format success and failure messages.
@@ -312,7 +313,7 @@ type WebsocketDialer interface {
 }
 
 // Reporter is used to report failures.
-// *testing.T, AssertReporter, FatalReporter and RequireReporter implement this interface.
+// *testing.T, AssertReporter, RequireReporter, and FatalReporter implement this interface.
 type Reporter interface {
 	// Errorf reports failure.
 	// Allowed to return normally or terminate test using t.FailNow().
--- a/expect.go
+++ b/expect.go
@@ -159,6 +159,7 @@ type Config struct {
 	//
 	// You can use AssertReporter, RequireReporter (they use testify),
 	// or *testing.T, or provide custom implementation.
+	// You can also Use FatalReporter that use testing.TB
 	Reporter Reporter
 
 	// Formatter is used to format success and failure messages.
--- a/expect.go
+++ b/expect.go
@@ -157,9 +157,11 @@ type Config struct {
 	// Config.Reporter is used by DefaultAssertionHandler, which is automatically
 	// constructed when AssertionHandler is nil.
 	//
-	// You can use AssertReporter, RequireReporter (they use testify),
-	// or *testing.T, or provide custom implementation.
-	// You can also Use FatalReporter that use testing.TB
+	// You can use:
+	// - AssertReporter / RequireReporter (non-fatal / fatal failures using testify package)
+	// - testing.T / FatalReporter (non-fatal / fatal failures 
+	//   using standard testing package)
+	// - custom implementation
 	Reporter Reporter
 
 	// Formatter is used to format success and failure messages.
@@ -310,7 +312,7 @@ type WebsocketDialer interface {
 }
 
 // Reporter is used to report failures.
-// *testing.T, AssertReporter, and RequireReporter implement this interface.
+// *testing.T, AssertReporter, FatalReporter and RequireReporter implement this interface.
 type Reporter interface {
 	// Errorf reports failure.
 	// Allowed to return normally or terminate test using t.FailNow().
--- a/expect.go
+++ b/expect.go
@@ -159,7 +159,7 @@ type Config struct {
 	//
 	// You can use:
 	// - AssertReporter / RequireReporter (non-fatal / fatal failures using testify package)
-	// - testing.T / FatalReporter (non-fatal / fatal failures 
+	// - testing.T / FatalReporter (non-fatal / fatal failures
 	//   using standard testing package)
 	// - custom implementation
 	Reporter Reporter
--- a/reporter.go
+++ b/reporter.go
@@ -53,6 +53,5 @@ func NewFatalReporter(backend testing.TB) *FatalReporter {
 
 // Errorf implements Reporter.Errorf.
 func (r *FatalReporter) Errorf(message string, args ...interface{}) {
-	r.backend.Logf(message, args...)
-	r.backend.Fail()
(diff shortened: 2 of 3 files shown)


gavv:
https://github.com/gavv/httpexpect/issues/187

reporter.go
```diff
@@ -38,3 +39,28 @@ func NewRequireReporter(t require.TestingT) *RequireReporter {
...
+func (r *FatalReporter) Report(result testing.TB) {
+	if result.Failed() {
+		r.Errorf("Test failed")
+	}
+}
+
+// NeFatalReporter returns a new FatalReporter object.
+func NewFatalReporter(backend testing.TB) *FatalReporter {
+	return &FatalReporter{backend}
+}
```
gavv, line 66: Nitpick: usually constructor goes right after struct, before other methods.

reporter.go
```diff
@@ -38,3 +39,28 @@ func NewRequireReporter(t require.TestingT) *RequireReporter {
...
+// Errorf implements Reporter.Errorf.
+func (r *FatalReporter) Errorf(message string, args ...interface{}) {
+	r.backend.Fatalf(message, args...)
+}
+
+// Report is a method that takes a testing.TB interface and checks if the test
+// has failed by calling the Failed() method. If the test has failed, it calls
+// the Errorf method with the provided testing.TB interface and an error message.
+func (r *FatalReporter) Report(result testing.TB) {
+	if result.Failed() {
```
gavv, line 58: Why do we need this method? When is it used?

reporter.go
```diff
@@ -38,3 +39,28 @@ func NewRequireReporter(t require.TestingT) *RequireReporter {
 func (r *RequireReporter) Errorf(message string, args ...interface{}) {
 	r.backend.FailNow(fmt.Sprintf(message, args...))
 }
+
+// FatalReporter is a struct that implements the testing.Reporter interface
+// and calls t.Fatalf() when a test fails.
```
gavv, line 44: `testing.Reporter` -> `Reporter`

expect.go
```diff
@@ -159,6 +159,7 @@ type Config struct {
 	//
 	// You can use AssertReporter, RequireReporter (they use testify),
 	// or *testing.T, or provide custom implementation.
+	// You can also Use FatalReporter that use testing.TB
```
gavv, line 162: Let's reword this a bit:

```
You can use:
  - AssertReporter / RequireReporter (non-fatal / fatal failures using testify package)
  - testing.T / FatalReporter (non-fatal / fatal failures using standard testing package)
  - custom implementation
```

gavv: Please also update comment for Reporter interface, as mentioned in the issue.

gavv (commented):
Thanks for PR!

gavv:
Could you cover new code with a unit test?

This will require a mock for testing.TB. Probably you can do it like this:

```
struct mockT {
    testing.T
}

func (mockT) Fatalf(...) { ... }
```

reporter.go
kairemor: it was about this comment in the issue "<If you don't use testify, and want fatal assertions, you have to write your own reporter that invokes t.Fatalf. It would be nice to fill this gap and provide FatalReporter that does that for you - provides Errorf method that just invokes t.Fatalf. >"

The Report method in the FatalReporter struct is used to check the status of a test and take appropriate action based on the result. It takes a testing.TB interface, which represents a test or benchmark, as its argument. It then checks if the test has failed by calling the Failed() method of the testing.TB interface. If the test has failed, it calls the Errorf method with the provided testing.TB interface and an error message, which will stop the execution of the test suite immediately, preventing any further tests from running.

The Report method is typically called by the Go testing framework at the end of a test or benchmark, after the test function has completed execution. It is used to report the final result of the test or benchmark and take any necessary action based on whether the test passed or failed but also  the Report method is only called when a test or benchmark has completed execution, so if a test fails due to a call to t.Fatalf() or t.FailNow(), the Report method will not be called.

reporter.go
gavv: Also `Ne` -> `New`

reporter.go
gavv: Sorry, I don't get it.

> The Report method is typically called by the Go testing framework at the end of a test or benchmark, after the test function has completed execution.

Are you saying that `testing` package from stdlib calls Report method? How is it supposed to know about our reporter object? I don't see any mention of it in docs.

Also I'm completely missing motivation for this method.

If you do this:

```
r := NewFatalReporter(t)
r.Report(t)
```

it will be equivalent to:

```
if t.Failed() {
    t.Fatalf("Test failed")
}
```

Why do we ever need this? Maybe I'm missing something?

kairemor: Ah ok I see,  so I can remove I thought it will be the way to implement our own reporter that invoke Fatalf as said in the issue comment. I will remove it

reporter.go
```diff
@@ -38,3 +39,20 @@ func NewRequireReporter(t require.TestingT) *RequireReporter {
...
+
+// NewFatalReporter returns a new FatalReporter object.
+func NewFatalReporter(backend testing.TB) *FatalReporter {
+	return &FatalReporter{backend}
+}
+
+// Errorf implements Reporter.Errorf .
+func (r *FatalReporter) Errorf(message string, args ...interface{}) {
+	r.backend.Logf(message, args...)
+	r.backend.Fail()
```
gavv, line 57: Can't we simplify this to `r.backend.Fatalf`?

reporter_test.go
```diff
@@ -0,0 +1,21 @@
+package httpexpect
+
+import (
+	"testing"
+)
+
+type mockT struct {
+	testing.T
+}
```
gavv, line 9: I suggested to wrap testing.T in a struct to redefine its Fatalf method. But you implemented the test without redefining a method (which is a better approach), so wrapping it into a struct is not needed too, we can just create an instance of testing.T directly.

gavv (commented):
Thanks for update

reporter.go
kairemor: I get runtime.Goexit issue with the test if I do so

gavv: Oh, I see; that's because Fail() just marks test as failed. We instead need Fatalf or FailNow.

So we indeed need to wrap testing.T and override Fatalf, as I suggested initially.

kairemor: oK I will overwrite it in  the test file right?

kairemor: also what will look like the overwrited method

reporter_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type mockT struct {
+	testing.T
+}
+
+var fatalfInvoked = false
```
gavv, line 13: Let's make this to be field of mockT, so that different instances of mockT wont affect each other.

reporter_test.go
```diff
@@ -0,0 +1,29 @@
...
+	fatalfInvoked = true
+}
+
+func TestFatalReporter(t *testing.T) {
+	mockBackend := &mockT{}
+	reporter := NewFatalReporter(mockBackend)
+
+	t.Run("Test Errorf", func(t *testing.T) {
+		reporter.Errorf("Test failed with backend: %v", mockBackend)
+		if !mockBackend.Failed() {
```
gavv, line 25: Let's check fatalfInvoked instead. We want to know that Fatalf was invoked.

gavv: We can remove `if !mockBackend.Failed() {` and just keep `assert.True(t, fatalfInvoked)`.

gavv:
Thanks!

gavv:
https://github.com/gavv/httpexpect/issues/187

PR #230, opened 2023-01-21
Implement usage check for Expect in Request and Websocket

### LINKED ISSUE #162: Improve usage checks in Request and Websocket
`Request` and `Websocket` have the following workflow:

* create instance
* likely, call `WithXXX()` methods multiple times
* call `Expect()` method

It is not allowed to call `WithXXX()` after `Expect()` is called. It is also not allowed to call `Expect()` more than once. However, these requirements are not documented and checked.

What we should do:

* Document these requirements in comment for `Expect`.
* Report failure if these requirements are violated with `Type` set to `AssertUsage` and an explanatory message.
* Add unit tests that cover these failure reports.

chunweii:
Fixes #162.

- [X] Add documentation for Expect in Request
- [X] Add usage checks for Expect and WithXXX in Request
- [X] Add unit tests for Expect and WithXXX in Request

### DIFF
--- a/request.go
+++ b/request.go
@@ -208,7 +208,7 @@ func (r *Request) WithName(name string) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithName") {
+	if !r.checkOrder(opChain, "WithName") {
 		return r
 	}
 
@@ -235,7 +235,7 @@ func (r *Request) WithMatcher(matcher func(*Response)) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithMatcher") {
+	if !r.checkOrder(opChain, "WithMatcher") {
 		return r
 	}
 
@@ -269,7 +269,7 @@ func (r *Request) WithTransformer(transform func(*http.Request)) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithTransformer") {
+	if !r.checkOrder(opChain, "WithTransformer") {
 		return r
 	}
 
@@ -309,7 +309,7 @@ func (r *Request) WithClient(client Client) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithClient") {
+	if !r.checkOrder(opChain, "WithClient") {
 		return r
 	}
 
@@ -346,7 +346,7 @@ func (r *Request) WithHandler(handler http.Handler) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithHandler") {
+	if !r.checkOrder(opChain, "WithHandler") {
 		return r
 	}
 
@@ -395,7 +395,7 @@ func (r *Request) WithContext(ctx context.Context) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithContext") {
+	if !r.checkOrder(opChain, "WithContext") {
 		return r
 	}
 
@@ -437,7 +437,7 @@ func (r *Request) WithTimeout(timeout time.Duration) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithTimeout") {
+	if !r.checkOrder(opChain, "WithTimeout") {
 		return r
 	}
 
@@ -506,7 +506,7 @@ func (r *Request) WithRedirectPolicy(policy RedirectPolicy) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithRedirectPolicy") {
+	if !r.checkOrder(opChain, "WithRedirectPolicy") {
 		return r
 	}
 
@@ -538,7 +538,7 @@ func (r *Request) WithMaxRedirects(maxRedirects int) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithMaxRedirects") {
+	if !r.checkOrder(opChain, "WithMaxRedirects") {
 		return r
 	}
 
@@ -606,7 +606,7 @@ func (r *Request) WithRetryPolicy(policy RetryPolicy) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithRetryPolicy") {
+	if !r.checkOrder(opChain, "WithRetryPolicy") {
 		return r
 	}
 
@@ -638,7 +638,7 @@ func (r *Request) WithMaxRetries(maxRetries int) *Request {
 		return r
 	}
 
-	if !r.checkOrder("WithMaxRetries") {
+	if !r.checkOrder(opChain, "WithMaxRetries") {
 		return r
 	}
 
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -40,6 +40,7 @@ type Request struct {
 	minRetryDelay time.Duration
 	maxRetryDelay time.Duration
 	sleepFn       func(d time.Duration) <-chan time.Time
+	hasExpected   bool
```
gavv, line 43: The name is a bit confusing because it also sounds like we have some `expected` value. Maybe rename it to `expectCalled`?

Also this block of fields is for retry stuff, let's move new field to another block, dedicated for similar stuff (flags and info about what was called already):

```
	bodySetter string
	typeSetter string
	forceType  bool
```

request.go
```diff
@@ -200,6 +201,11 @@ func (r *Request) WithName(name string) *Request {
 	r.chain.enter("WithName()")
 	defer r.chain.leave()
 
+	if r.hasExpected {
+		r.chain.fail(r.unexpectedExpectError("WithName"))
+		return r
+	}
```
gavv, line 207: Can we please refactor this to:

```
if !r.checkOrder("WithName") {
	return r
}
```

Why:

* this way checks wont be duplicated
* if we'll add more order checks, they will be automatically used everywhere
* it matches how we do such things in other places

request.go
```diff
@@ -2097,6 +2278,15 @@ func (r *Request) setBody(setter string, reader io.Reader, len int, overwrite bo
 	r.bodySetter = setter
 }
 
+func (r *Request) unexpectedExpectError(funcCall string) AssertionFailure {
+	return AssertionFailure{
+		Type: AssertUsage,
+		Errors: []error{
+			errors.New("unexpected call to " + funcCall + ": Expect has already been called"),
```
gavv, line 2285: Nitpick: let's use `fmt.Errorf`

request.go
```diff
@@ -1663,6 +1837,11 @@ func (r *Request) Expect() *Response {
 	r.chain.enter("Expect()")
 	defer r.chain.leave()
 
+	if r.hasExpected {
+		r.chain.fail(r.unexpectedExpectError("Expect"))
+		return nil
```
gavv, line 1842: Expect should never return nil. On error, it should return non-nil, but failed Response. Like this:

```
		return newResponse(responseOpts{
			config: r.config,
			chain:  r.chain,
		})
```

request_test.go
```diff
@@ -2231,6 +2232,396 @@ func TestRequest_UsageChecks(t *testing.T) {
...
+	})
+
+	t.Run("WithName after an Expect", func(t *testing.T) {
+		config := Config{
+			Reporter: newMockReporter(t),
+			Client:   &mockClient{},
+		}
+		req := NewRequestC(config, "GET", "/")
+		req.Expect()
+		assert.Equal(t, req, req.WithName("Test"))
```
gavv, line 2254: Here and below, let's use assert.Same instead of assert.Equal

gavv (commented):
Thanks for PR! Here are a few comments.

gavv:
Checks are missing in:

* WithRedirectPolicy
* WithMaxRedirects

gavv:
Also, I just noticed that WithName and Expect are missing this check:

```
	if r.chain.failed() {
		return r
	}
```

Could you please add it as well?

gavv:

Instead of:

```go
	r.chain.enter("XXX()")
	defer r.leave()

	if r.failed() {
		return r
	}
```

we should now use:

```go
	defer opChain.leave()

	if opChain.failed() {
		return r
	}
```

Related commits:

request.go
```diff
@@ -2097,6 +2264,23 @@ func (r *Request) setBody(setter string, reader io.Reader, len int, overwrite bo
...
 
+func (r *Request) checkOrder(funcCall string) bool {
+	if r.expectCalled {
+		r.chain.fail(r.unexpectedExpectError(funcCall))
+		return false
+	}
+	return true
+}
+
+func (r *Request) unexpectedExpectError(funcCall string) AssertionFailure {
```
gavv, line 2275: Since unexpectedExpectError() is used only in one place one, we can get rid of the function and inline it.

request_test.go
```diff
@@ -2231,6 +2232,418 @@ func TestRequest_UsageChecks(t *testing.T) {
 		req.Expect()
 		req.chain.assertFailed(t)
 	})
+
+	t.Run("Expect after an Expect", func(t *testing.T) {
```
gavv, line 2236: Since there are a lot of new subtests, let's create a new test TestRequest_OrderChecks.

request.go
```diff
@@ -1668,6 +1819,20 @@ func (r *Request) Expect() *Response {
 	opChain := r.chain.enter("Expect()")
 	defer opChain.leave()
 
+	if opChain.failed() {
+		return newResponse(responseOpts{
+			config: r.config,
+			chain:  r.chain,
```
gavv, line 1825: `r.chain` -> `opChain`

request.go
```diff
@@ -1668,6 +1819,20 @@ func (r *Request) Expect() *Response {
...
+		})
+	}
+
+	if !r.checkOrder(opChain, "Expect") {
+		return newResponse(responseOpts{
+			config: r.config,
+			chain:  r.chain,
```
gavv, line 1832: `r.chain` -> `opChain`

request.go
```diff
@@ -2107,6 +2274,19 @@ func (r *Request) setBody(
 	r.bodySetter = setter
 }
 
+func (r *Request) checkOrder(opChain *chain, funcCall string) bool {
+	if r.expectCalled {
+		opChain.fail(AssertionFailure{
+			Type: AssertUsage,
+			Errors: []error{
+				fmt.Errorf("unexpected call to " + funcCall + ": Expect has already been called"),
```
gavv, line 2282: Nitpick: let's replace it with:

```
fmt.Errorf("unexpected call to %s: Expect() has already been called", funcCall)
```

request.go
```diff
@@ -203,6 +204,14 @@ func (r *Request) WithName(name string) *Request {
 	opChain := r.chain.enter("WithName()")
 	defer opChain.leave()
 
+	if opChain.failed() {
+		return r
+	}
+
+	if !r.checkOrder(opChain, "WithName") {
```
gavv, line 211: When we pass function names to enter(), setType(), setBody(), etc, we use form `"Foo()"` instead of `"Foo"`.

For consistency, could you please adjust checkOrder calls accordingly? So that

```
r.checkOrder(opChain, "WithName")
```

will be replaced with

```
r.checkOrder(opChain, "WithName()")
```

and so on.

gavv (commented):
Thanks for update! A few more minor comments.

gavv:
Thanks you!!

Issue #191, opened 2022-12-05
Migrate Number from float64 to big.Float


gavv:
Currently, Number always converts its value to float64. In many cases it works well, however float64 can't represent big integer values (larger than 2^62) without precision loss.

Actually, there is no native type that can represent ALL values of int64, uint64, and float64 without loss. However, there is a library type that can do it: big.Float (see https://pkg.go.dev/math/big).

Here are the steps:

* replace underlying value of Number from float64 to big.Float

  * by default, use precision that is sufficient to represent all values of int64, uint64, and float64 without loss

  * allow user to set higher precision in Config (this may be useful if user wants to parse Number from String that contains bigger integer or float)

* rework canonNumber() to return big.Float instead of float64; it should automatically detect precision and create big.Float from any of these types:

  * all native float and integer types
  * type definitions for those native types (e.g. `type MyInt int`)
  * big.Float and big.Int
  * json.Number

* rework NewNumber constructor: it should accept interface{} instead of float64 and use canonNumber() to convert it to big.Float

* rework all Number operations, so that they will operate on big.Float values (one stored in Number, another returned from canonNumber for operation argument)

* rework Number.EqualDelta and Number.NotEqualDelta: they should accept interface{} instead of float64; this will allow user to pass big.Float instead of float64, when necessary

* rework JSON decoding:

  * find places where we decode JSON: canonValue, Response, WebsocketMessage; configure JSON decoder to use json.Number instead of float64 (see https://pkg.go.dev/encoding/json#Decoder.UseNumber)

  * ensure that canonValue() and other canonization functions does not cause precision loss

* rework Number creation

  * Expect.Number() should accept interface{} instead of float64

  * rework Value.Number() - it should not force float64 type

  * rework String.AsNumber() - it should correctly parse big floats and integers

  * find all invocations to newNumber() that have unnecessary casts to float64 and pass integers instead; e.g. `newNumber(a.chain, float64(len(a.value)))` can be now replaced with `newNumber(a.chain, len(a.value))` because Number now supports integers

* add new getters for Number:

  * add 4 new methods: AsInt (returns int64), AsUint (return uint64), AsFloat (return float64), AsBig (return big.Float)

  * each method should report failure if underlying value can't be converted to requested types without loss

  * add IsBig (similar to other Is methods, see #155)

[...]

PR #232, opened 2023-01-24
WIP: Migration from float64 to big.Float

angshumanHalder:


gavv:

Instead of:

```go
	r.chain.enter("XXX()")
	defer r.leave()

	if r.failed() {
		return r
	}
```

we should now use:

```go
	defer opChain.leave()

	if opChain.failed() {
		return r
	}
```

Related commits:

gavv:
BTW if you don't mind, could you rollback reordering of methods in *this* PR? No problem with reordering itself, but they make the diff a mess and it's hard to tell what changed exactly.

angshumanHalder:
Sure. Would rollback the ordering. Actually When I raised this PR initially I had conflicts which upon resolving messed up the ordering of the functions. I will revert it to initial ordering.

gavv:
#191

gavv:
Could you please revert unrelated changes (like formatting) in tests? This is a big pr, and it makes reviewing harder.

gavv:
Also please remove commented code

gavv:
Could you also provide a short summary of what is addressed by this pr and what is not?

angshumanHalder:
Sure. Let me resolve conflicts and revert formatting changes. Give me few days.

angshumanHalder:
Hey @gavv reverted the formatting changes. But can you check why the `linter for example` job is failing? Because when I run it on my local system it doesn't fail. Doesn't output any timeout or errors.

PR #238, opened 2023-01-26
Decode method

Rohith-Raju:
Continuation pr for #192


object_test.go
```diff
@@ -170,13 +170,16 @@ func TestObject_Decode(t *testing.T) {
...
 		value := NewObject(reporter, m)
 
-		actualStruct := S{123, []interface{}{"123", 234.0},
-			map[string]interface{}{"a": "b"}, struct{ a int }{0},
+		actualStruct := S{
+			123,
+			[]interface{}{"123", 234.0},
+			map[string]interface{}{"a": "b"},
+			struct{ A int }{123},
 		}
```
gavv, line 183: Please add initializers (`Foo:`, `Bar:`, etc)

array.go
```diff
@@ -74,6 +74,7 @@ func (a *Array) Raw() []interface{} {
 // target should be one of these:
 //
 // - pointer to an empty interface
+//
 // - pointer to a slice of any type
```
gavv, line 78: I guess here and below this changes are unintentional? I think you did not rebase on fresh master.

gavv:
Thanks for PR

array.go
Rohith-Raju: No, it was intentional. without the extra gap, all list items will be formated in a single line. Should I change it back?

Rohith-Raju: I missed this detail for all the previous commits.

gavv: Oh, I see. Nope, it was me, not you: 1623932e79bafe5a3e6d87623a14611851606b4d

Actually no need for empty lines, instead we need one more space. Instead of:

```
// - pointer to an empty interface
// - pointer to a slice of any type
```

we need:

```
//  - pointer to an empty interface
//  - pointer to a slice of any type
```

Would be awesome if you could fix it.

gavv:
> [Coverage Status](https://coveralls.io/builds/56318742)
> 
> Coverage: 94.829% (-0.02%) from 94.847% when pulling **[9d224c6](https://github.com/gavv/httpexpect/commit/9d224c6b84c5178a413840409a23c669027c173c) on Rohith-Raju:DecodeMethod** into **[eb535fe](https://github.com/gavv/httpexpect/commit/eb535fe85f25570429c24f5633672db35cfa5a63) on gavv:master**.

Please take a look at the uncovered branch.

array.go
Rohith-Raju: Oh ok yea sure.

Rohith-Raju:
oops, vscode linter went crazy. let me fix that

gavv:
Thanks for update. I've pushed a few minor fixes.

Rohith-Raju:
Thank you!!

Issue #233, opened 2023-01-24
Make structs thread-safe


gavv:
Background: https://github.com/gavv/httpexpect/discussions/229

I've pushed a few changes that make the `chain` struct thread-safe:

* 0547acd58636b8dc199f43fd7206d9bb629e7852
* a96122195174a5b9df2804e41c8abe92a90fc52b
* 90675e11729f5dfa5121bbf41a1e071dbf9e52e7

This automatically makes many structs thread-safe too: Expect, Value, Array, Object, etc. It's true for matcher structs which contains only a chain + immutable data.

The next step is to manually implement thread-safety for the following structs:

- [ ] Environment
- [ ] Request
- [ ] Websocket

These three structs has mutable state and should be protected manually. For each of them, we should add sync.RWMutex field and protect all operations.

Request struct needs special care: it contains `transforms` and `matchers` callbacks that are invoked during Expect() call. We should NOT call them under a lock because otherwise it would be easy to write code with deadlocks (if you use the same request from this callbacks).

In addition, we should improve our testing suite:

* [ ] add `make race` target which runs all tests under race detector
* [ ] add corresponding step to CI
* [ ] implement a stress test that uses various httpexpect structs from multiple goroutines; it can create Expect instance, pass it to a few goroutines, which repeatedly create requests, also pass them to a few goroutines, and so on

Rohith-Raju:
@gavv I'd like to take this up. Sounds interesting!!

gavv:
Great!

gavv:
I suggest to split work into three PRs:

* make Environment thread-safe (easy part)
* make Websocket thread-safe (harder)
* make Request thread-safe (harder)
* add `make race` to makefile and CI
* add stress test (not very hard but big and time consuming)

gavv:
Added this paragraph to the issue:

> Websocket also needs some special handling. It should allow to read and write messages concurrently and to call disconnect concurrently (similarly as gorilla websocket which it uses under the hood).

gavv:
@Rohith-Raju 

I was reviewing #254 and realized that there is a problem with our approach.

We acquire a lock, then perform a check, and then report failure, under the lock. A potential problem is that reporting failure means invoking AssertionHandler, Reporter, Logger, Formatter, which may be implemented by user. In other words, we invoke user code under a lock.

If this user code will try to access our object, we'll get deadlock. For example, if AssertionHandler tries to get a value from env, env method will try to acquire a lock and will hang.

[...]

PR #254, opened 2023-01-29
Initial commit for environment test

Rohith-Raju:
Signed-off-by: ROHITH RAJU <rohithraju488@gmail.com>
1st pr for #233

### DIFF
--- a/environment.go
+++ b/environment.go
@@ -2,6 +2,7 @@ package httpexpect
 
 import (
 	"errors"
+	"sync"
 	"time"
 )
 
@@ -13,9 +14,9 @@ import (
 //	env.Put("key", "value")
 //	value := env.GetString("key")
 type Environment struct {
-	noCopy noCopy
-	chain  *chain
-	data   map[string]interface{}
+	rw    sync.RWMutex
+	chain *chain
+	data  map[string]interface{}
 }
 
 // NewEnvironment returns a new Environment.
@@ -55,6 +56,9 @@ func newEnvironment(parent *chain) *Environment {
 //	env.Put("key1", "str")
 //	env.Put("key2", 123)
 func (e *Environment) Put(key string, value interface{}) {
+	e.rw.Lock()
+	defer e.rw.Unlock()
+
 	opChain := e.chain.enter("Put(%q)", key)
 	defer opChain.leave()
 
@@ -69,6 +73,9 @@ func (e *Environment) Put(key string, value interface{}) {
 //	env.Put("key1", "str")
 //	env.Delete("key1")
 func (e *Environment) Delete(key string) {
+	e.rw.Lock()
+	defer e.rw.Unlock()
+
 	opChain := e.chain.enter("Delete(%q)", key)
 	defer opChain.leave()
 
@@ -83,6 +90,9 @@ func (e *Environment) Delete(key string) {
 //	   ...
 //	}
 func (e *Environment) Has(key string) bool {
+	e.rw.RLock()
+	defer e.rw.RLocker()
+
 	opChain := e.chain.enter("Has(%q)", key)
 	defer opChain.leave()
 
@@ -99,6 +109,9 @@ func (e *Environment) Has(key string) bool {
 //	value1 := env.Get("key1").(string)
 //	value2 := env.Get("key1").(int)
 func (e *Environment) Get(key string) interface{} {
+	e.rw.RLock()
+	defer e.rw.RLocker()
+
 	opChain := e.chain.enter("Get(%q)", key)
 	defer opChain.leave()
 
@@ -115,6 +128,9 @@ func (e *Environment) Get(key string) interface{} {
 //
 //	value := env.GetBool("key")
 func (e *Environment) GetBool(key string) bool {
+	e.rw.RLock()
+	defer e.rw.RLocker()
+
 	opChain := e.chain.enter("GetBool(%q)", key)
 	defer opChain.leave()
 
@@ -147,6 +163,9 @@ func (e *Environment) GetBool(key string) bool {
 //
 //	value := env.GetInt("key")
 func (e *Environment) GetInt(key string) int {
+	e.rw.RLock()
+	defer e.rw.RLocker()
+
 	opChain := e.chain.enter("GetInt(%q)", key)
 	defer opChain.leave()
 
@@ -232,6 +251,9 @@ func (e *Environment) GetInt(key string) int {
 //
 //	value := env.GetFloat("key")
 func (e *Environment) GetFloat(key string) float64 {
+	e.rw.RLock()
+	defer e.rw.RLocker()
+
 	opChain := e.chain.enter("GetFloat(%q)", key)
 	defer opChain.leave()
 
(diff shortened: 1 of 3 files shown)


Rohith-Raju:
tests taking abnormal amount of time, looking if something wrong.

Rohith-Raju:
@gavv your thoughts?

gavv:
Hi,

Which tests, how much time, and on what machine? 

Changes look good.

environment.go
```diff
@@ -13,9 +14,9 @@ import (
 //	env.Put("key", "value")
 //	value := env.GetString("key")
 type Environment struct {
-	noCopy noCopy
-	chain  *chain
-	data   map[string]interface{}
+	rw    sync.RWMutex
```
gavv, line 17: Nitpick: let's rename it to mu, a conventional name for mutexes.

gavv:
https://github.com/gavv/httpexpect/issues/233#issuecomment-1407738099

Rohith-Raju:
> Hi,
> 
> > Tests taking an abnormal amount of time, looking if something is wrong.
> 
> Which tests, how much time, and on what machine?
> 
> Or is it fixed already?
> 
> Changes look good.

Yes @gavv, fixed.

gavv:
> https://github.com/gavv/httpexpect/issues/233#issuecomment-1407738099

Also it would be nice to add a test that addresses this problem.

The test will:

* set custom reporter that invokes env.Get or any other env method 

* call each env method in a way so that the mthid should fail

* check that reporter was invoked each time

Here we simulate accessing env from user code. If we call user code under a lock, this test will hang.

We'll need similar tests for other structs  to which we'll add locks.

gavv:
I also created an issue for creating a similar test for chain: https://github.com/gavv/httpexpect/issues/255

gavv:
The changes look good.

To merge this pr, we'll need to make another change first which I described in https://github.com/gavv/httpexpect/issues/233#issuecomment-1407738099

And also we'll need to add a test I described to be sure that deadlock problem is prevented.

gavv:
Oh, and you forgot to reorder chaining and locking in some methods.

environment_test.go
```diff
@@ -53,6 +53,34 @@ func TestEnvironment_Basic(t *testing.T) {
 	env.chain.assertFailed(t)
 }
 
+type customReporter struct {
+	testing  *testing.T
+	reported bool
+}
```
gavv, line 59: Small improvement idea: instead of creating a new struct `customReporter`, we can modify `mockReporter` as follows:

* add func() field to struct, say `reportCb func()`
* in Errorf, check if the field is non-nil, and if so, invoke it

Then in our test we can create mock reporter and set it's reportCb to a closure that invokes env.

Rationale:

* less code
* same reporter will be reused for deadlock tests for other structs too

environment_test.go
```diff
@@ -53,6 +53,34 @@ func TestEnvironment_Basic(t *testing.T) {
...
+
+func newCustomReporter(t *testing.T) *customReporter {
+	return &customReporter{testing: t}
+}
+
+func (c *customReporter) Errorf(message string, args ...interface{}) {
+	env := NewEnvironment(c)
+	env.Get("key1")
```
gavv, line 67: This does not make sense. We should call `Get` on the same Environment struct that invokes reporter, not on the new one.

The idea is that env calls reporter, and reporter calls back env. And that's why deadlock appears: because the second call to env from reporter try to lock the same and already locked mutex.

environment_test.go
```diff
@@ -53,6 +53,34 @@ func TestEnvironment_Basic(t *testing.T) {
...
+	c.testing.Logf("Fail: "+message, args...)
+	c.reported = true
+}
+func TestEnvironment_Deadlock(t *testing.T) {
```
gavv, line 71: Let's rename this test to `TestEnvironment_Reentrant`.

TestEnvironment_Deadlock name suggests that we want that env has deadlock.

TestEnvironment_Reentrant name suggests that env is reentrant, i.e. it's ok if it indirectly calls itself (via reporter) and there is *no* deadlock.

https://en.wikipedia.org/wiki/Reentrancy_(computing)

environment_test.go
```diff
@@ -53,6 +53,34 @@ func TestEnvironment_Basic(t *testing.T) {
...
+	reporter := newCustomReporter(t)
+
+	env := NewEnvironment(reporter)
+
+	env.Put("key1", map[string]string{
+		"Foo": "123",
+	},
+	)
+	env.GetString("key1")
```
gavv, line 80: Instead of Put + GetString, we can just call Get with an nonexistent key. This is a shorter way to get failure.

environment_test.go
```diff
@@ -53,6 +53,34 @@ func TestEnvironment_Basic(t *testing.T) {
...
+	env.chain.failed()
```
gavv, line 81: `env.chain.assertFailed(t)`

gavv:
In your current test there is no deadlock, there is stack overflow. See https://github.com/gavv/httpexpect/actions/runs/4062692800/jobs/6994044503

That's because this:

```
	env := NewEnvironment(c)
	env.Get("key1")
```

Causes failure again, and its calls reporter again, and it causes failure again, and so on. You have an infinite loop that aborts only when the stack becomes too large.

There is no deadlock because of this: https://github.com/gavv/httpexpect/pull/254#discussion_r1092959358

To avoid infinite loop, reporter should make a call that will NOT fail. It should Get an existing key, for example. Or we can replace it with env.Put.

Rohith-Raju:
> In your current test there is no deadlock, there is stack overflow. See https://github.com/gavv/httpexpect/actions/runs/4062692800/jobs/6994044503
> 
> That's because this:
> 
> ```
> 	env := NewEnvironment(c)
> 	env.Get("key1")
> ```
> 
> causes env to fail again, and it calls reporter again, and it causes failure again, and so on. You have an infinite loop that aborts only when the stack becomes too large.
> 
> There is no deadlock because of this: [#254 (comment)](https://github.com/gavv/httpexpect/pull/254#discussion_r1092959358)
> 
> To avoid infinite loop, reporter should make a call that will NOT fail. It should Get an existing key, for example. Or we can replace it with env.Put.

Yes, that makes complete sense. I should've checked!!

environment_test.go
```diff
@@ -53,6 +53,19 @@ func TestEnvironment_Basic(t *testing.T) {
...
+func TestEnvironment_Reentrant(t *testing.T) {
+	reporter := newMockReporter(t)
+
+	env := NewEnvironment(reporter)
+
+	reporter.reportCb = func() {
+		env.Put("good_key", 123)
+	}
+
+	env.Get("key")
```
gavv, line 65: Nitpick: let's call it "bad_key" to make it clear that this causes failure.

mocks_test.go
```diff
@@ -179,13 +179,17 @@ func (l *mockLogger) Logf(message string, args ...interface{}) {
...
 	return &mockReporter{testing: t}
 }
 
 func (r *mockReporter) Errorf(message string, args ...interface{}) {
+	if r.reportCb != nil {
+		r.reportCb()
+	}
 	r.testing.Logf("Fail: "+message, args...)
 	r.reported = true
 }
```
gavv, line 195: Nitpick: let's set reported first and then call reportCb. So that when callback is called, it will observe that reported is already true.

gavv (commented):
The test looks good now.

gavv:
LGTM

gavv:
Thanks! I've updated the test a bit.

Issue #250, opened 2023-01-28
Implement InList and NotInList methods


gavv:
Some matcher structs (Number, DateTime) have `InRange` method that checks that the value belongs to given range `[min; max]`.

Another convenient thing would be `InList` method, that checks that the value equals to one of the given values `a, b, c, ...`.

For example:

```go
array.Element(1).String().InList("green", "red")
```

We need to add `InList` and counterpart `NotInList` methods to the following structs:

* Value
* Object
* Array
* String
* Number
* Boolean (just for completeness)
* Duration
* DateTime

The new methods can use assertions of type AssertBelongs and AssertNotBelongs. Please place these methods after `InRange`, or after `Equal*` if there is no `InRange` for this struct.

Each method should have documentation comment with code example and be covered by a unit test. It would be also nice to demonstrate usage of InRange and InList in `_examples/fruits_test.go`.

fahimbagar:
Hi @gavv, I have created PR for this issue: https://github.com/gavv/httpexpect/pull/256

Could you check it? and let me know how it goes

PS.
1. Since `AssertionList` rejects a single slice as a param, single array param in `InList` will behave like `IsEqual` and `NotInList` will behave like `NotEqual`. Added in https://github.com/gavv/httpexpect/commit/649ee8caf0e8c918aa9a2de8a8d6f6c5c3ec1cf7
https://github.com/gavv/httpexpect/blob/86f366a2d10b8a8ddd3cb1a9b2a5c00f37a1df68/assertion_validation.go#L186-L190

2. I'll update `_examples/fruits_test.go` if this PR is merged. Or do you have an alternative without modifying `go.mod` under `_examples`?

gavv:
Thanks for pr! Will take a look shortly.

Re: PS

1. Let's just remove this check from assertion validation. It was useful euristics to catch bugs, but it's not working well with InList.

2. Yes, let's update examples separately, thanks.

fahimbagar:
Addressed in https://github.com/gavv/httpexpect/pull/256/commits/40c6e50cc826924a245632406c46278e2c0f9a98.

Sure, thanks for your awesome repo as well :100:

gavv:
Reopening after auto-close because we also need to update example.

gavv:
Created follow-up issue: https://github.com/gavv/httpexpect/issues/261

gavv:
I've tagged v2.11.0, which includes InList.

fahimbagar:
[...]

PR #256, opened 2023-01-30
Add InList and NotInList

fahimbagar:
This PR does the following:
- Add `InList` and `NotInList` for given types:
  - [x] Value
  - [x] Object
  - [x] Array
  - [x] String
  - [x] Number
  - [x] Boolean (just for completeness)
  - [x] Duration
  - [x] DateTime
-Since `AssertionList` rejects a single slice as a param, single array param in `InList` will behave like `IsEqual` and `NotInList` will behave like `NotEqual`. Addressed in commit https://github.com/gavv/httpexpect/commit/649ee8caf0e8c918aa9a2de8a8d6f6c5c3ec1cf7

https://github.com/gavv/httpexpect/blob/86f366a2d10b8a8ddd3cb1a9b2a5c00f37a1df68/assertion_validation.go#L186-L190

For issue https://github.com/gavv/httpexpect/issues/250

### DIFF
--- a/array.go
+++ b/array.go
@@ -776,10 +776,6 @@ func (a *Array) InList(values ...interface{}) *Array {
 		return a
 	}
 
-	if len(arr) == 1 {
-		return a.IsEqual(arr[0])
-	}
-
 	var isListed bool
 	for _, v := range values {
 		expected, ok := canonArray(opChain, v)
@@ -834,10 +830,6 @@ func (a *Array) NotInList(values ...interface{}) *Array {
 	arr, ok := canonArray(opChain, values)
 	if !ok {
 		return a
-	}
-
-	if len(arr) == 1 {
-		return a.NotEqual(arr[0])
 	}
 
 	var isListed bool
--- a/array.go
+++ b/array.go
@@ -750,8 +750,7 @@ func (a *Array) Equal(value interface{}) *Array {
 // Before comparison, both array and value are converted to canonical form.
 //
 // values should be an array of slice of any type. This comparison adheres
-// element order. For values with length 1, it behaves the same way with
-// IsEqual.
+// element order.
 //
 // Example:
 //
@@ -812,8 +811,7 @@ func (a *Array) InList(values ...interface{}) *Array {
 // Before comparison, both array and value are converted to canonical form.
 //
 // values should be an array of slice of any type. This comparison adheres
-// element order. For values with length 1, it behaves the same way with
-// NotEqual.
+// element order.
 //
 // Example:
 //
--- a/array.go
+++ b/array.go
@@ -771,8 +771,14 @@ func (a *Array) InList(values ...interface{}) *Array {
 		return a
 	}
 
-	arr, ok := canonArray(opChain, values)
-	if !ok {
+	if len(values) == 0 {
+		opChain.fail(AssertionFailure{
+			Type: AssertUsage,
+			Errors: []error{
+				errors.New("unexpected empty list argument"),
+			},
+		})
+
 		return a
 	}
 
@@ -792,7 +798,7 @@ func (a *Array) InList(values ...interface{}) *Array {
 		opChain.fail(AssertionFailure{
 			Type:     AssertBelongs,
 			Actual:   &AssertionValue{a.value},
-			Expected: &AssertionValue{AssertionList(arr)},
+			Expected: &AssertionValue{AssertionList(values)},
 			Errors: []error{
 				errors.New("expected: arrays are listed"),
 			},
@@ -827,13 +833,19 @@ func (a *Array) NotInList(values ...interface{}) *Array {
 		return a
 	}
 
-	arr, ok := canonArray(opChain, values)
-	if !ok {
+	if len(values) == 0 {
+		opChain.fail(AssertionFailure{
+			Type: AssertUsage,
+			Errors: []error{
+				errors.New("unexpected empty list argument"),
+			},
+		})
+
 		return a
 	}
 
 	var isListed bool
-	for _, v := range arr {
+	for _, v := range values {
 		expected, ok := canonArray(opChain, v)
 		if !ok {
 			return a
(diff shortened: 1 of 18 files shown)


gavv:
I think we should have separate check for zero-length list.

Such call will always fail, but AssertBelongs will look confusing.

Since it anyway will fail, let's forbid this case. If length of list is zero, let's fail with AssertUsage with error "unexpected empty list".

fahimbagar:
> Since it anyway will fail, let's forbid this case. If length of list is zero, let's fail with AssertUsage with error "unexpected empty list argument".

> Also would be nice to cover these usage checks with tests.

Addressed both, let me know if I miss out anything. Also removed unneeded `canonArray` in https://github.com/gavv/httpexpect/pull/256/commits/b7fc213ce722a152f674015a898290fe3475e965 and https://github.com/gavv/httpexpect/pull/256/commits/b885435c99ef42ba2d7c82b34021ef29f6de229c

array.go
```diff
@@ -746,6 +746,128 @@ func (a *Array) Equal(value interface{}) *Array {
 	return a.IsEqual(value)
 }
 
+// InList succeeds if array is listed by given [values...].
```
gavv, line 749: Here and in other methods, I suggest to reword this part a bit:

```
InList succeeds if whole array is equal to one of the elements from given list of arrays.
```

array.go
```diff
@@ -746,6 +746,128 @@ func (a *Array) Equal(value interface{}) *Array {
...
+// Before comparison, both array and value are converted to canonical form.
```
gavv, line 750: Let's reword this a bit too:

```
Before comparison, both array and each value are converted to canonical form.
```

array.go
```diff
@@ -746,6 +746,128 @@ func (a *Array) Equal(value interface{}) *Array {
...
+//
+// values should be an array of slice of any type. This comparison adheres
+// element order.
```
gavv, line 753: And this:

```
Each value should be a slice of any type.
```

array.go
```diff
@@ -746,6 +746,128 @@ func (a *Array) Equal(value interface{}) *Array {
...
+// Example:
+//
+//	array := NewArray(t, []interface{}{"foo", 123})
+//	array.InList([]interface{}{"foo", 123})
+//
+//	array := NewArray(t, []interface{}{"foo", "bar"})
+//	array.InList([]string{}{"foo", "bar"})
+//
+//	array := NewArray(t, []interface{}{123, 456})
+//	array.InList([]int{}{123, 456})
```
gavv, line 764: Let's provide example with multiple arguments to InList, to make it clear that it accepts list of slices, not a single slice.

array.go
```diff
@@ -746,6 +746,128 @@ func (a *Array) Equal(value interface{}) *Array {
...
+		}
+	}
+
+	if !isListed {
+		opChain.fail(AssertionFailure{
+			Type:     AssertBelongs,
+			Actual:   &AssertionValue{a.value},
+			Expected: &AssertionValue{AssertionList(values)},
+			Errors: []error{
+				errors.New("expected: arrays are listed"),
```
gavv, line 802: Here and in other methods, let's reword error message:

```
expected: arrays is equal to one of the values
```

boolean.go
```diff
@@ -162,6 +162,99 @@ func (b *Boolean) Equal(value bool) *Boolean {
...
+}
+
+func boolList(values []bool) []interface{} {
+	l := make([]interface{}, 0, len(values))
+	for _, v := range values {
+		l = append(l, v)
+	}
+
+	return l
+}
```
gavv, line 256: Nitpick: please move this to the end of file.

array.go
```diff
@@ -746,6 +746,128 @@ func (a *Array) Equal(value interface{}) *Array {
...
+	var isListed bool
+	for _, v := range values {
+		expected, ok := canonArray(opChain, v)
+		if !ok {
+			return a
+		}
+
+		if reflect.DeepEqual(expected, a.value) {
+			isListed = true
+		}
```
gavv, line 793: Let's add `break` here and in similar places

boolean.go
```diff
@@ -162,6 +162,99 @@ func (b *Boolean) Equal(value bool) *Boolean {
...
+	opChain.fail(AssertionFailure{
+		Type:     AssertBelongs,
+		Actual:   &AssertionValue{b.value},
+		Expected: &AssertionValue{AssertionList(boolList(values))},
+		Errors: []error{
+			errors.New("expected: boolean is listed"),
+		},
+	})
+
+	return b
```
gavv, line 205: Nitpick: usually httpexpect methods use the following pattern:

* method can have early returns in case of failures
* there is only one successful return, at the end of the method

You used exactly this layout in Array.InList, but in Boolean.InList you used the opposite layout: early return for success, failure in the end of the method.

I suggest to refactor all InList to follow the same layout as in Array.InList, for consistency.

gavv (commented):
Thanks for update! A few comments

boolean.go
fahimbagar: Thanks for the review.

Previously, I tried to use early returns but I had a doubt since types like `Array` receives `interface{}` as param, so I tried to iterate everything first to check each of the params and then decided whether it's succeeded.

I will take this as a guideline
> - method can have early returns in case of failures
> - there is only one successful return, at the end of the method

array.go
fahimbagar: Addressed in `Array` and others as well.

array.go
fahimbagar: Addressed in `Array` and others as well.

array.go
fahimbagar: Addressed in `Array` and `Number`

array.go
fahimbagar: Addressed in `Array` and others as well.

array.go
fahimbagar: Addressed in `Array` and others as well.

boolean.go
fahimbagar: addressed in https://github.com/gavv/httpexpect/pull/256/commits/0d5cba174efafb67610d8f5eda37b1fa4ea8e193

boolean.go
fahimbagar: `InList` uses `isListed` and breaks early in each type. While `NotInList` add failed step and breaks early if same.

Let me know if I miss something there, @gavv.

gavv:
Thanks!

gavv:
Small follow-up commit: 8fe2ffba3601d60a56a162077d20f4a40833233b

gavv:
PS. If you like chat format, welcome to our [discord](https://discord.gg/5SCPCuCWA9) :)

PR #257, opened 2023-01-31
Add Value.IsXxx

### LINKED ISSUE #253: Add Value.IsXxx
`Value` has several methods to cast value to specific type:

* Object
* Array
* String
* Number
* Boolean

Each of them checks that the value has correct type and returns a new object for the value.

For each of these methods, we should add a counterpart method `IsXxx` (e.g. IsObject for Object) which just checks that the value has specific type, but does not create a new object, but instead returns the original (receiver) object.

New methods should have documentation comments and unit tests.

Related: https://github.com/gavv/httpexpect/issues/252

nhAnik:
Closes #253

### DIFF
--- a/value.go
+++ b/value.go
@@ -624,3 +624,295 @@ func (v *Value) NotInList(values ...interface{}) *Value {
 
 	return v
 }
+
+// IsObject succeeds if the underlying value is an object.
+//
+// If underlying value is not an object (map[string]interface{}), failure is reported.
+//
+// Example:
+//
+//	value := NewValue(t, map[string]interface{}{"foo": 123})
+//	value.IsObject()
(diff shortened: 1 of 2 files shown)


gavv:
Thanks, the changes look good!

Sorry, seems I forgot to mention in the issue: in httpexpect each Is method should have counterpart Not method. For example, for IsObject, we also need NotObject. Could you please add them too?

nhAnik:
Sure, I will add them.

gavv:
LGTM!

gavv:
BTW, welcome to our [discord chat](https://discord.gg/5SCPCuCWA9)!

nhAnik:
Thank you 🙂

PR #266, opened 2023-02-01
Add lazy reading of response body

### LINKED ISSUE #244: Lazy reading of response body
Currently, response body is read entirely and closed inside Response constructor. This prevents Response to work with stream responses, e.g. SSE.

To fix this, Response should not read entire body in constructor, but instead read it only on demand when it's needed first time, e.g. when the user invoked Text() or JSON(). This way it'll continue to work as before for regular responses and wont break with stream responses, given that the user does not try to invoke inappropriate methods on them.

We need to find all places when `r.content` is used in response.go and replace it with a function call which will check if response body is already read, and read it if necessary.

We should add unit test for Response that checks that body is not read until a call to Text/JSON/etc.

chunweii:
Fixes #244


request_test.go
```diff
@@ -843,7 +843,7 @@ func TestRequest_BodyChunked(t *testing.T) {
 	assert.Equal(t, "METHOD", client.req.Method)
 	assert.Equal(t, "url", client.req.URL.String())
 	assert.Equal(t, make(http.Header), client.req.Header)
-	assert.Equal(t, "body", string(resp.content))
+	assert.Equal(t, "body", string(resp.getContent(resp.chain)))
```
gavv, line 846: `resp.getContent(resp.chain)` is not quite correct usage.

getContent may call chain.fail, and chain.fail requires to be called between chain.enter / chain.leave, i.e. inside an operation. If you try to call it outside of operation, it will panic instead of reporting failure.

I suggest to replace:

```go
string(resp.getContent(resp.chain))
```

with:

```go
resp.Body().Raw()
```

and replace:

```go
bytes.NewReader(resp.getContent(resp.chain))
```

with:

```go
strings.NewReader(resp.Body().Raw())
```

response.go
```diff
@@ -30,6 +30,8 @@ type Response struct {
 
 	content []byte
 	cookies []*http.Cookie
+
+	hasRead bool
```
gavv, line 34: Let's rename it to something more specific, maybe `contentReceived`?

response.go
```diff
@@ -159,6 +160,14 @@ func getResponseContent(opChain *chain, resp *http.Response) []byte {
 	return content
 }
 
+func (r *Response) getContent(opChain *chain) []byte {
+	if !r.hasRead {
+		r.content = getResponseContent(opChain, r.httpResp)
```
gavv, line 165: getResponseContent is invoked only from getContent, and getContent is very small. So I suggest to merge getResponseContent into getContent.

response.go
```diff
@@ -159,6 +160,14 @@ func getResponseContent(opChain *chain, resp *http.Response) []byte {
 	return content
 }
 
+func (r *Response) getContent(opChain *chain) []byte {
```
gavv, line 163: getContent may report failures. Previously it was invoked in constructor, and if it reported failure, all subsequent method calls became no-op because each method has this check at the beginning:

```go
	if opChain.failed() {
		return
	}
```

Now getContent is not invoked in constructor, and failure may happen in the middle of a method when it invokes getContent. We need to handle this failures: if getContent fails, we should abort method early.

Let's add bool flag to getContent:

```go
func (r *Response) getContent(opChain *chain) ([]byte, bool)
```

And everywhere where content is needed, use it like this:

```go
func (r *Response) Body() *String {
	opChain := r.chain.enter("Body()")
	defer opChain.leave()

	if opChain.failed() {
		return newString(opChain, "")
	}

	content, ok := r.getContent(opChain)
	if !ok {
		return newString(opChain, "")
	}

	return newString(opChain, string(content))
}
```

Here we make r.getContent call early right after opChain.failed() check, and we exit from method if it fails.

gavv: We will also need to add a test that checks how all these methods (Body, Text, Form, etc) handle getContent failures.

This test should set body to a reader that will fail, then construct response, call corresponding method (e.g. Body) and check that it returned non-nil but empty value and that the chain both in response and in returned value is marked as failed.

gavv (commented):
Thanks for PR! Please see comments below.

gavv:
Thanks for update! I'll review the code tomorrow, but here is one preliminary comment: we should wrap body into newBodyWrapper in response constructor, not in getContent. Because it's not guaranteed that getContent will be ever called. And adding newBodyWrapper is needed to ensure that if getContent is not called, the finalizer will close the body. (bodyWrapper has finalizer inside it).

response_test.go
```diff
@@ -114,6 +114,60 @@ func TestResponse_Constructors(t *testing.T) {
...
+
+	t.Run("Constructor does not read http response body first", func(t *testing.T) {
+		reporter := newMockReporter(t)
+		resp := NewResponse(reporter, &http.Response{
+			StatusCode: http.StatusOK,
+			Body:       ioutil.NopCloser(bytes.NewBufferString("body")),
+		})
+		assert.Nil(t, resp.content)
+		assert.False(t, resp.contentReceived)
+	})
```
gavv, line 126: Isn't this test duplication of TestResponse_ResponseBodyLazyRead?

response_test.go
```diff
@@ -114,6 +114,60 @@ func TestResponse_Constructors(t *testing.T) {
...
+	t.Run("getContent failures handled by NoContent()", func(t *testing.T) {
+		reporter := newMockReporter(t)
+		body := newMockBody("body string")
+		body.readErr = errors.New("read error")
+		resp := NewResponse(reporter, &http.Response{
+			StatusCode: http.StatusOK,
+			Body:       body,
+		})
+		resp.NoContent().chain.assertFailed(t)
+	})
```
gavv, line 170: Aren't these two tests duplication of tests below?

response_test.go
```diff
@@ -857,6 +961,32 @@ func TestResponse_FormBadType(t *testing.T) {
...
+	httpResp := &http.Response{
+		StatusCode: http.StatusOK,
+		Header:     http.Header(headers),
+		Body:       body,
+	}
+
+	resp := NewResponse(reporter, httpResp)
+
+	respForm := resp.Form()
+	assert.True(t, respForm.Raw() == nil)
```
gavv, line 983: Nitpick: can be simplified to:

```
assert.Nil(t, respForm.Raw())
```

response_test.go
```diff
@@ -971,6 +1101,31 @@ func TestResponse_JSONCharsetBad(t *testing.T) {
...
+	httpResp := &http.Response{
+		StatusCode: http.StatusOK,
+		Header:     http.Header(headers),
+		Body:       body,
+	}
+
+	resp := NewResponse(reporter, httpResp)
+
+	respJSON := resp.JSON()
+	assert.True(t, respJSON.Raw() == nil)
```
gavv, line 1123: Nitpick: can be simplified to:

```
assert.Nil(t, respJSON.Raw())
```

response_test.go
```diff
@@ -553,27 +609,50 @@ func TestResponse_NoContentNil(t *testing.T) {
...
+			StatusCode: http.StatusOK,
+			Header:     http.Header(headers),
+			Body:       body,
+		}
+
+		resp := NewResponse(reporter, httpResp)
+
+		respNoContent := resp.NoContent()
+		respNoContent.chain.assertFailed(t)
+		resp.chain.assertFailed(t)
```
gavv, line 654: Nitpick: respNoContent var is not needed, because NoContent returns original resp.

gavv (commented):
A few minor comments.

chunweii:
Sorry for the test case duplication. My code is ready for review again, @gavv

gavv:
Thanks for update! I've added a few small fixes:

* replaced contentReceived boolean with contentState enum; in addition to two existing states (not read; successfully read) I added a third state indicating that a read was attempted but failed; in this case we also don't want to read body again

* covered this addition with tests

* improved mock body a bit for better checks in tests of what was called; now wee test how many times read and close were called

Issue #272, opened 2023-02-02
Move e2e tests to separate package


gavv:
We have a set of end-to-end tests in files named `e2e_xxx_test.go`. These tests are intended to be blackbox, i.e. they should use only public API of httpexpect and should not access private fields.

In this issue we want to move all such tests into a separate subdirectory `e2e`, and accordingly separate package (but same module).

Some e2e tests violate blackbox rule and access private fields. These tests should be fixed. For example, if a test accesses chain field to invoke assertNotFailed or assertFailed, the test can be reworked to use mock reporter. Instead of checking chain field directly, it can provide custom reporter and then check if a failure was reported or not.

After completing the task, we should also do the following:

* [x] add e2e tests to Makefile targets
* [x] add e2e tests to CI (building, linting, running)
* [x] on CI, merge test coverage from unit and e2e test; useful links: [[1]](https://stackoverflow.com/questions/63842225/go-test-coverage-over-different-packages), [[2]](https://github.com/wadey/gocovmerge)

regularcoder:
I'm interested in signing up for this.

gavv:
Great, thanks!

regularcoder:
Hello, yes I do, apologies for the delay. I will try to get it done this weekend - by 2nd April.

gavv:
@regularcoder thanks, no worries, just checking if it's abandoned,  which happens often. Take your time :)

gavv:
Follow-up commits:

* 7c851155bb17c148cb24266cb96264dac7a0957b - add e2e to makefile targets
* 1965b0891c6159e8cb84e6101bd1720635be79cb - add e2e to CI steps

gavv:
One more:

* fc338613061e29bb244318e39b7df391f5686928 - add e2e to coverage reports

gavv:
Final commit: 907b2e90d49485f6605a5fe824901cc0c92830af

gavv:
Seems now everything is done. @regularcoder Thanks!

PR #275, opened 2023-02-03
Move e2e tests to separate package (#272)

regularcoder:


regularcoder:
@gavv, raised a WIP PR, would like opinions/inputs on a few things before I proceed to refactor remaining tests:

- Moved mockPrinter to e2e/e2e_printer_test.go since nobody else uses it. Alternatively I can move it to a new internal/mocks package.
- Changed e2e/e2e_context_test.go to use DefaultAssertionHandler with DefaultFormatter so that there is no dependency on mockAssertionHandler. Alternatively I can move mockAssertionHandler to a new internal/mocks package.
- Created custom booleanReporter for e2e/e2e_redirect_test.go and broke one larger test into numerous sub-tests.

gavv:
Hi, thanks for PR!

> @gavv, raised a WIP PR, would like opinions/inputs on a few things before I proceed to refactor remaining tests:
> 
>    * Moved mockPrinter to e2e/e2e_printer_test.go since nobody else uses it. Alternatively I can move it to a new internal/mocks package.
> 
>    * Changed e2e/e2e_context_test.go to use DefaultAssertionHandler with DefaultFormatter so that there is no dependency on mockAssertionHandler. Alternatively I can move mockAssertionHandler to a new internal/mocks package.
> 
>    * Created custom booleanReporter for e2e/e2e_redirect_test.go and broke one larger test into numerous sub-tests.

All these changes look good.

I don't think we need internal/mocks because many mocks depend on httpexpect types and it's not easy to move them.

e2e/e2e_basic_test.go
```diff
@@ -1,8 +1,9 @@
-package httpexpect
+package e2e
 
 import (
 	"crypto/tls"
 	"encoding/json"
+	"github.com/gavv/httpexpect/v2"
 	"net/http"
 	"net/http/httptest"
 	"testing"
```
gavv, line 9: Nitpick: let's move httpexpect import to the second import block, like this:

```
import (
	"crypto/tls"
	"encoding/json"
	"net/http"
	"net/http/httptest"
	"testing"

	"github.com/gavv/httpexpect/v2"
	"github.com/valyala/fasthttp/fasthttpadaptor"
)
```

Here and in other files.

e2e/e2e_context_test.go
```diff
@@ -438,15 +439,17 @@ func TestE2EContext_PerRequestRetry(t *testing.T) {
...
 
-		e := WithConfig(Config{
-			BaseURL:          ts.URL,
-			AssertionHandler: &mockAssertionHandler{},
+		e := httpexpect.WithConfig(httpexpect.Config{
+			BaseURL: ts.URL,
+			AssertionHandler: &httpexpect.DefaultAssertionHandler{
+				Formatter: &httpexpect.DefaultFormatter{},
+			},
 		})
```
gavv, line 447: Good catch, actually using mockAssertionHandler here looks like a bug because it will ignore failures.

Thought, new usage is not quite correct too: DefaultAssertionHandler requires Reporter field to be non-nil. Otherwise instead of reporting failure it will panic. I pushed a commit to catch this kind of misuses earlier: 67c5e0aa9eff1214259c54080aa63e5e2baf3d38

We could set Reporter field to `t`, but even better instead of setting `Config.AssertionHandler` we can just set `Config.Reporter` to `t`. In this case httpexpect will automatically use DefaultAssertionHandler with proper fields.

So it would be:

```
		e := httpexpect.WithConfig(httpexpect.Config{
			BaseURL: ts.URL,
			Reporter: t,
		})
```

e2e/e2e_redirect_test.go
```diff
@@ -1,14 +1,25 @@
...
 	"net/http/httptest"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/valyala/fasthttp"
 )
 
+type booleanReporter struct {
+	failed bool
+}
```
gavv, line 16: Maybe rename it to mockReporter? booleanReporter sounds a bit confusing to me. I think it's not a big deal that we have another mockReporter in another package.

e2e/e2e_basic_test.go
```diff
@@ -1,8 +1,9 @@
-package httpexpect
+package e2e
```
gavv, line 1: I think we can remove `e2e_` prefix from file names, it's redundant now.

gavv:
https://github.com/gavv/httpexpect/issues/272

e2e/e2e_context_test.go
regularcoder: Thank you for the detailed feedback - incorporated your comments. Will work on the remainder files.

Will also check if there is a way to get a formatter or linter to order the imports in the way we want.

gavv: > Will also check if there is a way to get a formatter or linter to order the imports in the way we want.

Oh, it would be perfect.

gavv: BTW welcome to [discord chat](https://discord.gg/5SCPCuCWA9) if you'll want to discuss something :)

gavv:
Seems that PR is slowly accumulating conflicts. If you wish, feel free to complete this PR (i.e. resolve conflicts, fix CI, and address remaining issues, if any) for only part of the tests, and we can merge it, postponing the rest to future PRs.

PR #278, opened 2023-02-05
Feature: Switch from net.Error.Temporary to net.Error.Timeout

### LINKED ISSUE #270: Switch from net.Error.Temporary to net.Error.Timeout
Currently httpexpect.Request has 4 retry policies:

* DontRetry
* RetryTemporaryNetworkErrors
* RetryTemporaryNetworkAndServerErrors
* RetryAllErrors

RetryTemporaryNetworkErrors and RetryTemporaryNetworkAndServerErrors use net.Error.Temporary to identify temporary network errors (see Request.shouldRetry). This method were deprecated in recent go versions:

```go
// An Error represents a network error.
type Error interface {
	error
	Timeout() bool // Is the error a timeout?

	// Deprecated: Temporary errors are not well-defined.
	// Most "temporary" errors are timeouts, and the few exceptions are surprising.
	// Do not use this method.
	Temporary() bool
}
```

Hence, we should migrate from Temporary to Timeout, but without breaking compatibility.

Steps:

* introduce two new retry policies: RetryTimeoutErrors, RetryTimeoutAndServerErrors
* RetryTimeoutErrors should work like RetryTemporaryNetworkErrors, but it should use net.Error.Timeout
* RetryTimeoutAndServerErrors should work like RetryTemporaryNetworkAndServerErrors, but it should use net.Error.Timeout
* mark RetryTemporaryNetworkErrors and RetryTemporaryNetworkAndServerErrors as deprecated and suggest to use corresponding new policies instead
* update tests to work with new policies; exclude deprecated policies from tests because these tests are huge already

We will remove deprecated policies in v3.

kcaashish:
Closes #270

### DIFF
--- a/request.go
+++ b/request.go
@@ -579,16 +579,9 @@ const (
 	DontRetry RetryPolicy = iota
 
 	// Deprecated: use RetryTimeoutErrors instead.
-	//
-	// RetryTemporaryNetworkErrors enables retrying only temporary network errors.
-	// Retry happens if Client returns net.Error and its Temporary() method
-	// returns true.
 	RetryTemporaryNetworkErrors
 
 	// Deprecated: use RetryTimeoutAndServerErrors instead.
-	//
-	// RetryTemporaryNetworkAndServerErrors enables retrying of temporary network
-	// errors, as well as 5xx status codes.
 	RetryTemporaryNetworkAndServerErrors
 
 	// RetryTimeoutErrors enables retrying of timeout errors.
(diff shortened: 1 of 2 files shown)


gavv:
Thank you!

gavv:
I forgot that e2e test also should be updated. Fixed: 1d92ad28c793a887b790d3dc5b0ce4f7ed9f322e

kcaashish:
I didn't know about the e2e tests, thank you for covering up for me.

PR #280, opened 2023-02-06
Modified chain.leave() to handle concurrancy

Rohith-Raju:
Signed-off-by: ROHITH RAJU <rohithraju488@gmail.com>
Merge this to #254 to make `environment` thread-safe.


Rohith-Raju:
@gavv could you take a look at this?

gavv:
The patch looks good now, thanks!

We also need the test for chain discussed in previous PR.

gavv:
Oh, you already mentioned the test. Cool, thanks.

gavv:
Oh, you already mentioned the test. Cool, thanks.

chain.go
```diff
@@ -413,19 +408,12 @@ func (c *chain) fail(failure AssertionFailure) {
...
-
-		if chainValidation {
-			if err := validateAssertion(&failure); err != nil {
-				panic(err)
-			}
+	if chainValidation {
+		if err := validateAssertion(&failure); err != nil {
+			panic(err)
 		}
 	}
```
gavv, line 418: Minor issue: let's move this piece to leave(), after Failure() call.

It's more convenient to panic on ill-formed failure after reporting it, not before, because this way we'll first see the report itself (assertion handler will print failure) and then will see the panic. Otherwise we'll see only panic.

gavv:
https://github.com/gavv/httpexpect/issues/233

Rohith-Raju:
@gavv the tests are a rough sketch.

gavv:
LGTM

gavv:
Thanks.

I've rebased PR on fresh master and added a few cosmetic fixes:

* comments
* reordering of some fields / vars to follow the same order everywhere
* refactoring of tests

Issue #155, opened 2022-11-23
Implement Number IsInt, IsUint, IsFloat, IsNan


gavv:
Implement new `Number` assertions:

* `IsInt` - succeeds if number is signed or unsigned integer (does not have fractional part)
* `IsUint` - succeeds if number is unsigned integer (does not have fractional part and is >= 0)
* `IsFloat` - succeeds if number is not integer (have fractional part, is opposite to `IsInt`)
* `IsNan` - succeeds if number is NaN

Each of them, except `IsNan`, should have optional argument `bits` which defines maximum allowed bitness for the number, e.g. `IsInt(32)` means that the number should fit in 32-bit signed integer. See `String.AsNumber` for en example of optional argument.

For every assertion, also add negative version (`NotInt`, `NotUint`, etc).

Each assertion should have documentation comment with example and be covered by unit tests. Don't forget to cover corner cases like NaN, inf, etc.

deepto98:
I'd like to give this a shot. Can I pick this up?

gavv:
Sure, thanks!

gavv:
@deepto98 Hi, do you still have plans on this?

gavv:
Unassigning, so that someone could pick this up.

fahimbagar:
@gavv, before I continue, could you check the WIP PR and see whether the solution is fine? 

I haven't finished `IsFloat` and `IsNaN`, need to clarify something
1. The current solution for integer is to convert them to string first and then use `ParseInt`. Is this correct?
    - I am wondering why the bit is too flexible. I mean something like this could easily check whether it's integer `math.Mod(floatValue, 1.0) == 0`. Or maybe limit it to 32 and 64. Wdyt?
2. `strconv.ParseFloat` recognizes `NaN` and `Inf`. Should it be failed on `IsFloat` case?
> ParseFloat recognizes the string "NaN", and the (possibly signed) strings "Inf" and "Infinity" as their respective special floating point values. It ignores cases when matching.
https://pkg.go.dev/strconv#ParseFloat

gavv:
@fahimbagar thanks for taking up this. Will reply shortly.

gavv:
Hi, here are some thoughts.

1. Using fmt.Sprintf is an interesting idea, and probably is correct, but instead I'd suggest to use `math.Modf` and check that the value belongs to desired range (e.g. `[0; 2^bits]`). Rationale:

[...]

PR #282, opened 2023-02-07
Number type assertion

fahimbagar:
This PR does the following:
- [x] `IsInt` and `NotInt`
- [x] `IsUint` and `NotUint`
- [ ] ~`IsFloat` and `NotFloat`~ Not supported
- [ ] ~`IsNaN` and `NotNaN`~ Not supported
- [ ] ~`IsInf` and `NotInf`~ Not supported
- [x] `IsFinite` and `NotFinite`
- [x] Use `big.Int` and `big.Float`


gavv:
https://github.com/gavv/httpexpect/issues/155

gavv:
Thanks for update, I'll review it tomorrow.

One preliminary comment: for comparison with min/max int, we should switch from floating point to integer math. Instead of pow, you can use but shifts.

Floats can't represent integers larger than 2^52 without losing precision. Could you please add a test with big integer? It will fail with current code, but will be fixed after switching to integer math.

fahimbagar:
> One preliminary comment: for comparison with min/max int, we should switch from floating point to integer math. Instead of pow, you can use but shifts.

I've tried to use shifting for the bit size but it doesn't work. Could you check what do I miss with this shift playground? https://go.dev/play/p/BrKAjFvs4IE

> Could you please add a test with big integer? It will fail with current code, but will be fixed after switching to integer math.

Noted.

gavv:
https://go.dev/play/p/wlauK6mJlqN

gavv:
Also, could you add a check that bits has valid value? I.e. from 1 to 64.

fahimbagar:
Ah thanks for the help, @gavv. Sure, I'll add the checker for bit value.

gavv:
BTW note the casts. We need to be sure that we're shifting unsigned ints to avoid messing up with the sign bit.

number.go
```diff
@@ -618,3 +619,236 @@ func (n *Number) Le(value interface{}) *Number {
 
 	return n
 }
+
+// IsInt succeeds if number is integer.
+//
+// If bit is omitted, uses 64 bit as default. Otherwise, uses given bit.
+// Bit defines maximum allowed bitness for the given number.
```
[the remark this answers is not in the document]
gavv: Also should mention that bits should be in range `[1; 64]`

gavv (commented):
Here are some comments:

fahimbagar:
@gavv, help me to review this PR again, thanks

gavv:
Hi, using big.Float is a good idea! Is it possible to also switch integer math from 64 bit to big.Int, could you check? 

I was thinking about doing this later, after we finish migration of Number to big nums. But since you're already using them, we can implement it right now I guess. 

In this case shifts or powers would be done by big.Int methods, and limitation for maximum value of 64 for bits would be removed.

fahimbagar:
> Hi, using big.Float is a good idea!

Thanks for your help as well :smile:.

> Is it possible to also switch integer math from 64 bit to big.Int, could you check?

I'll check .... I haven't gotten used to this `big` package. :sweat_smile: 


1 more thing, do we need `IsInf` as well?

gavv:
> 1 more thing, do we need IsInf as well?

Probably IsInf is not very useful. While you can often encounter NaN, usually indicating a bug, inf is more rare and often doesn't indicate a bug, and thus is not so special. 

But. I checked math lib from python: https://docs.python.org/3/library/math.html

It has isnan, isinf, isfinite. IsFinite could definitely be useful - when we want to check that we have a "regular" number, not inf or nan.

And if we'll add IsFinite, then it's logical to add IsInf for completeness. Because IsFinite = !IsNan && !IsInf.

So, I think we can add IsFinite and IsInf. You can do it in this pr, or I can create a separate issue.

fahimbagar:
I see. Thanks for the explanation.

I'll handle it in this PR as well if that's okay for you.

fahimbagar:
@gavv, I've migrated most of math int to `big.Int`, not all since NaN is not handled well in the big package, and also I can't find the correct method to check fractional for `Uint`.

Let me know if I'm missing something there.

gavv:
Thanks for update! Will review in upcoming days.

gavv:
Since we switched to big.Float in our checks, we should remove limitation of maximum 64 bit width. Let's remove checks and update comments. By default, let's allow unlimited width.

gavv:
> and also I can't find the correct method to check fractional for Uint

What went wrong with uint?

gavv:
Given it one more thought.. I think we should remove IsFloat/NotFloat. Its use case is a bit strange: allow floats with non-zero fractional part, but disallow zero fractional part. It's more likely that user would want to check that number is either int or float, but not nan and inf. And for this, we now have IsFinite.

gavv:
Nitpick: I'd place IsNan and IsInf after IsFinite.

gavv:
Actually it seems that IsFinite is much more useful than IsNan and IsInf, and is what the user need in most cases - just to check that we have a "normal" number.

So... I suggest to remove IsNaN and IsInf too, to keep API more compact. The use case is rare, and there is equivalent in form IsEqual/NonEqual.

Sorry that I did not realize this earlier before you spent your time on them.

gavv:
Besides these comments, the patch looks good!

fahimbagar:
> > and also I can't find the correct method to check fractional for Uint
> 
> What went wrong with uint?

`Uint64` behaves weirdly when it encounters a decimal point, therefore fractals can't be caught in:
https://github.com/gavv/httpexpect/blob/5ad39212d0d3eb5d827deb4fb2e3fc094707819d/number.go#L872-L873

That's why I resort to:
https://github.com/gavv/httpexpect/blob/5ad39212d0d3eb5d827deb4fb2e3fc094707819d/number.go#L858-L860

Let me adjust it, it feels stutter somehow to still use `math.Modf` only for `Uint` but not for `Int`.


> Actually it seems that IsFinite is much more useful than IsNan and IsInf, and is what the user need in most cases - just to check that we have a "normal" number.
> 
> So... I suggest to remove IsNaN and IsInf too, to keep API more compact. The use case is rare, and there is equivalent in form IsEqual/NonEqual.
> 
> Sorry that I did not realize this earlier before you spent your time on them.

Thanks for the review. It's okay, I'm happy to help leverage `httpexpect` usability. I'll address it.

gavv:
But why do we need `big.Float.Uint64`? It can't handle large numbers anyway. Can't we just use `big.Float.Int` and check that it's not negative and is not out of bounds (if max bit width is set)?

fahimbagar:
> But why do we need `big.Float.Uint64`? It can't handle large numbers anyway. Can't we just use `big.Float.Int` and check that it's not negative and is not out of bounds (if max bit width is set)?

Yeah, that's what I just realize. That's why I need to modify Uint.
I'll let you know.

fahimbagar:
Added:
- Remove `IsInf`, `NotInf`, `IsNaN`, `NotNaN`
- Remove limitation of 64-bit width from Uint and Int
- Remove `IsFloat` and `NotFloat`
- Remove `math.modf` from `Uint` functions

@gavv, let me know if I'm missing something

gavv:
We should update handling of omitted bits arg:

* in comments, remove restriction of maximum value of 64
* if bits is omitted in IsInt, don't set it to 64; instead, omit boundary checks
* if bits is omitted in IsUint, replace boundary check with a check that value is non negative

gavv:
NotUint still have Modf check. Is it there accidentally?

fahimbagar:
> 

Right, I missed this. Addressed in https://github.com/gavv/httpexpect/pull/282/commits/d8ee5d1a9437e945441c5e9fdd6b9ef88f094187


> We should update handling of omitted bits arg:
> 
> * in comments, remove restriction of maximum value of 64
> * if bits is omitted in IsInt, don't set it to 64; instead, omit boundary checks
> * if bits is omitted in IsUint, replace boundary check with a check that value is non negative

Addressed in https://github.com/gavv/httpexpect/pull/282/commits/b0d2aceb13bbf80a55527fb8a36f7dee98e70ea5 and https://github.com/gavv/httpexpect/pull/282/commits/bae45fc6de4637a354074a6892ed0c5ff5e51ef5.

number.go
```diff
@@ -618,3 +620,376 @@ func (n *Number) Le(value interface{}) *Number {
...
+		imin.Neg(imax)
+		imin.Sub(imin, big.NewInt(1))
+		if inum.Cmp(imin) < 0 || inum.Cmp(imax) > 0 {
+			opChain.fail(AssertionFailure{
+				Type:   AssertInRange,
+				Actual: &AssertionValue{n.value},
+				Expected: &AssertionValue{AssertionRange{
+					fmt.Sprintf("-2^%d", bitSize-1),
+					fmt.Sprintf("2^%d", bitSize-1),
+				}},
```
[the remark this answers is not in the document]
gavv: Also I suggest to improve reporting this ranges like this:

```go
Expected: &AssertionValue{AssertionRange{
    Min: intBoundary{imin, -1, bitSize-1},
    Max: intBoundary{imax, +1, bitSize-1},
}}

....

type intBoundary struct {
    val *big.Int
    sign int
    bits int
}

func (b intBoundary) String() string {
    if sign > 0 {
        return fmt.Sprintf("2^%d-1 (%s)", b.bits, val)
    } else {
        return fmt.Sprintf("-2^%d (%s)", b.bits, val)
    }
}
```

Explanations:

* we will give the user both logical form (2^N) and actual numeric value
* we will reuse this logic for all functions
* since intBoundary implements Stringer interface, it will be handled specially by httpexpect.DefaultFormatter: it will omit quotes when including it in error messages; i.e., instead of `"-2^7 (-128)"` we will see `-2^7 (-128)`, which looks prettier

number.go
```diff
@@ -618,3 +620,376 @@ func (n *Number) Le(value interface{}) *Number {
...
+				Errors: []error{
+					fmt.Errorf("expected: number is %d-bit integer", bitSize),
+				},
```
gavv, line 699: Let's clarify this to: `expected: number is %d-bit signed integer`

Because signedness  reduces allowed range.

Let's add this clarification ("signed integer" or "unsigned integer") to all error messages.

number.go
```diff
@@ -618,3 +620,376 @@ func (n *Number) Le(value interface{}) *Number {
...
+			Actual: &AssertionValue{n.value},
+			Errors: []error{
+				errors.New("expected: number is unsigned integer"),
+			},
+		})
+		return n
+	}
+
+	inum, acc := big.NewFloat(n.value).Int(nil)
+	if !(acc == big.Exact && inum.IsUint64()) {
```
gavv, line 827: IsUint64() check forbids larger integers, but we want to handle them. We should remove this check.

Instead, when bits is omitted, we should check that value is non-negative (because we want *unsigned* integer).

gavv: Same for NotUint

number.go
```diff
@@ -618,3 +620,376 @@ func (n *Number) Le(value interface{}) *Number {
...
+//	number.IsFinite()
+func (n *Number) IsFinite() *Number {
+	opChain := n.chain.enter("IsFinite()")
+	defer opChain.leave()
+
+	if opChain.failed() {
+		return n
+	}
+
+	if !(math.IsInf(n.value, 0) || math.IsNaN(n.value)) {
```
gavv, line 952: The check should be reversed:

```
if math.IsInf(n.value, 0) || math.IsNaN(n.value) {
```

gavv: The check in NotFinite should be reversed too.

number_test.go
```diff
@@ -584,3 +590,177 @@ func TestNumber_ConvertLesser(t *testing.T) {
 	value.chain.assertFailed(t)
 	value.chain.clearFailed()
 }
+
+func TestNumber_IsInt(t *testing.T) {
```
gavv, line 594: Let's check how `IsInt()`, `NotInt()`, `IsInt(32)`, `NotInt(32)` work for the following boundary cases:

```go
-2^31 - 1
-2^31
-2^31 + 1
2^31 - 1
2^31
2^31 + 1
```

This would give us confidence that bitness checks are fully correct.

I suggest to replace subtests `without bit size` and `with bit size` with a single subtest that checks the values above.

We can also add a similar set of checks for other bitness, e.g. 16.

number_test.go
```diff
@@ -584,3 +590,177 @@ func TestNumber_ConvertLesser(t *testing.T) {
...
+			chain.assertFailed(t)
+
+		NewNumber(reporter, math.MaxInt64).NotInt(32).
+			chain.assertNotFailed(t)
+
+		NewNumber(reporter, math.MaxInt8).NotInt(32).
+			chain.assertFailed(t)
+	})
+
+	t.Run("check edge cases", func(t *testing.T) {
```
gavv, line 639: Let's ensure that +inf and -inf is covered for both IsInt and NotInt in this subtest.

Also, I'd rename this test to something like "non-numbers".

number_test.go
```diff
@@ -584,3 +590,177 @@ func TestNumber_ConvertLesser(t *testing.T) {
...
+
+	t.Run("big integer cases", func(t *testing.T) {
+		t.Skip("skipped until migration to big Float is done")
+
+		NewNumber(reporter, math.MaxInt64).IsInt().
+			chain.assertNotFailed(t)
+
+		NewNumber(reporter, math.MaxInt64).NotInt(64).
+			chain.assertFailed(t)
+	})
```
gavv, line 667: We can remove tests for 64 bit ints at all.

Tests for lower bitnesses (32, 16, etc) will fully cover functionality of IsInt, IsUnt, etc, because they are not really tied to specific bitness.

After migration to big.Float we will revise all tests and decide where we need to cover more cases.

number_test.go
gavv: Similar for IsUnit test.

number_test.go
gavv: Similar for IsUnit test.

number_test.go
gavv: Similar for IsUnit test.

number_test.go
```diff
@@ -584,3 +590,177 @@ func TestNumber_ConvertLesser(t *testing.T) {
...
+
+		NewNumber(reporter, math.MaxUint64).IsUint().
+			chain.assertNotFailed(t)
+
+		NewNumber(reporter, math.MaxUint64).NotUint(64).
+			chain.assertFailed(t)
+	})
+}
+
+func TestNumber_IsFinite(t *testing.T) {
```
gavv, line 746: Let's cover both -inf and +inf here, and also both ints and floats.

gavv (commented):
Here are some final (I think) comments. Thanks for your patience :)

fahimbagar:
Hi @gavv, above feedback has been addressed. Let me know if I miss something.

I have removed `int64` limitation on Uint and NotUint as well.

gavv:
Perfect! Thanks for update.

I'm pushing a follow-up commit to this branch and merging it.

What I changed in my commit: a3b882422758f5118204973b32f7b953f88c004e

* Remove `value should have numeric ...` part from docs. That part is a copy-paste from other methods where it describes requirements for `value` argument, but new methods don't have such an argument.

* Expand code examples. I just feel that we should be a bit more verbose here.

* Add usage checks for non-positive bits argument. Seems we forgot about it.

* Replace AssertType with AssertValid, as it fits better. AssertType is more suitable when the value has wrong Go type, and it includes Go type in the assertion. AssertValid is a more generic check.

* Improve formatting in intBoundary. Use sign == 0 for zero boundary.

* Convert tests to table-driven style. When I started reviewing them, I realized that it's very hard to see what cases are covered because the tests are too verbose. Table-driven form is more declarative and helps a lot in this case.

* Add tests for invalid bitness.

PR #287, opened 2023-02-09
Add Array.IsValueEqual and Array.NotValueEqual

### LINKED ISSUE #286: Add Array.IsValueEqual and NotValueEqual
These two method should work similar to Object.IsValueEqual and NotValueEqual, but accept index instead of key.

New methods should have documentation comments with code examples, and be covered by unit tests.

nhAnik:
Closes #286

### DIFF
--- a/array.go
+++ b/array.go
@@ -1452,6 +1452,114 @@ func (a *Array) NotContainsOnly(values ...interface{}) *Array {
 	return a
 }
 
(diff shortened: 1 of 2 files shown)


gavv:
Thanks!

gavv:
Note: changed AssertEqual to AssertNotEqual in NotValueEqual, and reordered methods to have the same order as in Object.

PR #289, opened 2023-02-09
Made request thread safe

Rohith-Raju:
Thread safety for request struct 
#233 

- [x] Added lock and unlock methods
- [x] Added request_Reeterant test

Signed-off-by:_ ROHITH RAJU <rohithraju488@gmail.com>

### DIFF
--- a/request.go
+++ b/request.go
@@ -16,6 +16,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ajg/form"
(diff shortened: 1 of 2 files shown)


Rohith-Raju:
@gavv your thoughts?

Rohith-Raju:
> Request struct needs special care: it contains transforms and matchers callbacks invoked during Expect() call. We > should NOT call them under a lock because otherwise it would be easy to write code with deadlocks (if you use the same > > request from this callbacks).

Yes, I'm sorry. I totally ignored this. Won't happen again.

Rohith-Raju:
Deleted this as the changes are irrelevant.

PR #290, opened 2023-02-11
Implement Digit separator

### LINKED ISSUE #274: Implement DefaultFormatter.DigitSeparator
Implement new option DigitSeparator with four values:

* DigitSeparatorUnderscore (default)
* DigitSeparatorComma
* DigitSeparatorApostrophe
* DigitSeparatorNone

When DigitSeparatorUnderscore is used, integers and floats should be formatted with thousand separators like this:

```
1_000_000
3.141_592_653
```

DigitSeparatorComma and DigitSeparatorApostrophe are similar, but use `,` and `'` as separator, instead of `_`. DigitSeparatorNone corresponds to current behavior when no separator is used.

All numbers are formatted using DefaultFormatter.formatValue, so DigitSeparator should be handled there. It already has branches for float32 and float64; now we'll need to add branch for integers (we can use refIsNum function).

For an example of similar option and its handling, see DefaultFormatter.FloatFormat.

New option should have documentation. Formatter tests should be updated.

https://en.wikipedia.org/wiki/Decimal_separator
https://en.wikipedia.org/wiki/Integer_literal

harishbohara11:
fixes #274


gavv:

formatter.go
```diff
@@ -428,11 +449,15 @@ func (f *DefaultFormatter) fillDelta(
...
 	}
 
 	if flt := extractFloat64(value); flt != nil {
-		return f.formatFloatValue(*flt, 64)
+		return f.addDigitGrouping(f.formatFloatValue(*flt, 64))
+	}
+
+	if intg := extractInt(value); intg != nil {
+		return f.addDigitGrouping(strconv.Itoa(*intg))
 	}
```
gavv, line 461: Here we will catch only int, but we should handle other integers types too: uint, int8, int16, etc.

The easiest way to do it is:

* check if isNumber returns true; it means that it's an integer, because floats are already handled above
* if yes, format it using fmt.Sprintf with %v - this should be equivalent to how we format integers below
* apply digit groupping to that string, and return it

formatter.go
```diff
@@ -472,6 +497,68 @@ func (f *DefaultFormatter) formatFloatValue(value float64, bits int) string {
...
 }
 
+func (f *DefaultFormatter) addDigitGrouping(numStr string) string {
```
gavv, line 500: I suggest to rename this func to reformatNumber, and groupDigitsInString to insertDigitSeparator.

The first name suggests that we should call it when we're formatting any number, float or int. The caller doesn't bother what exactly it will do.

The second name correlates with the name of the DigitSeparator option.

formatter.go
```diff
@@ -472,6 +497,68 @@ func (f *DefaultFormatter) formatFloatValue(value float64, bits int) string {
...
+	var sign string
+	if numStr[0] == '-' || numStr[0] == '+' {
+		sign = string(numStr[0])
+		numStr = numStr[1:]
+	}
+
+	parts := strings.SplitN(numStr, ".", 2)
```
gavv, line 507: To make this code easier to follow, I suggest to extract a function, say, parseNumber, that returns all parts of the number: integer, fractional, and exponent.

Here we can call it, and apply groupping to integer and fractional parts.

Then we can compose number from parts.

formatter.go
```diff
@@ -472,6 +497,68 @@ func (f *DefaultFormatter) formatFloatValue(value float64, bits int) string {
...
+	intPart := f.groupDigitsInString(parts[0])
+
+	var fracPart string
+	var expPart string
+	if len(parts) == 2 {
+		parts = strings.SplitN(parts[1], "e", 2)
+		// using Reverse(parts[0]) because digit grouping in
+		// fractional part is required from Most Significant Bit
+		fracPart = "." + Reverse(f.groupDigitsInString(Reverse(parts[0])))
```
gavv, line 516: Instead of using Reverse func, I suggest to pass direction flag (-1 or +1) to groupDigitsInString. I think the code will be easier to follow on the caller side

formatter_test.go
```diff
@@ -587,6 +587,120 @@ func TestFormatter_FloatFormat(t *testing.T) {
 	}
 }
 
+func TestFormatter_DigitSeparator(t *testing.T) {
```
gavv, line 590: We also need to cover a few more cases:

* big integer part, small decimal part, non scientific 
* big integer part, small decimal part, scientific 
* big integer part, big decimal part, non scientific 
* big integer part, big decimal part, scientific 

By small decimal part, I mean 1 to 3 digits, by big - more than 3 digits.

formatter_test.go
```diff
@@ -587,6 +587,120 @@ func TestFormatter_FloatFormat(t *testing.T) {
...
+				&AssertionFailure{
+					Type:   AssertValid,
+					Actual: &AssertionValue{tc.value},
+				})
+			assert.Equal(t, tc.wantText, formatData.Actual)
+		})
+	}
+}
+
+func TestFormatter_DigitSeparatorType(t *testing.T) {
```
gavv, line 647: We also need a separate test that check how separator is applied to different fliatand integer types (float32, float64, int, uint)

gavv (commented):
Thanks for pr. Here are some comments

Issue #288, opened 2023-02-09
Add more canonization tests for Array


gavv:
Most httpexpect methods perform convert value to its canonical form before comparison. See docs: https://pkg.go.dev/github.com/gavv/httpexpect/v2#hdr-Value_equality

In array_test.go, tests for many Array methods have "canonization" subtest, which checks that corresponding methods performs that conversion. For example: TestArray_IsEqual/canonization, TestArray_ConsistsOf/canonization, and others.

We need to add similar "canonization" subtests to tests that are missing it:

* [x] TestArray_IsEqualUnordered
* [x] TestArray_InList
* [x] TestArray_ContainsAny
* [x] TestArray_ContainsOnly
* [ ] ~TestArray_Filter~ TestArray_Transform (produced array should be in canonical form)

Pull requests that address only part of this issue are welcome too!

Related: #283

Darrellbor:
Hey @gavv i'm new to open source and find this project interesting and would like to contribute by starting with this
i'm currently going through the codebase as well as the `HACKING.MD` file. if there's anything else i should be aware of, i'd appreciate

Darrellbor:
yes

Darrellbor:
@gavv I have pushed a PR that addresses `TestArray_IsEqualUnordered` please review and let me know if I'm on the right track so i can take corrections before pushing the rest

gavv:
@Darrellbor I've created a separate issue for Transform tests: https://github.com/gavv/httpexpect/issues/297 and described in more details what do we need to do. Please leave a comment there in case you would like to work on it.

gavv:
Closing this one, thank you.

PR #292, opened 2023-02-13
Canonization tests for arrays

Darrellbor:
PR for [issue 288](https://github.com/gavv/httpexpect/issues/288). 
Addresses Canonization tests:
- [x] `TestArray_IsEqualUnordered` 
- [x] `TestArray_InList`
- [x] `TestArray_ContainsAny`
- [x] `TestArray_ContainsOnly`
- [ ] `TestArray_Transform` (produced array should be in canonical form)

### DIFF
--- a/array_test.go
+++ b/array_test.go
@@ -1334,6 +1334,54 @@ func TestArray_ContainsOnly(t *testing.T) {
 		value.chain.assertFailed(t)
 		value.chain.clearFailed()
 	})
+
+	t.Run("canonization", func(t *testing.T) {
+		type (
+			myInt int
+		)
+
+		value := NewArray(reporter, []interface{}{123, 456, 456})
+
+		value.ContainsOnly(456.0)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.NotContainsOnly(456.0)
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+
+		value.ContainsOnly(myInt(123), 456.0)
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+
+		value.NotContainsOnly(myInt(123), 456.0)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.ContainsOnly(123.0, 456.0, 456.0)
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+
+		value.NotContainsOnly(123.0, 456.0, 456.0)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.ContainsOnly(myInt(123), myInt(456), 456.0)
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+
+		value.NotContainsOnly(myInt(123), 456.0, myInt(456))
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.ContainsOnly(myInt(123), "456", 456.0)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.NotContainsOnly(myInt(123), 456.0, "456")
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+	})
 }
 
 func TestArray_IsValueEqual(t *testing.T) {
--- a/array_test.go
+++ b/array_test.go
@@ -485,18 +485,6 @@ func TestArray_IsEqual(t *testing.T) {
 		value3.chain.clearFailed()
 	})
 
-	t.Run("nil check", func(t *testing.T) {
-		value := NewArray(reporter, []interface{}{123, 456})
-
-		value.IsEqual(nil)
-		value.chain.assertFailed(t)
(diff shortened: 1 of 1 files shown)


gavv:
Thanks for PR! The approach looks good.

I think we can reduce number of checks though. In canonization test, we don't need to check how IsEqualUnordered performs equality comparison (i.e. test combinations of different order, etc).

We assume that it's already covered by other subtests and everything is working, except conversion. And add a few tests for conversions.

Also please let me know if you want to create separate PRs for different tests, or one PR for all of them.

Darrellbor:
ok noted 
I would take the approach of having individual PR's for each of the canonization subtest to make it easy for review.

Darrellbor:
@gavv I've pushed commits that address 4 of the canonization tests and awaiting reviews although I am not entirely sure what to assert against with Filter, `Equal`, `Contains` or what?

gavv:
Regarding Filter: sorry, it was a typo. The function of interest is Transform. The thing to test is: when the new value returned from user func is not in canonical form, in the resulting array the corresponding element will be in canonical form.

array_test.go
```diff
@@ -604,78 +604,245 @@ func TestArray_IsEqualUnordered(t *testing.T) {
...
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.IsEqualUnordered(myArray{456.0, myInt(123.0), "foo"})
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+
+		value.NotEqualUnordered(myArray{456.0, myInt(123.0), "foo"})
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
```
gavv, line 633: This check is like previous one, but with changed order. We can assume that the ordering part of IsEqualUnordered is already tested and works well, and remove this check from canonization subtest. We need to test only for different types, but we don't need to test all permutations of them.

array_test.go
```diff
@@ -604,78 +604,245 @@ func TestArray_IsEqualUnordered(t *testing.T) {
...
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+
+		value.IsEqualUnordered(nil)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.NotEqualUnordered(nil)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
```
gavv, line 657: This is a good check, but I suggest to extract it into a separate subtest, like, "nil argument" or "invalid argument", because it's more like a special case.

I understand that you likely modeled it after TestArray_IsEqual/canonization, but actually we need to do the same there too (you can do it or I'll do it later, as you wish).

array_test.go
```diff
@@ -604,78 +604,245 @@ func TestArray_IsEqualUnordered(t *testing.T) {
...
+		duplicateValue.chain.assertNotFailed(t)
+		duplicateValue.chain.clearFailed()
+
+		duplicateValue.IsEqualUnordered(nil)
+		duplicateValue.chain.assertFailed(t)
+		duplicateValue.chain.clearFailed()
+
+		duplicateValue.NotEqualUnordered(nil)
+		duplicateValue.chain.assertFailed(t)
+		duplicateValue.chain.clearFailed()
```
gavv, line 697: I think that all these tests with duplicateValue are not needed. In canonization test, we don't need to bother about how it handles ordering, duplicates, etc.

array_test.go
gavv: Same for `array.InList(func() {})` and similar tests

array_test.go
gavv: Same for TestArray_ContainsOnly/canonization - the number of checks can be reduced I think.

gavv (commented):
Thanks for update!

Darrellbor:
@gavv noted 
I'd get to these later today.

array_test.go
Darrellbor: I've segmented the nil and invalid arguments into their own tests

array_test.go
Darrellbor: I've reduced the tests for these 
i.e `TestArray_IsEqualUnordered` and `TestArray_ContainsOnly`

array_test.go
Darrellbor: I've updated this subtest to remove these

Darrellbor:
> Regarding Filter: sorry, it was a typo. The function of interest is Transform. The thing to test is: when the new value returned from user func is not in canonical form, in the resulting array the corresponding element will be in canonical form.

This bit is not too clear for me so i haven't pushed any commits that address this part yet

gavv:
Thanks for update.

I've rebased this PR on fresh master and pushed a few small commits to this branch:

* unify invalid arguments tests; now each of them is called "invalid argument" and is the last subtest
* don't share reporter between subtests; reporter has mutable state affected by test, and it's better to isolate tests from each other
* fixed mess with `t` and `ts` vars in subtests (it was there before this PR)

gavv:
Merged this without changes for Transform, to avoid accumulating conflicts. Regarding transform, I'll reply in the issue. Thanks!

PR #293, opened 2023-02-15
 Wip : Make request thread safe

Rohith-Raju:
Initial commit to making request thread-safe

In this implementation, I Introduced a new request object whenever expect() is called to avoid data race. 
why expect?
Since expect() is the one that transfers the request object to matches and transformers.. we can supply a new Request object to them so that each go-routine has its own copy.

Things that are going wrong here are
-  While creating a new request object, we are also copying locks.
-  `checkOrder` relies on struct var `expectCalled` and since a new Request object is being created, `expectCalled` is not being updated and hence fails.
- `Request_Redirects` also fails

### DIFF
--- a/request.go
+++ b/request.go
@@ -16,6 +16,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ajg/form"
@@ -28,9 +29,9 @@ import (
 // Request provides methods to incrementally build http.Request object,
 // send it, and receive response.
 type Request struct {
-	noCopy noCopy
 	config Config
 	chain  *chain
+	mu     sync.RWMutex
 
 	redirectPolicy RedirectPolicy
 	maxRedirects   int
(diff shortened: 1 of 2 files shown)


Rohith-Raju:
Ping: @gavv

gavv:
Hi,

So... this patch is broken in many ways:

1. The patch is doing a shallow copy of Request, but there is still mutable state shared across these shallow copies: httpReq, transformers and matchers slices, multipart writer, etc, etc.  So this copying didn't actually make anything thread-safe.

1. On the other hand, Request object is not supposed to be copyable, as you can see from noCopy field. That's why some tests broke. I wonder why you didn't try to address the failures though.

1. There is a requirement that Expect can be called only once. Actually, trying to parallelize Expect calls is a dead-end approach in general, parallel executions of Expect should not be allowed at all.

1. As you correctly mentioned, copying locks is forbidden and is a bug. Again, I wonder why you didn't try to fix it before submitting PR.

BTW I'd recommend to do self-review of your code before submitting it. There are some clear copy-paste errors, like `var resp Request` or using `RWMutex` instead of `Mutex`. This is usually a bad sign for reviewer.

I'll do the rest of the task by myself. Thanks for PR, anyway.

Issue #284, opened 2023-02-08
Refactor some unit tests


gavv:
Some of our tests declare multiple variables and then do various checks on each variable, like this:

```go
value1 := NewMatch(reporter, []string{"m"}, nil)
value2 := NewMatch(reporter, []string{}, nil)

value1.IsEmpty()
value1.chain.assertFailed(t)
value1.chain.clearFailed()

value2.IsEmpty()
value2.chain.assertNotFailed(t)
value2.chain.clearFailed()
```

This approach is error-prone: it's easy to make a check on wrong variable and often it can remain unnoticed.

----

This tests need to be refactored in one of the following ways:

1. In simple tests, instead of declaring multiple vars, we can reuse a single one, like this:

    ```go
    var value *Match

    value = NewMatch(reporter, []string{"m"}, nil)
    value.IsEmpty()
    value.chain.assertFailed(t)
    value.chain.clearFailed()

    value = NewMatch(reporter, []string{}, nil)
    value.IsEmpty()
    value.chain.assertNotFailed(t)
    value.chain.clearFailed()
    ```

2. In more complicated tests, we can split the test into subtests, and isolate local vars there.

3. We can also convert tests to table-driven style.

----

Feel free to choose one of the approaches for each test based on your own judgement.

Tests to be refactored:

* [x] TestMatch_Empty
* [x] TestMatch_ValuesEmpty
* [x] TestNumber_EqualNaN
* [x] TestArray_IsEqual/types (split into subtests)
* [x] TestString_Empty
* [x] TestString_IsAscii
* [x] TestString_NotAscii
* [x] TestString_AsNumber
* [x] TestString_AsDateTime
* [x] TestValue_GetObject
* [x] TestValue_GetArray
* [x] TestValue_GetNumber
* [x] TestValue_GetBoolean
* [x] TestValue_IsObject
* [x] TestValue_IsArray
* [x] TestValue_IsString
* [x] TestValue_IsNumber
* [x] TestValue_IsBoolean
* [x] TestWebsocketMessage_CodeAndType
* [x] TestWebsocketMessage_NoContent

----

Pull requests that address only part of this issue are welcome too!

Related: https://github.com/gavv/httpexpect/issues/283

CarbonGabe:
Hi @gavv. Can I work on this issue?

gavv:
@gpjservais Sure, thanks! I'd suggest splitting work into a few PRs.

gavv:
[...]

PR #294, opened 2023-02-16
Refactor string unit tests

gpjservais:
This PR addresses refactoring the following unit tests from #284:
- [x] TestString_Empty
- [x] TestString_IsAscii
- [x] TestString_NotAscii
- [x] TestString_AsNumber
- [x] TestString_AsDateTime

I updated `TestString_IsAscii` and `TestString_NotAscii` to just reuse the same variable, `value` since it was simple and the formatting was similar to other tests. However, we could also update it to be one of the following formats instead:
* Format would be similar to TestString_AsBoolean and this could also make it easier to add additional test cases in the future
```go
func TestString_IsAscii(t *testing.T) {

	asciiValues := []string{
		"Ascii",
		string(rune(127)),
	}

	notAsciiValues := []string{
		"Ascii is アスキー",
		"アスキー",
		string(rune(128)),
	}

	for _, str := range asciiValues {
		t.Run(str, func(t *testing.T) {
			value := NewString(reporter, str)

			value.IsASCII()
			value.chain.assertNotFailed(t)
		})
	}

	for _, str := range notAsciiValues {
		t.Run(str, func(t *testing.T) {
			value := NewString(reporter, str)

			value.IsASCII()
			value.chain.assertFailed(t)
		})
	}
}
```

* Alternatively, we could use a more table-driven style.
```go
func TestString_IsAscii(t *testing.T) {

	cases := []struct {
		str  string
		fail bool
	}{
		{"Ascii", false},
		{string(rune(127)), false},
		{"Ascii is アスキー", true},
		{"アスキー", true},
		{string(rune(128)), true},
	}

	for _, value := range cases {
		t.Run(value.str, func(t *testing.T) {
			childValue := NewString(reporter, value.str)

			childValue.IsASCII()

			if value.fail {
				childValue.chain.assertFailed(t)
			} else {
				childValue.chain.assertNotFailed(t)
			}
		})
	}
}
```
  
I broke out `TestString_AsNumber` and `TestString_AsDateTime` into additional subtests to address the additional variables. If the subtest names should be updated, let me know.

We could also do something more table driven for those two tests as well. Like this:
```go
func TestString_AsNumber(t *testing.T) {

	cases := map[string]struct {
		str         string
		base        []int
		fail        bool
		expectedNum float64
	}{
		"default_base_integer": {
			str:         "1234567",
			fail:        false,
			expectedNum: float64(1234567),
		},
		"default_base_float": {
			str:         "11.22",
			fail:        false,
			expectedNum: float64(11.22),
		},
		"default_base_bad": {
			str:  "a1",
			fail: true,
		},
		"base10_integer": {
			str:         "100",
			base:        []int{10},
			fail:        false,
			expectedNum: float64(100),
		},
		"base10_float": {
			str:         "11.22",
			base:        []int{10},
			fail:        false,
			expectedNum: float64(11.22),
		},
		"base16_integer": {
			str:         "100",
			base:        []int{16},
			fail:        false,
			expectedNum: float64(0x100),
		},
		"base16_float": {
			str:  "11.22",
			base: []int{16},
			fail: true,
		},
		"base16_large_integer": {
			str:         "4000000000000000",
			base:        []int{16},
			fail:        false,
			expectedNum: float64(0x4000000000000000),
		},
		"default_float_percision_max": {
			str:  "4611686018427387905",
			fail: true,
		},
		"base10_float_percision_max": {
			str:  "4611686018427387905",
			base: []int{10},
			fail: true,
		},
		"base16_float_percision_max": {
			str:  "8000000000000001",
			base: []int{16},
			fail: true,
		},
		"base16_float_percision_min": {
			str:  "-4000000000000001",
			base: []int{16},
			fail: true,
		},
		"multiple_base": {
			str:  "100",
			base: []int{10, 16},
			fail: true,
		},
	}

	for name, value := range cases {
		t.Run(name, func(t *testing.T) {
			childValue := NewString(reporter, value.str)
			num := childValue.AsNumber(value.base...)

			if value.fail {
				childValue.chain.assertFailed(t)
				num.chain.assertFailed(t)
				assert.Equal(t, float64(0), num.Raw())
			} else {
				childValue.chain.assertNotFailed(t)
				num.chain.assertNotFailed(t)
				assert.Equal(t, value.expectedNum, num.Raw())
			}

		})
	}
}
```

Let me know if there's any issues. I plan to use separate PRs for refactoring the other tests mentioned in #284 after this one looks good.

### DIFF
--- a/string_test.go
+++ b/string_test.go
@@ -180,8 +180,6 @@ func TestString_Length(t *testing.T) {
 }
 
 func TestString_IsEmpty(t *testing.T) {
-	reporter := newMockReporter(t)
-
 	cases := map[string]struct {
 		str       string
 		wantEmpty bool
@@ -192,21 +190,20 @@ func TestString_IsEmpty(t *testing.T) {
 
 	for name, value := range cases {
 		t.Run(name, func(t *testing.T) {
-			childValue := NewString(reporter, value.str)
+			reporter := newMockReporter(t)
 
 			if value.wantEmpty {
-				childValue.IsEmpty()
-				childValue.chain.assertNotFailed(t)
+				NewString(reporter, value.str).IsEmpty().
+					chain.assertNotFailed(t)
 
-				childValue.NotEmpty()
-				childValue.chain.assertFailed(t)
+				NewString(reporter, value.str).NotEmpty().
+					chain.assertFailed(t)
 			} else {
-				childValue.IsEmpty()
-				childValue.chain.assertFailed(t)
-				childValue.chain.clearFailed()
+				NewString(reporter, value.str).IsEmpty().
+					chain.assertFailed(t)
 
-				childValue.NotEmpty()
-				childValue.chain.assertNotFailed(t)
+				NewString(reporter, value.str).NotEmpty().
+					chain.assertNotFailed(t)
 			}
 		})
 	}
@@ -484,8 +481,6 @@ func TestString_MatchInvalid(t *testing.T) {
 }
 
 func TestString_IsAscii(t *testing.T) {
-	reporter := newMockReporter(t)
-
 	cases := []struct {
 		str       string
 		wantASCII bool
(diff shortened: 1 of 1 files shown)


gavv:
Thanks for PR!

Table-driven style is much appreciated. If you're willing to spend some extra bit of time to make test table-driven, this option is almost always preferred. Actually many of our tests are not tables only because of laziness :)

The patch looks good, the suggested snippets look even better, let's apply them.

Regarding TestString_IsAscii, the second one looks better, and it would be also nice to merge tests for IsAscii and NotAscii into one.

gavv:
BTW also welcome to the [discord chat](https://discord.gg/5SCPCuCWA9)

gpjservais:
@gavv this is ready for review again.

I ended up merging `TestString_IsAscii` and `TestString_IsAscii` into a single test and converted `TestString_Empty`, `TestString_Ascii`, and `TestString_AsNumber` into a more table-driven style. I kept `TestString_AsDateTime` as it was in my last revision though, just splitting it out into subtests.

gavv:
Thanks for update!

I've pushed a small follow-up commit to this branch: 7caf92c38aa053b4ef2631ec4afc75b1e31d67ee

What I changed:

* Modify TestString_IsEmpty and TestString_IsAscii to create a new value for each check, instead of reusing value. I first wanted to ask you to return those clearFailed calls, because otherwise it's easy to forget one when reordering checks. But then I realized that it's much better to avoid reusing values at all.

* Create reporter inside subtest, instead of outside. Reporter is attached to testing.T, and in subtest we have a new instance of it.

* Minor formatting cleanup.

Issue #297, opened 2023-02-18
Add canonization tests for Array.Transform and Object.Transform


gavv:
Follow-up issue for https://github.com/gavv/httpexpect/issues/288

Array.Transform applies a function for each array element and returns a new, transformed array, with the results of the function. As many other functions in this library, Transform performs canonization of the values. For example, if the function returned `myInt(123)`, corresponding element will be converted to its canonical form, i.e. `float64(123.0)`.

The same is true for Object.Transform.

We should add "canonization" subtests to TestArray_Transform and TestObject_Transform, which check that: if transformation function returns value in non-canonical form, the corresponding element in array or object returned from Transform will have canonical form.

Darrellbor:
I would like to pick up on this task @gavv

PR #299, opened 2023-02-20
Array/Object transform canonization

Darrellbor:
PR for issue #297 
Implement canonization for 
- [x] Array Transform Canonization
- [x] Object Transform Canonization

### DIFF
--- a/array_test.go
+++ b/array_test.go
@@ -1600,9 +1600,8 @@ func TestArray_Transform(t *testing.T) {
 			return nil
 		})
 
-		assert.Equal(t, []interface{}{float64(2), float64(4), float64(6)}, newArray.Raw())
+		assert.Equal(t, []interface{}{2.0, 4.0, 6.0}, newArray.Raw())
 		newArray.chain.assertNotFailed(t)
-		newArray.chain.clearFailed()
 	})
 }
 
--- a/array_test.go
+++ b/array_test.go
@@ -1601,15 +1601,15 @@ func TestArray_Transform(t *testing.T) {
 			return nil
 		})
 
-		newArray.IsEqual([]interface{}{myFloat(2), myFloat(4), myFloat(6)})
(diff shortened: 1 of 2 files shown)


Darrellbor:
@gavv completed and ready for review

gavv:
Thanks for PR!

There is one minor issue. In this test, we want to check how Transform performs canonization, and we do NOT want to check how IsEqual performs canonization.

So instead of calling IsEqual with various types, we should call Transform with various types and then check that newArray.Raw() always returns canonized value. We don't need IsEqual at all in this test.

Darrellbor:
@gavv I have adjusted the array transform to `assert` equals rather than direct equals; although i can no longer use wrapped types like 
```
type (
 myArray []interface{}
 myFloat float64
)
```
As it complains that `float64` does not equal `myFloat` for example.

Do i also do the same for object transform?

gavv:
Wrapped types should be returned from Transform, not passed to assert.Equal. myArray is not relevant, but myFloat can be used.

> Do i also do the same for object transform?

Yep.

array_test.go
```diff
@@ -1582,6 +1582,35 @@ func TestArray_Transform(t *testing.T) {
...
+		newArray.chain.assertNotFailed(t)
+		newArray.chain.clearFailed()
+
+		assert.NotEqual(t, []interface{}{"2.0", float64(4), float64(6)}, newArray.Raw())
+		newArray.chain.assertNotFailed(t)
+		newArray.chain.clearFailed()
+
+		assert.Equal(t, []interface{}{2.0, float64(4), 6.0}, newArray.Raw())
+		newArray.chain.assertNotFailed(t)
+		newArray.chain.clearFailed()
```
gavv, line 1612: These three checks can be reduced to only one:

```go
		assert.Equal(t, []interface{}{2.0, 4.0, 6.0}, newArray.Raw())
		newArray.chain.assertNotFailed(t)
```

Darrellbor: done

gavv:
Thanks!

Issue #260, opened 2023-02-01
Add Environment.Clear


gavv:
Add new method Environment.Clear which removes all values.

New method should have documentation comment with code example and be covered by tests.

e-berman:
I'll give this one a go @gavv.

Sorry for the stupid question, but you want to keep the keys and only clear the values, correct?

gavv:
Cool, thank you. Nooe, sorry for misunderstanding, I meant just to remove everything, keys and values. To allow the user to reset state.

PR #300, opened 2023-02-20
Add Environment.Clear method

e-berman:
PR addresses the following issue #260 

Any change recommendations or best practice tips are greatly appreciated.


environment.go
```diff
@@ -82,6 +82,27 @@ func (e *Environment) Delete(key string) {
...
+//
+// Example:
+//
+//	env := NewEnvironment(t)
+//	env.Put("key1", 123)
+//	env.Put("key2", 456)
+//	env.Clear()
+func (e *Environment) Clear() {
+	opChain := e.chain.enter("Clear Data")
+	defer opChain.leave()
```
gavv, line 95: The string passed to enter() should be method name, in this case it should be:

```
enter("Clear()")
```

environment.go
```diff
@@ -82,6 +82,27 @@ func (e *Environment) Delete(key string) {
...
+
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	for key := range e.data {
+		delete(e.data, key)
+	}
```
gavv, line 102: IIRC, this way of clearing map will keep allocated buckets inside the map.

Suppose the user have a few large groups of tests. Each group clears environment first and then tests in this group extensively use environment.

In this case, I think it's more convenient to allow GC to fully collect environment memory from previous tests, to avoid unnecessary heap growth.

Long story short, I suggest to replace this with:

```go
e.data = make(map[string]interface{})
```

:)

environment_test.go
```diff
@@ -90,6 +90,26 @@ func TestEnvironment_Delete(t *testing.T) {
...
+		assert.NotNil(t, env.Get(key))
+		assert.Equal(t, i, env.Get(key).(int))
+		env.chain.assertNotFailed(t)
+	}
+
+	env.Clear()
+	env.chain.assertNotFailed(t)
+
+	assert.Zero(t, len(env.data))
+	env.chain.assertNotFailed(t)
```
gavv, line 110: I'd also add checks that Has and Get will fail for the previously added keys.

gavv (commented):
Thanks for PR! A few comments.

environment.go
e-berman: This makes perfect sense. Thank you for clearing this up, and I will make sure to make the change.

gavv:
LGTM!

PR #301, opened 2023-02-21
Add Environment.List and Find

### LINKED ISSUE #259: Add Environment.List and Find
Add two methods to Environment:

* List - returns a sorted slice of keys
* Find - accepts a glob pattern and returns a sorted slice of keys that match pattern

Methods should have documentation comment with code example and be covered by tests.

nhAnik:
Closes #259


environment.go
```diff
@@ -424,6 +427,75 @@ func (e *Environment) GetTime(key string) time.Time {
...
+	keys := []string{}
+	glb, err := glob.Compile(pattern)
+	if err != nil {
+		opChain.fail(AssertionFailure{
+			Type:   AssertType,
+			Actual: &AssertionValue{pattern},
+			Errors: []error{
+				errors.New("expected: valid glob pattern"),
+			},
+		})
```
gavv, line 483: Let's use AssertUsage instead of AssertType.

AssertUsage is intended for cases when the library is used incorrectly.

AssertType is for cases when the user asks to check value type, and value has wrong type.

environment_test.go
```diff
@@ -489,3 +489,53 @@ func TestEnvironment_Time(t *testing.T) {
...
+	env.Put("abc", 4)
+	assert.Equal(t, []string{"abc", "k1", "k2", "k3"}, env.List())
+
+	env.Delete("k2")
+	env.Delete("k3")
+	env.Put("pqr", 5)
+	assert.Equal(t, []string{"abc", "k1", "pqr"}, env.List())
+}
+
+func TestEnvironment_Match(t *testing.T) {
```
gavv, line 512: Could you please divide this test into two: one for invalid glob pattern, another for normal checks? (We use same approach in most tests).

environment.go
```diff
@@ -424,6 +427,75 @@ func (e *Environment) GetTime(key string) time.Time {
...
+// keys that match pattern.
+//
+// Example:
+//
+//	env := NewEnvironment(t)
+//	env.Put("key1", "v1")
+//	env.Put("key2", "v2")
+//	env.Put("abc", "v3")
+//	keys := env.Match("k*")
+func (e *Environment) Match(pattern string) []string {
```
gavv, line 467: As for me, `Match` is a bit confusing here:

* env.Match call looks like we want to match contents of environment; but we instead want to find keys which match pattern
* in httpexpect, String.Match* method family work with regexps, and this one works with glob

Maybe rename it to Find or Glob? Unlike env.Match, it's hard to imagine how env.Glob can match anything besides keys, so Glob seems better.

environment_test.go
nhAnik: Split the test into two subtests.

environment.go
nhAnik: Changed to **Glob**

gavv:
Thanks!

PR #303, opened 2023-02-22
Add test TestE2EReport_LineWidth

### LINKED ISSUE #273: Add test TestE2EReport_LineWidth
Add a new test that checks work of DefaultFormatter.LineWidth option.

The test should produce a failure with long lines (i.e. long assertion path and long error) and see how it's formatted with several variants of LineWidth:

* zero value (means use default limit): all lines should be shorter than default limit
* small positive value: all lines should be shorter than configured limit
* very large positive value: there are some long lines
* negative value (means no limit): there are some long lines

This test should not be very strict to the report format. Ideally it should check only width of lines in produced report and occurrence of assertion path and error message in that report.

Related: #165

dhanusaputra:
close #273


gavv:
Hi, thanks for PR!

Two comments:

1. Instead of using custom template, it would be much more useful to check default template, to find potential bugs in it. With current test, we're actually only testing wrap and join functions, but not how they're used in default template.

2. I think we should not trim spaces when checking line width, because spaces should be covered by line length limit.

e2e_report_test.go
```diff
@@ -75,3 +77,116 @@ func TestE2EReport_Aliases(t *testing.T) {
 
 	assert.Contains(t, rep.reported, "foo.Object().ContainsKey()")
 }
+
+const (
+	intSize = 32 << (^uint(0) >> 63) // 32 or 64
+	maxInt  = 1<<(intSize-1) - 1
+)
```
gavv, line 84: Can we just use, say, 1000000? I guess there is no real need for using max int here?

e2e_report_test.go
```diff
@@ -75,3 +77,116 @@ func TestE2EReport_Aliases(t *testing.T) {
...
+				JSON().
+				Object().
+				NotContainsValue(123)
+
+			var skip bool
+			for _, s := range strings.Split(strings.Trim(rep.reported, "\n"), "\n") {
+				// only assert first errors block and assertPath block in defaultFailureTemplate
+				if strings.HasPrefix(s, "assertion:") {
+					skip = false
+				}
```
gavv, line 160: Do we really need this parsing?

Can we instead make very simple assertions, like:

* in this test, there should be no lines longer than N
* in this test, there should be at least one line longer than N

This way we still check that line limit, well, limits line width, and at the same time we don't bother how and where exactly it is applied.

e2e_report_test.go
```diff
@@ -75,3 +77,116 @@ func TestE2EReport_Aliases(t *testing.T) {
...
+
+				lenLastWord := len(ss[len(ss)-1])
+
+				// additional indent in template
+				var lenIndent int
+				if strings.HasPrefix(s, defaultIndent) {
+					lenIndent = len(defaultIndent)
+				}
+
+				lenBeforeWrapped := len(s) - lenLastWord - lenIndent - 1
```
gavv, line 184: Could you clarify why do we need these calculations?

gavv (commented):
Thanks for update. Some more comments.

e2e_report_test.go
dhanusaputra: good idea, thank you

e2e_report_test.go
dhanusaputra: basically want to get len of chars before it get wrapped so we can compare it directly with the LineWidth config in formatter

for example we have this line `    JSON().Object().`
want to get len of `  JSON()`
calculation become whole length minus: 
- len `Object()` - indent outside wrap or join func https://github.com/gavv/httpexpect/blob/79ad562e6ea5c7d34a4e1748bed7e93cb3b388b1/formatter.go#L698 
- len`.`

but yeah looks complicated, guess will just ignore this

gavv:
Thanks for update.

The new version is almost perfect, but there is one downside: we should hard-code maximum line width for failure, which can easily break test when formatting slightly changes.

Also it was suspicions that for `LineWidth: 15` we set `wantMaxLineWidth: 26`.

I checked, and your test really found a bug in DefaultFormatter: maximum line limit can be exceeded.

I fixed the bug and refactored test a bit: now instead of expected exact size for longest line, we specify expected range for longest line (i.e. we expect that it's above X and below Y). We ensure that without limit, the longest line is above 100 chars, and with limit, longest line is below limit.

I've pushed these two commits to PR's branch and now going to mere it.

Thanks!

gavv:
BTW welcome to httpexpect discord https://discord.gg/5SCPCuCWA9 :)

dhanusaputra:
nice! thanks for your clear explanation

PR #305, opened 2023-02-23
Convert unit tests into subtest and to be table-driven

gpjservais:
This is the final PR addressing refactoring tests from #284. This addresses the following tests:
- [x] TestMatch_IsEmpty
- [x] TestMatch_ValuesEmpty
- [x] TestNumber_EqualNaN
- [x] TestArray_IsEqual/types (split into subtests)
- [x] TestWebsocketMessage_CodeAndType
- [x] TestWebsocketMessage_NoContent

I converted all except `TestArray_IsEqual` to table driven style. I only split `TestArray_IsEqual/types` into subtests to address the multiple variable issue.

Also, as noted in the issue, I ended up merging `TestNumber_EqualNaN` into `TestNumber_IsEqual` (and also `TestNumber_InDelta` since half of the test related to InDelta/NotInDelta tests for NaN values). I ended up converting those tests to table-driven format as well.

This is ready for review.

PR #306, opened 2023-02-24
Implement Number.InDeltaRelative

### LINKED ISSUE #156: Implement Number.InDeltaRelative
Implement new `Number` assertion `InDeltaRelative` (and counterpart `NotInDeltaRelative`).

It should work similar to `InDelta`, however `delta` parameter is treated as relative to value, instead of absolute error.

For example, if `delta` is `0.01`, it means that values should not differ by more than `1%`.

This is similar to `InDelta` vs `InEpsilon` in testify.

Assertions should have documentation comment with example and be covered by unit tests.

When failure is reported, we should report delta as `%`. We can achieve this as follows:

```go
type relativeDelta float64

// formatting engine will use it to print delta
func (d relativeDelta) String() {
    return fmt.Sprintf(%f%%", d)
}

...

AssertionFailure{
    ...
    Delta: &AssertionValue{relativeDelta(delta)},
    ...
}
```

selborsolrac:
This PR resolves #156.

Please let me know if anything needs to be changed!


number.go
```diff
@@ -273,6 +273,126 @@ func (n *Number) NotInDelta(value, delta float64) *Number {
 	return n
 }
 
+// InDeltaRelative succeeds if two numerals are within a relative delta of each other.
+//
+// The relative delta is expressed as a percentage (i.e. 0.01 is 1%).
```
gavv, line 278: nit:

* I wouldn't say it's percentage - user should pass `0.01`, not `1`
* In comment, let's also provide the actual formula we use: `Abs(number-value) / Abs(number) < delta`
* In comment, let's mention that number can't be zero
* In comment, let's mention that number, value, and delta can't be NaN

number.go
```diff
@@ -273,6 +273,126 @@ func (n *Number) NotInDelta(value, delta float64) *Number {
...
+
+	if math.IsNaN(n.value) || math.IsNaN(value) || math.IsNaN(delta) {
+		opChain.fail(AssertionFailure{
+			Type:     AssertEqual,
+			Actual:   &AssertionValue{n.value},
+			Expected: &AssertionValue{value},
+			Delta:    &AssertionValue{relativeDelta(delta)},
+			Errors: []error{
+				errors.New("expected: numbers are comparable"),
+			},
```
gavv, line 300: I suggest to refine error a bit, to something like:

```go
Errors: []error{
	errors.New("expected: can compare values with relative delta"),
	errors.New("actual value is NaN"), // included if n.value is NaN
	errors.New("expected value is NaN"), // included if value is NaN
	errors.New("delta is NaN"), // included if delta is NaN
},
```

(Feel free to reword if my English is poor).

number.go
```diff
@@ -273,6 +273,126 @@ func (n *Number) NotInDelta(value, delta float64) *Number {
...
+
+	if n.value == 0 {
+		opChain.fail(AssertionFailure{
+			Type:     AssertEqual,
+			Actual:   &AssertionValue{n.value},
+			Expected: &AssertionValue{value},
+			Delta:    &AssertionValue{relativeDelta(delta)},
+			Errors: []error{
+				errors.New("expected: number value is not 0"),
+			},
```
gavv, line 313: I suggest to refine this error too:

```go
Errors: []error{
	errors.New("expected: can compare values with relative delta"),
	errors.New("actual value is zero, which is not allowed"),
},
```

number.go
```diff
@@ -1091,3 +1211,9 @@ func (b intBoundary) String() string {
 	}
 	return fmt.Sprintf("%s", b.val)
 }
+
+type relativeDelta float64
+
+func (rd relativeDelta) String() string {
+	return fmt.Sprintf("%f%%", rd*100)
```
gavv, line 1218: nit: let's print both original value and percentage, like this:

```
0.01 (1%)
```

number_test.go
```diff
@@ -256,6 +256,44 @@ func TestNumber_InDelta(t *testing.T) {
 	value.chain.clearFailed()
 }
 
+func TestNumber_InDeltaRelative(t *testing.T) {
```
gavv, line 259: * Could you also add a subtests that cover corner cases when n.value, value, and delta are nan, inf, and zero?
* Also please add calls to new methods to TestNumber_FailedChain.

gavv (commented):
Thanks for PR! See a few comment below.

gavv:
After merging https://github.com/gavv/httpexpect/pull/305, TestNumber_InDelta is now table-driven. Could you please make TestNumber_InDeltaRelative table-driven too?

https://github.com/gavv/httpexpect/blob/b47b582179ffb01661c6d169accc342633e96957/number_test.go#L212

selborsolrac:
Hello,

I've implemented the requested changes as I understood them. I'm not sure I properly implemented the corner cases though. For example, I set the test to fail if any provided parameter is `0`, `Inf`, or `NaN`, but I'm not sure that was your intention. For example, a corner case where the `value` is `0` is fine if the `delta` is `1` and a corner case where the `delta` is `0` is fine if the `value` is equal to `n.value`, so we may not want to necessarily error out in those cases. Let me know what I should further tweak!

number.go
```diff
@@ -1091,3 +1341,9 @@ func (b intBoundary) String() string {
 	}
 	return fmt.Sprintf("%s", b.val)
 }
+
+type relativeDelta float64
+
+func (rd relativeDelta) String() string {
+	return fmt.Sprintf("%.2f (%.f%%)", rd, rd*100)
```
gavv, line 1348: nit: `.2f` may be too rough, let's just use `%v`

gavv:
So let's break down the corner cases

Given it another thought, mathematically, I think here's how it should work:

| actual / expected | 123       | 0         | -inf      | +inf      | nan       |
|-------------------|-----------|-----------|-----------|-----------|-----------|
| **123**           | **true**  | false     | false     | false     | _invalid_ |
| **0**             | false     | **true**  | false     | false     | _invalid_ |
| **-inf**          | false     | false     | **true**  | false     | _invalid_ |
| **+inf**          | false     | false     | false     | **true**  | _invalid_ |
| **nan**           | _invalid_ | _invalid_ | _invalid_ | _invalid_ | _invalid_ |

Here:

* true means that InDeltaRelative succeeds and NotInDeltaRelative fails
* false means that InDeltaRelative fails and NotInDeltaRelative succeeds
* invalid means that both of them fail

What do you think about changing formula from:

```go
math.Abs(n.value-value) / math.Abs(n.value) > delta
```

to:

```go
math.Abs(n.value-value) > math.Abs(n.value) * delta
```

This way zero will be handled automatically. If `n.value` is zero, InDeltaRelative will succeed when and only when `value` is zero (and delta is anything finite).

We'll still need to handle manually inf and nan. If either of values is nan, we should fail. If either of values is inf, we should succeed only if they're both infs of the same sign.

BTW we should also check that delta is non-negative, and fail with AssertUsage otherwise.

gavv:
Hm, on the other hand, the formula I suggested will make precision worse.

Then a better option is to use the old formula, but pretend that we're using the new one by handling zero specially.

selborsolrac:
Alright, I've made some changes to try and accommodate all of the edge cases. I think the logic and tests are correct, but please review and confirm.

Additional cases added:
- Output assert usage error if delta is negative
- Special formula if target is 0
- Add test for reference = 0
- Error out if target and/or reference are Inf

A note on the Inf values:
It appears that (`+-Inf`) / (`+-Inf`) = `NaN`, so we error out if the target and/or reference are `+-Inf`. If I'm missing something and there's a way to make this work, please let me know!

A note on the  `relativeDelta.String()` function:
It looks like using `%v` formats the decimal so that it outputs the minimum number of decimal places (e.g. 1.25000 is output as 1.25). However, we can't use `%v` in the Stringer function because it produces error `fmt.Sprintf format %v with arg rd causes recursive` so I opted to use `strconv` to replicate the same behavior. Let me know if there's a better way to handle this!

selborsolrac:
I think I can make the +-Inf cases work if I use the same rearranged formula for target = 0. I'll try implementing that.

gavv:
It seems we can combine two checks:

```go
	if n.value == 0 {
		if math.Abs(n.value-value) > math.Abs(n.value)*delta {
```

```go
	relativeDiff := math.Abs(n.value-value) / math.Abs(n.value)

	if relativeDiff > delta {
```

into one:

```go
if (n.value == 0 && value != 0) || math.Abs(n.value-value) / math.Abs(n.value) > delta {
    ...
}
```

gavv:
Or, generalizing it to inf too, as you mentioned above:

```go
if ((n.value == 0 || math.IsInf(n.value, 0)) && value != n.value) || 
  math.Abs(n.value-value) / math.Abs(n.value) > delta {
    ...
}
```

gavv:
>  It looks like using %v formats the decimal so that it outputs the minimum number of decimal places (e.g. 1.25000 is output as 1.25). However, we can't use %v in the Stringer function because it produces error fmt.Sprintf format %v with arg rd causes recursive so I opted to use strconv to replicate the same behavior. Let me know if there's a better way to handle this!

I think this can be avoided by casting value to float64 before passing to Sprintf.

number_test.go
```diff
@@ -334,6 +336,148 @@ func TestNumber_InDelta(t *testing.T) {
 	}
 }
 
+func TestNumber_InDeltaRelative(t *testing.T) {
+	cases := map[string]struct {
+		number           float64
+		reference        float64
```
gavv, line 342: nit: could you please use "value" instead of "reference"? (or other name). To avoid confusion with AssertionFailure.Reference, which may be part of equality assertion (but not in this case).

number_test.go
```diff
@@ -334,6 +336,148 @@ func TestNumber_InDelta(t *testing.T) {
 	}
 }
 
+func TestNumber_InDeltaRelative(t *testing.T) {
+	cases := map[string]struct {
```
gavv, line 340: nit: could you use slices of structs instead of map of structs?

See https://github.com/gavv/httpexpect/issues/324

selborsolrac:
Hello! I've made the requested changes.

I spent some time trying to wrap my head around infinite numbers and ensuring that the logic accounted for them, but I think it complicates things significantly. Because infinite numbers are not real, we cannot use them with any formula to get a defined answer.

For example:

+-Inf - +-Inf = undefined / NaN
+-Inf * 0 = undefined / NaN

I think it might be best to only accept real numbers and error out if infinite numbers are input for number, value and/or delta, but I leave that to your discretion. Let me know what you think!

gavv:
Hi, thanks for update!

Your last comment makes sense.

I think that we have to support two cases with Inf:

* `Number(+Inf).InDeltaRelative(+Inf, 0)` should succeed
* `Number(-Inf).InDeltaRelative(-Inf, 0)` should succeed

because it's expected that `InDeltaRelative(x, 0)` is equivalent to `IsEqual(x)`, and it will work in these cases.

Since we allow should `InDeltaRelative(Inf, 0)`, it also makes sense to allow any finite delta here, because the second is logically expected from the first I think.

Now, if `value == +Inf || value == -Inf` and `value != n.value`, then it seems we can safely declare that these numbers are not comparable (we should fail anyway, and failing as with this error seems fitting). We can add this condition to the first check (`if anyNumIsNaN`) and add corresponding error messages (`actual value is -Inf`).

We can also add `delta == Inf || delta == -Inf` to that condition as well.

This way, we will filter out all weird cases with Inf, right?

number.go
```diff
@@ -273,6 +273,186 @@ func (n *Number) NotInDelta(value, delta float64) *Number {
...
+	}
+
+	anyNumIsNaN := math.IsNaN(n.value) || math.IsNaN(value) || math.IsNaN(delta)
+
+	if anyNumIsNaN {
+		var assertionErrors []error
+		assertionErrors = append(
+			assertionErrors,
+			errors.New("expected: can compare values with relative delta"),
+		)
```
gavv, line 306: I think we can extract this errors building into a function, given that it's become bigger.

number.go
```diff
@@ -273,6 +273,186 @@ func (n *Number) NotInDelta(value, delta float64) *Number {
...
+			Type: AssertUsage,
+			Errors: []error{
+				errors.New("unexpected negative delta argument"),
+			},
+		})
+		return n
+	}
+
+	if ((n.value == 0 || math.IsInf(n.value, 0)) && value != n.value) ||
+		math.Abs(n.value-value)/math.Abs(n.value) > delta {
```
gavv, line 350: I suggest to extract this into a function as well, with multiple if's and return's, for readability.

gavv:
Thanks for update. Please rebase on fresh master to fix CI.

selborsolrac:
Hello, I've made the requested changes related to Inf numbers, let me know what you think.

The CI failed but it looks like maybe it's a transient issue that can be resolved with a retry.

gavv:
Thanks a lot for update!

Here is a small follow-up commit: 2b1131c5dcb3b2d14a5d0227562bc079b0b0af56

I've rearranged checks a bit and changed one bit of behavior: NotInDeltaRelative now succeeds if numbers are opposite infs. Also added a few more tests.

Issue #248, opened 2023-01-28
Implement PanicReporter


gavv:
See #187 for details. PanicReporter will be similar to FatalReporter, but instead of calling t.Fatalf, it just calls panic.

This reporter may be useful in multithreaded tests because t.Fatal may be called only from test's goroutine. This should be mentioned in the comment.

PR #315, opened 2023-02-28
Implemented: PanicReporter (#248)

Dev-29:
Here is my implementation for PanicReporter (#248)


gavv:
Hi, thanks for PR. 

The new reporter does not actually use panic and is just a copy-paste of FatalReporter?

Also, could you please cover the new reporter with tests, similar to other reporters?

gavv:
Also please fix linter error.

PR #316, opened 2023-02-28
Add tests for custom templates in DefaultFormatter

### LINKED ISSUE #263: Add tests for custom templates for DefaultFormatter
DefaultFormatter allows setting custom templates for success and failure report, and custom template function map.

We should add a we tests covering this:

* custom success template
* custom failure template
* invalid template (first panic in applyTemplate)
* template that tries to access nonexistent fields from FormatData (second panic in applyTemplate)

Related: https://github.com/gavv/httpexpect/issues/165

nhAnik:
Closes #263

### DIFF
--- a/formatter_test.go
+++ b/formatter_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"text/template"
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


gavv:
Thanks for PR!

The test looks good.

Given it a second thought, I realized that it's better to move this test to e2e_report_test.go (end-to-end tests for formatting), which I did in this commit: ff14676cb254ecc1957513f6f3f13d1e06f71288

(Also I applied a few cosmetic refactorings)

Thanks!

PR #318, opened 2023-03-01
Support colored output

### LINKED ISSUE #161: Support colored output
Modify DefaultFormatter to support colorized output.

It should provide three color modes, controlled by an option:

* auto (used by default) - enable colors if testing.Verbose() is true and stdout is a tty / console
* always - always use colors
* never - never use colors

Color support should work for both unix and windows.

Colors should be injected into failure template using template functions, but should not be present in FormatData. This way users that use DefaultFormatter with custom templates won't be affected.

dhanusaputra:
close #161 

tried using https://github.com/TylerBrock/colorjson but not really sure since all jsonObj inputs are string (all color become green as default)
use configured color in template instead

need to change e2e_report_test.go because the color attribute add more length to string


gavv:
Hi, thanks for PR!

Regarding TestE2EReport_LineWidth. I think the correct way to fix it is to disable colors in that test.

We could do it manually in test, but I have a better idea - it would be convenient to disable colors by default if we're not sure that reporter writes to stdout (as in that test).

Current condition is:

```
testing.Verbose() && isatty(stdout)
```

the new condition would be:

```
<reporter uses testing.T> && testing.Verbose() && isatty(stdout)
```

To achieve this, we can do the following:

* add new boolean field to AssertionContext, say, TestingTB

* in newChainWithDefaults, set AssertionContext.TestingTB to true if reporter is AssertReporter/RequireReporter/FatalReporter/testing.TB

* in newChainWithConfig, set AssertionContext.TestingTB to true if AssertionHandler is DefaultAssertionHandler and DefaultAssertionHandler.Reporter is AssertReporter/RequireReporter/FatalReporter/testing.TB

* in DefaultFormatter, for ColorModeAuto, enable colors only if AssertionContext.TestingTB is true

formatter.go
```diff
@@ -215,6 +237,19 @@ func (f *DefaultFormatter) buildFormatData(
 		}
 	}
 
+	var enableColor bool
+	switch f.ColorMode {
+	case ColorModeAuto:
+		fd := os.Stdout.Fd()
+		enableColor = testing.Verbose() &&
+			(isatty.IsTerminal(fd) || isatty.IsCygwinTerminal(fd))
```
gavv, line 245: Ideally, this part of the check:

```
(isatty.IsTerminal(fd) || isatty.IsCygwinTerminal(fd))
```

should be done only once. We can achieve it by using sync.Once (it's better to use sync.Once instead of init() because this way we postpone the check until the first call to formatter and give the user a chance to manipulate with os.Stdout in test setup code).

Also, ColorModeAuto should disable colors by two additional conditions:

* if TERM env variable is set and starts with "dumb"
* or if NO_COLOR env variable is set and is non-empty

formatter.go
```diff
@@ -677,30 +712,49 @@ var defaultTemplateFuncs = template.FuncMap{
...
+		var ca color.Attribute
+		switch c {
+		case "red":
+			ca = color.FgRed
+		case "yellow":
+			ca = color.FgYellow
+		case "green":
+			ca = color.FgGreen
+		case "cyan":
+			ca = color.FgCyan
```
gavv, line 728: Let's list all defined colors, and in the same order, for convenience:

```
	FgBlack
	FgRed
	FgGreen
	FgYellow
	FgBlue
	FgMagenta
	FgCyan
	FgWhite
```

formatter.go
```diff
@@ -677,30 +712,49 @@ var defaultTemplateFuncs = template.FuncMap{
...
+		default:
+			ca = color.FgCyan
```
gavv, line 730: I guess white would be a better default

formatter.go
```diff
@@ -143,6 +151,8 @@ type FormatData struct {
 	Diff     string
 
 	LineWidth int
+
+	EnableColor bool
```
gavv, line 155: nit: let's rename it to EnableColor**s**

formatter.go
```diff
@@ -215,6 +237,19 @@ func (f *DefaultFormatter) buildFormatData(
 		}
 	}
 
+	var enableColor bool
+	switch f.ColorMode {
+	case ColorModeAuto:
+		fd := os.Stdout.Fd()
+		enableColor = testing.Verbose() &&
```
gavv, line 244: Let's move this logic to `fillDescription`. We can rename it to `fillGeneral` or something like this, if you think it'd be a better name.

formatter.go
```diff
@@ -677,30 +712,49 @@ var defaultTemplateFuncs = template.FuncMap{
...
+			ca = color.FgCyan
+		default:
+			ca = color.FgCyan
+		}
+		return color.New(ca).Sprint(s)
+	},
 }
 
-var defaultSuccessTemplate = `[OK] {{ join .AssertPath .LineWidth }}`
+var defaultSuccessTemplate = `[OK] {{ join .AssertPath .LineWidth | color .EnableColor "green" }}` //nolint
```
gavv, line 736: I think for success template we remove coloring. Otherwise the screen will be full of green lines :)

gavv:
Hi, I see some commits, please ping me or press "request review" when it's ready.

assertion.go
```diff
@@ -134,6 +134,10 @@ type AssertionContext struct {
 	// Environment shared between tests
 	// Comes from Expect instance
 	Environment *Environment
+
+	// TestingTB is true when reporter is
+	// AssertReporter/RequireReporter/FatalReporter/testing.TB
+	TestingTB bool
```
gavv, line 140: Let's make this comment more generic:

```
	// TestingTB is true when reporter is known to output to testing.TB
```

chain.go
```diff
@@ -126,6 +126,14 @@ func newChainWithConfig(name string, config Config) *chain {
 		c.context.Environment = newEnvironment(c)
 	}
 
+	ah, ok := c.handler.(*DefaultAssertionHandler)
+	if ok {
+		switch ah.Reporter.(type) {
+		case *AssertReporter, *RequireReporter, *FatalReporter, testing.TB:
+			c.context.TestingTB = true
+		}
+	}
```
gavv, line 135: Let's extract this snippet into a function and reuse it.

chain.go
```diff
@@ -126,6 +126,14 @@ func newChainWithConfig(name string, config Config) *chain {
 		c.context.Environment = newEnvironment(c)
```
gavv, line 126: Let's add a test to `chain_test.go` that sets various reporters and checks how chain ctor fills context.TestingTB

formatter.go
```diff
@@ -5,16 +5,26 @@ import (
...
+	"github.com/mattn/go-isatty"
 	"github.com/mitchellh/go-wordwrap"
 	"github.com/sanity-io/litter"
 	"github.com/yudai/gojsondiff"
 	"github.com/yudai/gojsondiff/formatter"
 )
 
+var (
+	once      sync.Once
+	withColor bool
```
gavv, line 25: These vars are global for the whole project, so let's give them some specific names, ideally related to each other, e.g.:

* colorsSupportedGuard
* colorsSupported

formatter.go
```diff
@@ -239,6 +267,24 @@ func (f *DefaultFormatter) fillDescription(
...
 	}
+
+	once.Do(func() {
+		fd := os.Stdout.Fd()
+		withColor = (isatty.IsTerminal(fd) || isatty.IsCygwinTerminal(fd)) &&
+			len(os.Getenv("NO_COLOR")) == 0 &&
+			!strings.HasPrefix(os.Getenv("TERM"), "dumb")
+	})
+
+	var enableColors bool
```
gavv, line 278: No need for this var, we can directly set data.EnableColors in switch below.

gavv (commented):
Thanks for update. A couple more comments.

gavv:
Thank you so much!

I've pushed a commit to this branch with a few minor changes: 9808df786ff32b63b4116b82f74ad445c249198b

* expand comments a bit
* use AssertionHandler instead of interface{} in isTestingTB
* merge Test_isTestingTB into TestChain_TestingTB (it's much more interesting to test how chain handles it)
* converted colors switch to map and added bright colors

I've removed colorization from diff (we'll better colorize diff syntactically later) and changed cyan to bright magenta - this color is readable on both dark and light background:

image

image

dhanusaputra:
thank you! it's better

Issue #308, opened 2023-02-24
Refactor TestRequest_Order to be table-driven test


gavv:

e-berman:
Hello! You can assign this one to me as well.

gavv:
Thanks!

gavv:
Landed

PR #319, opened 2023-03-02
refector order with table test

Priyank-j:
PR addresses issue [https://github.com/gavv/httpexpect/issues/308]

Any criticism/feedback is always appreciated.

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -2904,12 +2904,12 @@ func TestRequest_Order(t *testing.T) {
 				return req.WithWebsocketUpgrade()
 			},
 		},
-		//{
-		//	name: "WithWebsocketDialer after Expect",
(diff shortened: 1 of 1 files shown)


gavv:
@Priyank-j hi, actually this issue was already assigned to @e-berman 

@e-berman have you started working on it already?

e-berman:
Yes, but @Priyank-j can take the issue if they want!

gavv:
@e-berman since you were assigned earlier than this PR was created, let's wait for your implementation then and close this one.

@Priyank-j sorry for inconvenience. To avoid such situations, you can check that issue is not assigned already and leave a comment on the issue, so that you can be assigned. 

Thanks for PR anyway!

PR #321, opened 2023-03-02
WIP: Add withLimit functionality 

### LINKED ISSUE #271: Add Request.WithLimit
Add new method Request.WithLimit that set's a limit for maximum number of bytes that can be retrieved from response body.

It can be helpful to interrupt early on abnormal response (e.g. mistakenly trying to read infinite chunked response into a memory), thus avoiding insane memory consumption by tests, swapping, etc.

Request.WithLimit should store the provided limit into a private field of Request. Later, in Request.retryRequest we should check if limit is set, and if so, pass this limit to bodyWrapper when we're wrapping response body.

Then, we should teach bodyWrapper to handle the limit. When the limit is set and bodyWrapper detected that it read more bytes than desired, it should abort read and return error that explains that limit is exceed and what is the limit.

Ideally error message should contain limit both in bytes and in human readable form, like "5 MB".

New method should have documentation comment with code example and two tests:

* unit test for request (request_test.go)
* end-to-end (blackbox) test (we can create new file e2e_limit_test.go)

Rohith-Raju:
Initial commit that closes #271 

Changes Made are 
- [x] Added a new func `newBodyWrapperLimit`
- [x] Added a private `limit` field and introduced the `WithLimit()` func 
- [x] Test cases for withLimit func  

Pending changes
- [ ] Formatting error messages and making them human-readable.
- [ ] More unit tests
- [ ] End-to-End tests


Rohith-Raju:
@gavv Your thoughts on the implementation? I need some help with error messages. I'll reach out on chat 😄

PR #323, opened 2023-03-02
Refactor TestRequest_Order to table driven

e-berman:
PR addresses issue #308 

Let me know if anything needs to be changed, or any feedback!

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -2849,276 +2849,432 @@ func TestRequest_Usage(t *testing.T) {
 }
 
 func TestRequest_Order(t *testing.T) {
-	client := &mockClient{}
-
-	config := Config{
-		Client:   client,
-		Reporter: newMockReporter(t),
+	tests := []struct {
+		name         string
+		prepFunc     func(req *Request) *Request
+		expectFunc   func(req *Request) *Response
+		reqFunc      func(req *Request)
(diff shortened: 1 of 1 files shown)


gavv:
Hi, thanks for PR.

These checks:

```go
			if tt.expectSame {
				assert.Same(t, req, tt.prepFunc(req))
			}

			if tt.expectNotNil {
				assert.NotNil(t, tt.expectFunc(req))
			}
```

(I mean  assert.Same and assert.NotNil) are actually not very important for this test. It's main responsibility is to check that the chain fails when methods are called in incorrect order.

And if we remove them, it becomes possible to significantly simplify the test:

* remove expectSame, expectNotNil

* replace prepFunc, expectFunc, reqFunc with two funcs: one called before Expect, one called after (we can name them beforeFunc and afterFunc)

* in the loop, we can use them as follows:

    ```go
    req := NewRequestC()

    if beforeFunc != nil {
        beforeFunc(req)
    }
    assertNotFailed()

    Expect()
    assertNotFailed()

    afterFunc(req)
    assertFailed()
    ```

* most tests won't need beforeFunc, so we can omit it from those tests for clarity

gavv:
Also, can we remove `Client:   &mockClient{}` and use default client instead?

Issue #324, opened 2023-03-03
Refactor table-driven Value tests


gavv:
In `value_test.go`, find all table-driven tests that use:

```go
cases := map[string]struct {
   ...
```

and apply two minor refactorings:

* replace map of structs with a slice of structs (with "name" field for test name); this way, subtests will be executed in the same order each invocation, which in convenient for debugging

* rename "instance" to "tc"; "tc" stands for "test case" and is used in most other table-driven tests

gavv:
Thanks!

gavv:
Landed

PR #325, opened 2023-03-03
Fix value table-driven test cases

gpjservais:
Addresses #324 to replace map of structs with a slice of structs and rename "instance" to "tc."

This is to fix updates made while refactoring tests as part of #284 to use table-driven style.

This is ready for review.


gavv:
Thanks!!

Issue #283, opened 2023-02-08
Refactor request unit tests


gavv:
Some tests in request_test.go have a multiple checks mixed in the same test. The problem with this approach is that it's error prone: they have many local vars, and it's easy to make a check on a wrong var.

We need to split some of the tests into subtests and isolate vars in them.

Here is the list of tests to be refactored:

* TestRequest_URLConcatenate
* TestRequest_URLOverwrite
* TestRequest_URLInterpolate
* TestRequest_URLQuery
* TestRequest_Host
* TestRequest_BodyChunkedProto
* TestRequest_ContentLength
* TestRequest_ContentTypeOverwrite
* TestRequest_ErrorConflictBody
* TestRequest_ErrorConflictType
* TestRequest_ErrorConflictMultipart

Pull requests that address only part of this issue are welcome too!

MaKYaro:
Hello @gavv! Can I help with this issue?

gavv:
Sure, thanks!

gavv:
Some of these tests were converted to subtests in https://github.com/gavv/httpexpect/issues/285

gavv:
I've updated issue according to recent changes.

gavv:
@MaKYaro do you still have plans on this?

MaKYaro:
No, sorry

gavv:
Thanks for quick reply.

Unassigning, so that someone could pick this up.

Darrellbor:
hey @gavv i'd love to pick this up

gavv:
You're welcome!

PR #333, opened 2023-03-09
feat: refactor requests into subtests

Darrellbor:
PR for #283 
implement subtests for 
 - [x] TestRequest_URLConcatenate
 - [x] TestRequest_URLOverwrite
 - [x] TestRequest_Host
 - [x] TestRequest_ContentLength
 - [x] TestRequest_ContentType
 - [ ] TestRequest_Conflicts


Darrellbor:
@gavv please could you look through `TestRequest_URLConcatenate` and advise if it's what you had in mind regarding sub tests

Darrellbor:
@gavv ready for review

request_test.go
```diff
@@ -522,31 +523,122 @@ func TestRequest_URLConcatenate(t *testing.T) {
...
+		config      Config
+		method      string
+		path        string
+		pathArg     string
+		pathArgType int
+		fullURL     string
+	}{
+		{
+			name:    "config1_empty_url_and_empty_path",
+			config:  config1,
```
gavv, line 542: Instead of config1/config2/config3 it would be cleaner to add baseURL to the table, and construct config inside loop.

request_test.go
```diff
@@ -522,31 +523,122 @@ func TestRequest_URLConcatenate(t *testing.T) {
...
-	}
+	const (
+		pathArgType1 = iota
+		pathArgType2 = iota
+	)
 
-	for _, req := range reqs {
-		req.Expect().chain.assertNotFailed(t)
-		assert.Equal(t, "http://example.com/path", client.req.URL.String())
+	requests := []struct {
```
gavv, line 531: Let's rename `requests` to `cases` for consistency with (most of) other tests.

request_test.go
```diff
@@ -522,31 +523,122 @@ func TestRequest_URLConcatenate(t *testing.T) {
...
+		name        string
+		config      Config
+		method      string
+		path        string
+		pathArg     string
+		pathArgType int
+		fullURL     string
+	}{
+		{
+			name:    "config1_empty_url_and_empty_path",
```
gavv, line 541: In most tests we don't use underscores, instead we mostly use spaces and commas, like:

```
empty url, empty path
```

gavv: Same for other tests.

gavv (commented):
Thanks for PR!! Here are a few comment

gavv:
You marked checkbox for "TestRequest_Conflicts", but it seems it's not modified. Do you plan to handle it? Otherwise I'll just create a new issue for that.

Darrellbor:
@gavv I missed `TestRequest_Conflicts` I think it's ok being in its own issue

request_test.go
```diff
@@ -504,82 +504,212 @@ func TestRequest_URLConcatenate(t *testing.T) {
 
 	reporter := NewAssertReporter(t)
```
gavv, line 505: Let's use reporter creation into subtest, so that it will use local `t, and subtests would be more isolated. Client also can be moved there.

request_test.go
```diff
@@ -504,82 +504,212 @@ func TestRequest_URLConcatenate(t *testing.T) {
...
-		Client:   client,
-		Reporter: reporter,
+	cases := []struct {
+		name      string
+		baseURL   string
+		method    string
+		path      string
+		pathArgs  []interface{}
+		setupFunc func(request *Request)
+		fullURL   string
```
gavv, line 514: Let's rename fullURL to expectedURL or something like this, to make it clear that it's not input parameter, but expected output.

request_test.go
```diff
@@ -504,82 +504,212 @@ func TestRequest_URLConcatenate(t *testing.T) {
...
+					tc.path,
+				)
+				tc.setupFunc(req)
+			} else {
+				req = NewRequestC(Config{
+					BaseURL:  tc.baseURL,
+					Client:   client,
+					Reporter: reporter,
+				}, tc.method, tc.path)
+			}
```
gavv, line 624: We can simplify this branch:

First, unconditionally create req like this:

```go
				req = NewRequestC(
					Config{
						BaseURL:  tc.baseURL,
						Client:   client,
						Reporter: reporter,
					},
					tc.method,
					tc.path,
					tc.pathArgs...,
				)
```

(it should work if pathArgs is nil)

Then, check if setupFunc is non-nil, and call it.

request_test.go
gavv: Same for TestRequest_URLOverwrite

request_test.go
```diff
@@ -504,82 +504,212 @@ func TestRequest_URLConcatenate(t *testing.T) {
...
+		},
+		{
+			name:    "empty path",
+			baseURL: "http://example.com",
+			method:  "GET",
+			path:    "",
+			fullURL: "http://example.com",
+		},
+		{
+			name:    "slash empty path",
```
gavv, line 531: nit: I suggest "empty path, url with slash" for consistency with other names.

Also maybe rename "with slash" / "without slash" to "url with slash" / "url without slash" for clarify? Otherwise one can think that "with slash" is about path.

request_test.go
```diff
@@ -504,82 +504,212 @@ func TestRequest_URLConcatenate(t *testing.T) {
...
-		NewRequestC(config2, "GET", "/path").WithURL("http://example.com"),
-		NewRequestC(config2, "GET", "path").WithURL("http://example.com"),
-		NewRequestC(config2, "GET", "/path").WithURL("http://example.com/"),
-		NewRequestC(config2, "GET", "path").WithURL("http://example.com/"),
+	cases := []struct {
+		name    string
+		baseURL string
+		method  string
+		path    string
+		fullURL string
```
gavv, line 643: Let's rename fullURL to something like overwriteURL to make clear its purpose.

request_test.go
```diff
@@ -854,60 +984,68 @@ func TestRequest_BasicAuth(t *testing.T) {
...
-
-	config1 := Config{
-		Client:   client1,
-		Reporter: reporter1,
+	cases := []struct {
+		name      string
+		method    string
+		path      string
+		url       string
+		secondURL string
```
gavv, line 992: I suggest a couple of changes:

* remove "url" and "secondURL" fields
* add setupFunc field that will call WithHost and/or WithHeader
* add expectedHost field that will define what host to assert

I think this way it would be more clear from table what we're testing.

gavv:
Thanks for update! See a few more comments.

gavv:
> @gavv I missed TestRequest_Conflicts I think it's ok being in its own issue

No worries; I pushed some minimal refactoring to master, I think it'd be enough for it for now.

gavv:
Thank you!

PR #337, opened 2023-03-18
Add DefaultFormatter.EnableRequests and EnableResponses

### LINKED ISSUE #159: Add DefaultFormatter.EnableRequests and EnableResponses
Add two flags `EnableRequests` and `EnableResponses` to `DefaultFormatter`, which enable printing `http.Request` and `http.Response` on failure. New flags should be disabled by default.

Request and response instances are already present in `AssertionContext` passed to formatter. Formatted request and response should be added to `FormatData` and used in failure template. Formatting can be done using `httputil`, like we do it in `DebugPrinter`.

Formatted response should include round trip time (again like in `DebugPrinter`).

Related discussion: https://github.com/gavv/httpexpect/discussions/137

See also #160.

gdguesser:
Closes #159

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,8 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml
-# Editor-based HTTP Client requests
-/httpRequests/
-# Datasource local storage ignored files
-/dataSources/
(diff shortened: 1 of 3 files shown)


gavv:
Thanks!

I've pushed follow-up commit: 70906db5e8513d771eb9d1b67aff4ce214c0a4c5

What I changed:

* the new feature seems pretty useful and the output is rather compact, so I enabled it by default by replacing DisableRequest/DisableResponse with EnableRequest/EnableResponse

* I added HaveRequest and HaveResponse to FormatData; they're not necessary in this case, just for consistency with other fields

* fixed nil checks: refIsNil is not needed because typed nils are not possible here; on the other hand we need to check if httpReq/httpResp is non-nil

* replaced panics with returns; formatter is not critical part, so it's better to skip a section instead of refusing to work at all, in case of a bug

* added trimming

gavv:
Also welcome to our [discord server](https://discord.gg/5SCPCuCWA9)!

PR #340, opened 2023-03-21
Improved coverage for array and object test 

frankweate:
closes gavv#184
array_test.go:
Added test cases to IsEqualUnordered(), NotEqualUnordered() , NotConsistsOf(), ContainsOnly(), NotContainsOnly(), Every(), Filter()

object_test.go:
Added test cases to ContainsValue(), Every(), Filter()

Please let me know if there is anything  you wish for me to add or change, Thanks!

### DIFF
--- a/array_test.go
+++ b/array_test.go
@@ -596,6 +596,7 @@ func TestArray_IsEqualUnordered(t *testing.T) {
 		value.NotEqualUnordered([]interface{}{"foo", 1234})
 		value.chain.assertNotFailed(t)
 		value.chain.clearFailed()
+
 	})
 
 	t.Run("with duplicates", func(t *testing.T) {
@@ -640,7 +641,7 @@ func TestArray_IsEqualUnordered(t *testing.T) {
 		value.chain.clearFailed()
 
 		value.NotEqualUnordered([]interface{}{123})
-		value.chain.assertNotFailed(t)
+		value.chain.assertFailed(t)
 		value.chain.clearFailed()
 	})
 
@@ -682,11 +683,7 @@ func TestArray_IsEqualUnordered(t *testing.T) {
 		value.chain.assertFailed(t)
 		value.chain.clearFailed()
 
-		value.IsEqualUnordered("123")
-		value.chain.assertFailed(t)
-		value.chain.clearFailed()
-
-		value.NotEqualUnordered("123")
+		value.NotEqualUnordered(func() {})
 		value.chain.assertFailed(t)
 		value.chain.clearFailed()
 	})
@@ -895,9 +892,6 @@ func TestArray_ConsistsOf(t *testing.T) {
 		value.chain.assertFailed(t)
 		value.chain.clearFailed()
 
-		value.NotConsistsOf(make(chan int)) // bad type make(chan int)
-		value.chain.assertFailed(t)
-		value.chain.clearFailed()
 	})
 
 	t.Run("canonization", func(t *testing.T) {
@@ -1261,10 +1255,6 @@ func TestArray_ContainsOnly(t *testing.T) {
 		value.NotContainsOnly("foo", 123, "bar")
 		value.chain.assertNotFailed(t)
 		value.chain.clearFailed()
-
-		value.NotContainsOnly(make(chan int)) // bad type make(chan int)
-		value.chain.assertFailed(t)
-		value.chain.clearFailed()
 	})
 
 	t.Run("with duplicates", func(t *testing.T) {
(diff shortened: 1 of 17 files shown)


array_test.go
```diff
@@ -666,7 +682,11 @@ func TestArray_IsEqualUnordered(t *testing.T) {
 		value.chain.assertFailed(t)
 		value.chain.clearFailed()
 
-		value.NotEqualUnordered(func() {})
```
gavv, line 669: I guess the block:

```go
		value.NotEqualUnordered(func() {})
		value.chain.assertFailed(t)
		value.chain.clearFailed()
```

was lost accidentally, we should restore it.

array_test.go
```diff
@@ -666,7 +682,11 @@ func TestArray_IsEqualUnordered(t *testing.T) {
...
+		value.IsEqualUnordered("123")
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
+
+		value.NotEqualUnordered("123")
 		value.chain.assertFailed(t)
 		value.chain.clearFailed()
```
gavv, line 691: Which branches are we covering with these two blocks? Aren't they already covered with:

```go
		value.IsEqualUnordered(func() {})
		value.chain.assertFailed(t)
		value.chain.clearFailed()

		value.NotEqualUnordered(func() {})
		value.chain.assertFailed(t)
		value.chain.clearFailed()
```

?

array_test.go
```diff
@@ -874,6 +894,10 @@ func TestArray_ConsistsOf(t *testing.T) {
 		value.NotConsistsOf(123, "foo")
 		value.chain.assertFailed(t)
 		value.chain.clearFailed()
+
+		value.NotConsistsOf(make(chan int)) // bad type make(chan int)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
```
gavv, line 900: Isn't this the same as:

```go
		value.NotConsistsOf(func() {})
		value.chain.assertFailed(t)
		value.chain.clearFailed()
```

which we already have?

array_test.go
```diff
@@ -1229,6 +1253,18 @@ func TestArray_ContainsOnly(t *testing.T) {
...
+
+		value.NotContainsOnly("foo", 123, "bar")
+		value.chain.assertNotFailed(t)
+		value.chain.clearFailed()
+
+		value.NotContainsOnly(make(chan int)) // bad type make(chan int)
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
```
gavv, line 1267: Isn't this the same as:

```go
		value.NotContainsOnly(func() {})
		value.chain.assertFailed(t)
		value.chain.clearFailed()
```

which we already have?

array_test.go
```diff
@@ -1482,6 +1518,13 @@ func TestArray_Every(t *testing.T) {
 		array.chain.assertNotFailed(t)
 	})
 
+	t.Run("nil func", func(t *testing.T) {
```
gavv, line 1521: Let's rename this to "invalid argument" and move to the end of the test, for consistency with other tests.

array_test.go
```diff
@@ -1660,6 +1703,14 @@ func TestArray_Filter(t *testing.T) {
 		filteredArray.chain.assertNotFailed(t)
 	})
 
+	t.Run("nil func", func(t *testing.T) {
```
gavv, line 1706: Let's rename this to "invalid argument" and move to the end of the test, for consistency with other tests.

object_test.go
```diff
@@ -729,6 +729,10 @@ func TestObject_ContainsValue(t *testing.T) {
 		value.NotContainsValue("XXX")
 		value.chain.assertNotFailed(t)
 		value.chain.clearFailed()
+
+		value.ContainsValue(make(chan int))
+		value.chain.assertFailed(t)
+		value.chain.clearFailed()
```
gavv, line 735: Let's extract this into subtest "invalid argument" and move it to the end of the test.

Let's also add the similar check for NotContainsValue to the same subtest.

object_test.go
```diff
@@ -1433,6 +1444,13 @@ func TestObject_Filter(t *testing.T) {
 		object.chain.assertNotFailed(t)
 	})
 
+	t.Run("nil func", func(t *testing.T) {
```
gavv, line 1447: Let's rename this to "invalid argument" and move to the end of the test.

gavv (commented):
Thanks for PR! Here are a few comments.

array_test.go
frankweate: Removed these 2 cases

array_test.go
frankweate: added this back

array_test.go
frankweate: you're right, it follows the same path in the function, so i've removed it

array_test.go
frankweate: removed it

array_test.go
frankweate: Changed to the consistency of other tests

array_test.go
frankweate: Changed format to the consistency of other tests

object_test.go
frankweate: added this new subtest and the Not containsValue check

object_test.go
```diff
@@ -1196,6 +1200,13 @@ func TestObject_Every(t *testing.T) {
...
```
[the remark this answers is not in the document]
frankweate: Changed format to the consistency of other tests

object_test.go
frankweate: Changed format to the consistency of other tests

PR #341, opened 2023-03-21
PanicReporter implementation

### LINKED ISSUE #248: Implement PanicReporter

a-camarillo:
Hello, this is my implementation to fix #248
Changes:
- add PanicReporter implementation
- add test coverage for PanicReporter
- edit docs to mention PanicReporter

Please let me know if this is not expected behavior or if anything needs to be added and I will make the appropriate changes.

### DIFF
--- a/expect.go
+++ b/expect.go
@@ -162,6 +162,8 @@ type Config struct {
 	//    (non-fatal / fatal failures using testify package)
 	//  - testing.T / FatalReporter
 	//    (non-fatal / fatal failures using standard testing package)
+	//	- PanicReporter
+	//	  (failures that panic to be used in multithreaded tests)
 	//  - custom implementation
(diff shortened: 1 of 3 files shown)


gavv:
Thanks for PR,

I've pushed a few small fixes: https://github.com/gavv/httpexpect/pull/341/commits/26ad20cef23d429f7ed58f6d9ae3f72e9555f11e

1. testing.TB backend is not needed for this reporter
2. use fmt.Sprintf to format message
3. cleanup test
4. expand comment

Issue #328, opened 2023-03-03
Add more multipart form tests


gavv:
TestRequest_BodyMultipart covers usage of WithMultipart + WithForm/WithFormField/WithFile/etc.

coveralls [reports](https://coveralls.io/builds/57516032/source?filename=request.go) that there are several uncovered branches in various methods, related to multipart form handling.

You can find all such places by searching for `if r.multipart` and checking for red blocks nearby.

We should add subtests to TestRequest_BodyMultipart to cover all such branches.

neverbeenthisweeb:
hi @gavv , can I work on this?

neverbeenthisweeb:
btw, I am thinking to have a mocked `Request.multipart` to simulate errors for some cases (such as [request.go#L1672](https://coveralls.io/builds/57516032/source?filename=request.go#L1672) in the red blocks)

so, the mocked `Request.multipart` will be something like

```go
type mockMultipartImpl struct{}

func (mm *mockMultipartImpl) WriteField(fieldname, value string) error {
    return errors.New("error from mock")
}
```

however, this requires us to update `Request.multipart` from `struct` into `interface` so that in can be mocked

and provide a custom setter for `Request.multipart`

```go
// Existing
func (r *Request) WithMultipart() *Request {

// New
type multipartWriterInterface interface {}

func (r *Request) withCustomMultipart(m multipartWriterInterface) *Request {
```

wdyt about this solution? is it a bit..._overkill_?

neverbeenthisweeb:
> btw, I am thinking to have a mocked Request.multipart to simulate errors for some cases (such as for .WriteField in [request.go#L1672](https://coveralls.io/builds/57516032/source?filename=request.go#L1672) in the red blocks)

@gavv I've created a draft to demonstrate the point above (https://github.com/gavv/httpexpect/issues/328#issuecomment-1479206229)

please check https://github.com/gavv/httpexpect/pull/344

thank you

LMK if you have concern or anything

PS. sry if the structure is a bit messy, but it should convey my point

gavv:
IIRC we used similar approach with `Request.sleepFn`

neverbeenthisweeb:
> Request.WithMultipart calls multipart.NewWriter and passes r.formbuf to it, which implements io.Writer. It should be enough for us to intercept this call and pass our own io.Writer that will return error when needed.

I see, thanks for the insight.

neverbeenthisweeb:
@gavv I've updated the PR based on the suggestion above 

[...]

PR #344, opened 2023-03-23
Add more multipart form tests

neverbeenthisweeb:
Issue: https://github.com/gavv/httpexpect/issues/328

Description:

- Add multipart writer mock


neverbeenthisweeb:
@gavv Ready for review

request_test.go
```diff
@@ -1237,6 +1238,76 @@ func TestRequest_BodyMultipart(t *testing.T) {
...
+			assertFn func(t *testing.T, req *Request)
+		}{
+			{
+				name: "with form",
+				reqFn: func() *Request {
+					req := NewRequestC(config, "POST", "url")
+					req.multipartFn = func(w io.Writer) *multipart.Writer {
+						return multipart.NewWriter(&failureWriter{})
+					}
+					req.WithMultipart()
```
gavv, line 1255: Seems this part:

```go
					req := NewRequestC(config, "POST", "url")
					req.multipartFn = func(w io.Writer) *multipart.Writer {
						return multipart.NewWriter(&failureWriter{})
					}
					req.WithMultipart()
```

is same for all subtests, so it can be extracted to the loop. And reqFn can accept req as argument instead of returning req.

request_test.go
```diff
@@ -1237,6 +1238,76 @@ func TestRequest_BodyMultipart(t *testing.T) {
...
+					req.WithForm(map[string]string{"foo": "bar"})
+					return req
+				},
+				assertFn: func(t *testing.T, req *Request) {
+					req.chain.assertFailed(t)
+				},
```
gavv, line 1261: Seems this part:

```go
				assertFn: func(t *testing.T, req *Request) {
					req.chain.assertFailed(t)
				},
```

is same for all subtests, so we can remove assertFn and move the code to the loop.

Also please use assertOK(t, false) instead of assertFailed(t), I've started a (slow) process of migrating from assertFailed and assertNotFailed to assertOK.

request_test.go
```diff
@@ -1237,6 +1238,76 @@ func TestRequest_BodyMultipart(t *testing.T) {
...
+				tc.assertFn(t, req)
+			})
+		}
+	})
+}
+
+// failureWriter is a writer that returns error.
+type failureWriter struct {
+	io.Writer
+}
```
gavv, line 1307: Most our are in mocks_test.go; let's move this one there too?

We can rename it to "mockFailureWriter". Or we can make it more generic and name it "mockWriter", and add "err" field that is used like this:

```
func (fw *failureWriter) Write(p []byte) (n int, err error) {
    if fw.err != nil {
        return 0, fw.err
    }
    return fw.Writer.Write(p)
}
```

gavv (commented):
Thanks for PR! A few comments.

request_test.go
neverbeenthisweeb: ok will do

request_test.go
neverbeenthisweeb: updated https://github.com/gavv/httpexpect/pull/344/commits/04042a6e44d71805c60bd675cf13ff067f5db142

request_test.go
neverbeenthisweeb: updated https://github.com/gavv/httpexpect/pull/344/commits/04042a6e44d71805c60bd675cf13ff067f5db142

gavv:
Thanks!

Issue #345, opened 2023-03-23
Refactor Number and String tests to use chain.assertOK


gavv:
Number and String have many fragments like this:

```go
				if tc.isEqual {
					NewNumber(reporter, tc.number).IsEqual(tc.value).
						chain.assertNotFailed(t)

					NewNumber(reporter, tc.number).NotEqual(tc.value).
						chain.assertFailed(t)
				} else {
					NewNumber(reporter, tc.number).NotEqual(tc.value).
						chain.assertNotFailed(t)

					NewNumber(reporter, tc.number).IsEqual(tc.value).
						chain.assertFailed(t)
				}
```

With recently added `chain.assertOK`, such fragments can be simplified to:

```go
				NewNumber(reporter, tc.number).IsEqual(tc.value).
					chain.assertOK(t, tc.isEqual)

				NewNumber(reporter, tc.number).NotEqual(tc.value).
					chain.assertOK(t, !tc.isEqual)
```

This should be done in all appropriate tests in `number_test.go` and `string_test.go`.

tsukasaI:
@gavv 
Hi.
I have opend PR #347 
Please check this.

gavv:
Thanks!

PR #347, opened 2023-03-23
Refactor Number and String tests to use chain.assertOK

tsukasaI:
Issue: #345 

Description:

Refactor tests
- string_test
- number_test

### DIFF
--- a/number_test.go
+++ b/number_test.go
@@ -196,12 +196,10 @@ func TestNumber_IsEqual(t *testing.T) {
 			t.Run(tc.name, func(t *testing.T) {
 				reporter := newMockReporter(t)
 
-				NewNumber(reporter, tc.number).
-					IsEqual(tc.value).
+				NewNumber(reporter, tc.number).IsEqual(tc.value).
 					chain.assertOK(t, tc.isEqual)
 
-				NewNumber(reporter, tc.number).
-					NotEqual(tc.value).
+				NewNumber(reporter, tc.number).NotEqual(tc.value).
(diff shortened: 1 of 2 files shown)


number_test.go
```diff
@@ -315,25 +303,13 @@ func TestNumber_InDelta(t *testing.T) {
...
-					InDelta(tc.value, tc.delta).
-					chain.assertNotFailed(t)
-			} else {
-				NewNumber(reporter, tc.number).
-					InDelta(tc.value, tc.delta).
-					chain.assertFailed(t)
-			}
+			NewNumber(reporter, tc.number).
+				InDelta(tc.value, tc.delta).
+				chain.assertOK(t, tc.isInDelta)
```
gavv, line 308: nit: could you please reformat everything to a more compact 2-line format like this:

```go
			NewNumber(reporter, tc.number).InDelta(tc.value, tc.delta).
				chain.assertOK(t, tc.isInDelta)
```

gavv (commented):
Thanks for PR! One small comment

number_test.go
tsukasaI: thsnk you for athe comment.
reformated!

gavv: Sorry, I actually meant to switch from 3-line to 2-line format, not vise versa :)

I've pushed a commit: https://github.com/gavv/httpexpect/pull/347/commits/88fd692a0d3dc9dedb12d00157e65da147ff1ab6

gavv:
Thank you!

PR #363, opened 2023-03-30
convert value tests to table-driven style

### LINKED ISSUE #353: Convert Value tests to table-driven style
Convert two tests to table-driven style:

* TestValue_IsEqual
* TestValue_InList

`TestNumber_IsEqual/basic` can be used as a reference for table-driven style.

diPhantxm:
Closes #353

### DIFF
--- a/value_test.go
+++ b/value_test.go
@@ -551,35 +551,80 @@ func TestValue_IsBoolean(t *testing.T) {
 }
 
 func TestValue_IsEqual(t *testing.T) {
-	reporter := newMockReporter(t)
+	t.Run("basic", func(t *testing.T) {
+		cases := []struct {
+			name         string
+			value1       interface{}
+			value2       interface{}
+			wantEqual    chainResult
+			wantNotEqual chainResult
+		}{
+			{
+				name:         "compare equivalent values (strings)",
+				value1:       "baz",
+				value2:       "baz",
+				wantEqual:    success,
+				wantNotEqual: failure,
+			},
+			{
+				name:         "compare equivalent values (maps)",
+				value1:       map[string]interface{}{"foo": "bar"},
+				value2:       map[string]interface{}{"foo": "bar"},
+				wantEqual:    success,
+				wantNotEqual: failure,
+			},
+			{
+				name:         "compare non-equivalent values",
+				value1:       map[string]interface{}{"foo": "bar"},
+				value2:       "baz",
+				wantEqual:    failure,
+				wantNotEqual: success,
+			},
+			{
+				name:         "compare nil values",
+				value1:       nil,
+				value2:       nil,
+				wantEqual:    success,
+				wantNotEqual: failure,
+			},
+			{
+				name:         "compare nil and nil-value",
+				value1:       nil,
+				value2:       map[string]interface{}(nil),
+				wantEqual:    success,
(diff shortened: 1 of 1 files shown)


value_test.go
```diff
@@ -588,32 +633,105 @@ func TestValue_InList(t *testing.T) {
...
+				name:          "no list",
+				value:         data1,
+				list:          nil,
+				wantInList:    failure,
+				wantNotInList: failure,
+			},
+			{
+				name:          "empty list",
+				value:         data1,
+				list:          []interface{}{},
```
diPhantxm, line 654: While I was doing this issue, I noticed that this is not the same as NewValue(reporter, data1).InList([]interface{}{}).
Because in tests it's an empty list, but if you do it as it was originally, just in-line, it will be a list containing a slice of interfaces.
I've seen it in `TestNumber_InList/basic`

value_test.go
```diff
@@ -551,35 +551,80 @@ func TestValue_IsBoolean(t *testing.T) {
...
+				wantEqual:    failure,
+				wantNotEqual: success,
+			},
+			{
+				name:         "compare value and func",
+				value1:       map[string]interface{}{"foo": "bar"},
+				value2:       func() {},
+				wantEqual:    failure,
+				wantNotEqual: failure,
+			},
```
gavv, line 610: When you pass `func() {}`, the reported failure is that the argument can not be json-mashalled.

So I suggest to extract it to separate subtest "invalid argument" (like we do in many other tests).

After doing this, it seems that  it's always true that `wantEqual ==  !wantNotEqual`, so we can remove wantNotEqual here and use `!wantEqual` instead.

value_test.go
```diff
@@ -551,35 +551,80 @@ func TestValue_IsBoolean(t *testing.T) {
...
+		}
+	})
 }
 
 func TestValue_InList(t *testing.T) {
-	reporter := newMockReporter(t)
-
 	data1 := map[string]interface{}{"foo": "bar"}
 	data2 := "baz"
 	data3 := struct {
```
gavv, line 630: What do you think about embedding these values into tests? Most of them are one-line, so it's not much code duplication, but I think tests will be more readable (you don't need to jump to the top and see what is "data1" to understand the test).

value_test.go
```diff
@@ -588,32 +633,105 @@ func TestValue_InList(t *testing.T) {
...
+				wantInList:    failure,
+				wantNotInList: failure,
+			},
+			{
+				name:          "list of string and func",
+				value:         data1,
+				list:          []interface{}{data2, func() {}},
+				wantInList:    failure,
+				wantNotInList: failure,
+			},
```
gavv, line 720: I suggest to extract these checks into "invalid argument" subtest as well.

gavv (commented):
Thanks for PR!

value_test.go
gavv: Good catch, let's keep the new version.

I think both "empty list" and "no list" should be moved to "invalid argument" subtest, because they both are reported as usage error.

Also let's rename "no list" to "nil list".

diPhantxm:
I extracted `invalid argument` test cases into separate subtest, embedded data values into test case, but left definition of a structure on top, I think it looks better. Please check out

gavv:
Thanks!

PR #364, opened 2023-03-31
refactored some array tests to table driven style

frankweate:
closes gavv#351
Refactored the subtests in:

- TestArray_ConsistsOf
- TestArray_Contains
- TestArray_ContainsAll
- TestArray_ContainsAny
- TestArray_ContainsOnly
- TestArray_HasValue

to use a table-driven style. 

Let me know if you want me to make any changes or add anything :)

### DIFF
--- a/array_test.go
+++ b/array_test.go
@@ -892,7 +892,8 @@ func TestArray_ConsistsOf(t *testing.T) {
 		)
 		reporter := newMockReporter(t)
 
-		assert.Equal(t, []interface{}{123.0, 456.0}, NewArray(reporter, []interface{}{123, 456}).Raw())
+		assert.Equal(t, []interface{}{123.0, 456.0},
+			NewArray(reporter, []interface{}{123, 456}).Raw())
 
 		NewArray(reporter, []interface{}{123, 456}).ConsistsOf(myInt(123), 456.0).
 			chain.assert(t, success)
@@ -1007,7 +1008,8 @@ func TestArray_Contains(t *testing.T) {
 			})
 		}
 		reporter := newMockReporter(t)
-		assert.Equal(t, []interface{}{123.0, 456.0}, NewArray(reporter, []interface{}{123, 456}).Raw())
+		assert.Equal(t, []interface{}{123.0, 456.0}, NewArray(reporter,
+			[]interface{}{123, 456}).Raw())
 
 	})
 
@@ -1080,7 +1082,8 @@ func TestArray_ContainsAll(t *testing.T) {
 
 		reporter := newMockReporter(t)
 
-		assert.Equal(t, []interface{}{123.0, 456.0}, NewArray(reporter, []interface{}{123, 456}).Raw())
+		assert.Equal(t, []interface{}{123.0, 456.0},
+			NewArray(reporter, []interface{}{123, 456}).Raw())
 
 		NewArray(reporter, []interface{}{123, 456}).ContainsAll(myInt(123), 456.0).
 			chain.assert(t, success)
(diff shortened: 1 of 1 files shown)


array_test.go
```diff
@@ -842,234 +842,237 @@ func TestArray_InList(t *testing.T) {
...
 
-		value.NotConsistsOf(func() {})
-		value.chain.assert(t, failure)
-		value.chain.clear()
+		NewArray(reporter, []interface{}{}).NotConsistsOf(func() {}).
+			chain.assert(t, failure)
 	})
 }
 
 func TestArray_Contains(t *testing.T) {
```
gavv, line 916: Sorry, I completely forgot about one thing when I was creating this issue.

In next major release (v3), Array.(Not)Contains will be changed to accept only a single argument. Multiple arguments will be allowed only for ContainsAny, ContainsAll, etc.

Since you're modifying these tests now, could you please change the tests to check only for single argument cases? So that the tests will be ready for v3 change.

array_test.go
```diff
@@ -842,234 +842,237 @@ func TestArray_InList(t *testing.T) {
...
-		value.NotContains(func() {})
-		value.chain.assert(t, failure)
-		value.chain.clear()
+		NewArray(reporter, []interface{}{}).NotContains(func() {}).
+			chain.assert(t, failure)
 	})
 }
 
 func TestArray_ContainsAll(t *testing.T) {
 	t.Run("basic", func(t *testing.T) {
```
gavv, line 1028: Could you please also add one case, where:

* we pass multiple elements to ContainsAll (e.g. 2)
* they don't have duplicates (e.g. "foo", "bar")
* they are a subset of the array (e.g. array is "foo", "bar", baz")

Because it seems that for multiple arguments we check only "corner cases" (when we pass array copy or duplicate), but not checking a "regular" case.

gavv: Same for TestArray_ContainsAny

array_test.go
```diff
@@ -1079,313 +1082,313 @@ func TestArray_ContainsAll(t *testing.T) {
...
-				"a": "b",
+		cases := []struct {
+			name            string
+			index           int
+			value           interface{}
+			wantHasValue    chainResult
+			wantNotHasValue chainResult
+		}{
+			{
+				name:            "zeroth element check",
```
gavv, line 1338: zeroth -> zero ?

array_test.go
```diff
@@ -1079,313 +1082,313 @@ func TestArray_ContainsAll(t *testing.T) {
...
-
-		array.HasValue(2, map[string]interface{}{"a": "b"})
-		array.chain.assert(t, success)
-		array.chain.clear()
-
-		array.NotHasValue(2, map[string]interface{}{"a": "b"})
-		array.chain.assert(t, failure)
-		array.chain.clear()
+			{
+				name:            "first element check",
```
gavv, line 1345: Well it's not first, I think zero element is first :)

array_test.go
```diff
@@ -1079,313 +1082,313 @@ func TestArray_ContainsAll(t *testing.T) {
...
+			chain.assert(t, failure)
 
-		value.NotContainsOnly(func() {})
-		value.chain.assert(t, failure)
-		value.chain.clear()
 	})
 }
 
 func TestArray_HasValue(t *testing.T) {
 	t.Run("basic", func(t *testing.T) {
```
gavv, line 1329: Could you also add a case when index is in bound, but value is not equal?

gavv (commented):

frankweate:
I added your requested changes! Let me know if there is anything else for me to add or change :)

gavv:
Thanks for update!

I've pushed a small follow-up commit with the some cosmetic fixes 24ba8e147e828f157247ff1d3da68f5fa999ee29

* fixed formatting in some tests to be similar to the other tests
* renamed testArray to array (we usually use such names in other tests) and reordered fields (first goes array, then value)
* minor issues

PR #365, opened 2023-03-31
fixes #357, refactored changes

### LINKED ISSUE #357: Refactor Websocket tests
After introducing new chain.assert() method, we can simplify many websocket tests.

In `websocket_test.go`, find all table-driven tests that have `assertOk` field and refactor them as follows:

* replace `assertOk bool` with `result chainResult`
* instead of `assertOk: true`, use `result: success`
* instead of `assertOk: false`, use `result: failure`
* instead of:

    ```go
	if tc.assertOk {
		ws.chain.assert(t, success)
	} else {
		ws.chain.assert(t, failure)
	}
    ```

    use:

    ```go
	ws.chain.assert(t, tc.result)
    ```

antara-gandhi:
fixes #357 

refactored all table driven tests using `assertOk` in `websocket_test.go`.  Let me know if there are any changes necessary!

### DIFF
--- a/websocket_test.go
+++ b/websocket_test.go
@@ -133,7 +133,7 @@ func TestWebsocket_Expect(t *testing.T) {
 	cases := []struct {
 		name     string
 		args     args
-		assertOk bool
+		result chainResult
 	}{
 		{
 			name: "success",
@@ -141,7 +141,7 @@ func TestWebsocket_Expect(t *testing.T) {
 				wsPreSteps: noWsPreSteps,
 				wsConn:     &mockWebsocketConn{},
 			},
-			assertOk: true,
+			result: success,
 		},
 		{
 			name: "fail to read message from conn",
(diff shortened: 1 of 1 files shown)


antara-gandhi:
quick aside - what is the difference in chain.assert() vs assertOk here? would we rather use chain.assert() because it is faster? are there any downsides to doing so vs using assertOk?

gavv:
Thanks for PR!

> quick aside - what is the difference in chain.assert() vs assertOk here? would we rather use chain.assert() because it is faster? are there any downsides to doing so vs using assertOk?

It's just a small refactoring to make code cleaner and more compact. With boolean assertOk field we're duplicating this piece:

```
if tc.assertOk {
	ws.chain.assert(t, success)
} else {
	ws.chain.assert(t, failure)
}
```

in every test. With chainResult we replace it with one line.

Issue #326, opened 2023-03-03
Refactor and improve TestRequest_URLInterpolate


gavv:
TestRequest_URLInterpolate covers usage of WithPath and WithPathObject.

Two changes are needed:

1. Split test into subtests, so that each request variable will be isolated and tested in its own subtest.

2. Add subtests to cover the following cases (coveralls reports that they are [uncovered](https://coveralls.io/builds/57516032/source?filename=request.go)):

   - WithPath: invalid interpol string (e.g. unmatched brace)
   - WithPath: key not found
   - WithPath: value is nil
   - WithPathObject: invalid interpol string
   - WithPathObject: key not found
   - WithPathObject: value is nil

Related: #283

gavv:
Landed!

PR #367, opened 2023-04-01
refactor request url interpolate tests

diPhantxm:
#326 
@gavv Hi! Check out please. I split tests into subtests, but I couldn't think of a case when `invalid interpol string` error is returned. And there was an issue with checking value on nil, it was duplicated in [here](https://github.com/gavv/httpexpect/blob/master/request.go#L964) and [here](https://github.com/gavv/httpexpect/blob/master/request.go#L877)

### DIFF
--- a/request.go
+++ b/request.go
@@ -145,39 +145,42 @@ func newRequest(
 }
 
 func (r *Request) initPath(opChain *chain, path string, pathargs ...interface{}) {
-	var n int
+	if len(pathargs) != 0 {
+		var n int
 
-	path, err := interpol.WithFunc(path, func(k string, w io.Writer) error {
-		if n < len(pathargs) {
-			if pathargs[n] == nil {
-				opChain.fail(AssertionFailure{
-					Type:   AssertValid,
-					Actual: &AssertionValue{pathargs},
-					Errors: []error{
-						fmt.Errorf("unexpected nil argument at index %d", n),
-					},
-				})
+		var err error
+		path, err = interpol.WithFunc(path, func(k string, w io.Writer) error {
+			if n < len(pathargs) {
+				if pathargs[n] == nil {
+					opChain.fail(AssertionFailure{
+						Type:   AssertValid,
+						Actual: &AssertionValue{pathargs},
+						Errors: []error{
+							fmt.Errorf("unexpected nil argument at index %d", n),
+						},
+					})
+				} else {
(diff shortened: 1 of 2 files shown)


gavv:
Thanks for PR!

> I split tests into subtests.

Looks fine.

One suggestion: I'd merge two "basic" subtests into one, rename it to "constructor" (because it tests how NewRequestC works), and make it table-driven, with the following table fields: path, path args, expected url.

> I couldn't think of a case when invalid interpol string error is returned

Seems it can't be reached with current implementation. The error could be triggered only if interpol string is invalid, e.g. there is unmatched brace. You already have a test for it. However, this error will be handled early by NewRequestC, and WithPath just won't be executed.

After looking at the code a bit, I found it's strange that NewRequestC tries to perform interpolation even if there are not pathargs given. It would be safer to avoid messing with interpol in this case, e.g. modify initPath as follows:

```go
if len(pathargs) != 0 {
    <all the interpolation code goes here>
}

r.path = path
```

AND, magically, if we do this, we'll also be able to cover both errors (in initPath and in withPath), by providing invalid interpol string and providing args either to constructor or to WithPath.

> And t[here](https://github.com/gavv/httpexpect/blob/master/request.go#L964) was an issue with checking value on nil, it was duplicated in [here](https://github.com/gavv/httpexpect/blob/master/request.go#L877) and here

Good catch.

gavv:
Could you please also cover this branch? https://coveralls.io/builds/58476221/source?filename=request.go#L153

(nil patharg passed to constructor)

diPhantxm:
Yeah, sure

diPhantxm:
> After looking at the code a bit, I found it's strange that NewRequestC tries to perform interpolation even if there are not pathargs given. It would be safer to avoid messing with interpol in this case, e.g. modify initPath as follows

If I add this if, the following is success, but must be failure:
```Golang
req := NewRequestC(config, "GET", "{arg")
req.chain.assert(t, failure)
```
And I don't know why, but it also affects interpolation in constructor even if it doesn't seem to have any affect

diPhantxm:
I managed to fix the problem. It was because of assigning to a new variable `path` inside if-statement

gavv:
Thanks!

Based on your changes, I proceeded with refactoring of interpol tests in this commit: c3059a97fa350e4067b9585917c35aa786bd6f25

PR #368, opened 2023-04-02
Move e2e tests to e2e folder (#272)

regularcoder:


regularcoder:
Please can you review and merge @gavv?

regularcoder:
@gavv I fixed conflicts with latest master commits.

lint has been failing because Raw() has been deprecated but the tests still use it. I'm assuming we don't want to delete the test usages until the deprecated function is deleted as well.

gavv:
@regularcoder Thanks!

> lint has been failing because Raw() has been deprecated but the tests still use it. I'm assuming we don't want to delete the test usages until the deprecated function is deleted as well.

Can we switch those tests to using Conn instead of Raw?

Raw will be removed in future releases.

regularcoder:
Most of them use both Raw and Conn. So I thought it was intentional - to
keep tests using it until it's completely removed.

But sure I can remove the Raw usages and add Conn instead wherever
necessary.

gavv:
https://github.com/gavv/httpexpect/issues/272

Issue #265, opened 2023-02-01
Add unit test for error propagation


gavv:
Add a test to expect_test.go, which checks the following.

After an object's operation (e.g. Array.IsEqual) fails, we expect that:

* [x] subsequent operations on that object will be no-op, i.e. they won't report failures
* [x] newly created child objects (e.g. Array.Length) will have failed chain and won't report failures
* [x] previously created nested object won't be affected, i.e. they will report failures
* [x] newly created child objects of objects' parent won't be affected, i.e. they will report failures
* [x] previously created child objects of objects' parent won't be affected, i.e. they will report failures

This behavior is described [in docs](https://pkg.go.dev/github.com/gavv/httpexpect/v2#hdr-Failure_handling).

This test will be partially similar to TestExpect_Branches, however it should set different reporters to different objects to check if this or that action causes failure reporting or not.

neverbeenthisweeb:
hi @gavv can I work on this?

gavv:
Sure, thanks!

neverbeenthisweeb:
@gavv question

what does the syntax `flags&(flagFailed|flagFailedChildren)` do here? (or...what is this syntax called by? i'm clueless to figure out the keyword to google :umbrella: )

```go
if flags&(flagFailed|flagFailedChildren) == 0 {
    handler.Success(&context)
}
```
https://github.com/gavv/httpexpect/blob/master/chain.go#L387

thanks

gavv:
Yes, it's bitwise AND. Unlike logical and (&&), it allows working with separate buts in a single integer value instead of treating the whole value as true or false. 

Individual buts are used here to represent different flags. The first bit is for one flag, second for another, and so on. If the corresponding bit is 1, the flag is considered to be set; if the bit is 0, the flag is unset.

It's a common idiom for representing flags. It's especially useful when you have many flags, so instead of many boolean fileds you need only one integer field. Less memory and less clutter in code.

gavv:
Random article from google: https://blog.podkalicki.com/bit-level-operations-bit-flags-and-bit-masks/#:~:text=This%20is%20done%20by%20using,%E2%80%9C0%E2%80%9D%20(zero).

gavv:
The particular snippet you mentioned checks that both of the specified flags are unset, i.e. both of the corresponding  bits are zero.

neverbeenthisweeb:
ok thanks for the answer and references

_today i learned_

neverbeenthisweeb:
[...]

PR #369, opened 2023-04-02
Add unit test for error propagation

neverbeenthisweeb:
Issue: https://github.com/gavv/httpexpect/issues/265


expect_test.go
```diff
@@ -341,6 +341,99 @@ func TestExpect_Branches(t *testing.T) {
...
+		val := arr.Value(0)
+		val.IsEqual("bar") // Child does not report failure
+
+		arr.chain.assertFlags(t, (flagFailed | flagFailedChildren))
+		val.chain.assertFlags(t, flagFailed)
+		assert.Equal(t, true, reporter.reported)
+		assert.Equal(t, 1, ctr)
+	})
+
+	t.Run("previously created nested object", func(t *testing.T) {
```
gavv, line 378: nit: let's replace "nested" with "child" for consistency

expect_test.go
```diff
@@ -341,6 +341,99 @@ func TestExpect_Branches(t *testing.T) {
...
+		arr.IsEmpty()
+		val.IsEqual("bar") // Child reports failure
+
+		arr.chain.assertFlags(t, (flagFailed | flagFailedChildren))
+		val.chain.assertFlags(t, flagFailed)
+		assert.Equal(t, true, reporter.reported)
+		assert.Equal(t, 2, ctr)
+	})
+
+	t.Run("newly created child objects of object's parent", func(t *testing.T) {
```
gavv, line 396: nit: let's rename it to `newly created children of parent`

expect_test.go
```diff
@@ -341,6 +341,99 @@ func TestExpect_Branches(t *testing.T) {
...
+
+		arr := NewArray(reporter, []interface{}{"foo"})
+		arr.IsEmpty().IsEmpty() // The second .IsEmpty() call does not report failure
+
+		arr.chain.assertFlags(t, flagFailed)
+		assert.Equal(t, true, reporter.reported)
+		assert.Equal(t, 1, ctr)
+	})
+
+	t.Run("newly created child objects", func(t *testing.T) {
```
gavv, line 360: nit: for clarity, let's use `child` or `children` instead of `child object(s)` here and in other tests.

expect_test.go
```diff
@@ -341,6 +341,99 @@ func TestExpect_Branches(t *testing.T) {
...
+func TestExpect_ErrorPropagation(t *testing.T) {
+	t.Run("subsequent operations", func(t *testing.T) {
+		ctr := 0
+		reporter := newMockReporter(t)
+		reporter.reportCb = func() {
+			ctr++
+		}
+
+		arr := NewArray(reporter, []interface{}{"foo"})
+		arr.IsEmpty().IsEmpty() // The second .IsEmpty() call does not report failure
```
gavv, line 353: Here and in all other tests: could you rework tests to check flags and counter of all values after every operation? And also maybe add some comments.

In this case, it would be like:

```go
// failing operation
arr.IsEmpty()
arr.chain.assertFlags(...)
assert.Equal(t, 1, ctr) // reported

// subsequent failing operation on the same object
arr.IsEmpty()
arr.chain.assertFlags(...)
assert.Equal(t, 1, ctr) // second failure was not reported
``` 

(I think we can remove checks `assert.Equal(t, true, reporter.reported)` because we anyway check counter)

There are two benefits of this approach:

* tests are much more readable (though also more bloated), because by looking at the test you can understand what is going on at each step; with current test, you have to guess

* we can be sure that the counter or flag was set by exact desired operations, not some other, because we do a check after each operation

neverbeenthisweeb: Ok, I am agree with this. I will do the update on this PR.

Thanks!

neverbeenthisweeb:
@gavv Updated. Please check

Thanks!

gavv:
Thanks!

Small fix: https://github.com/gavv/httpexpect/pull/369/commits/aaa07803f93fc07a6ba039bc5abdc8cbdb0f9e37

- renames
- add checks for parent too

Issue #360, opened 2023-03-29
Convert canon tests to table-driven style


gavv:
Convert each of the following tests to table-driven style:

* TestCanon_Number
* TestCanon_Array
* TestCanon_Map
* TestCannon_Decode

VikranthBala:
@gavv Can I work on this?

gavv:
Sure, thanks!

VikranthBala:
Hey, I think I am done with tests, can I know, if there is an naming convention to be followed, for the file and the testCases, as we already have existing tests. How Should I move forward with it?

gavv:
I assumed that we just change canon_test.go, and I guess test names are not changed, they just are split into subtests (using table).

You can take a look at TestCannon_Decode for an example of naming subtests.

VikranthBala:
Hey @gavv  , I have raised a PR, can you have a look at it, and let me know, if that is how I should be doing it, or should I change it. PR: #370

VikranthBala:
I have raised a new PR, as I had some issues with the previous PR. The new PR is at #395. Do let me know if I have to make any changes, Thanks. If I made any mistakes, do let me know. Thanks @gavv

gavv:
Thanks. BTW, you can always just force-push to PR's branch, I think there is no real need to create a new PR.

PR #370, opened 2023-04-02
converted canon tests to table-driven style

VikranthBala:


VikranthBala:
issue:[360](https://github.com/gavv/httpexpect/issues/360)

gavv:
Thanks for PR!

A few comments, mostly regarding code style in this project (please take a look at existing tests in other files):

* for test case struct, we usually use use anonymous struct
* when there are only a few fields, we don't use separate structs for "input" and "output" of test, just a single anonymous struct
* instead of "description", we usually use "name"
* instead of "testCase", we usually use "tc"
* it's better to create fresh chain inside every subtest, instead of sharing state between them
* we usually define auxiliary types like myInt and myArray locally inside each test
* instead of `ok bool` (true or false), we usually use `result chainResult` (success or failure); this way, instead of:

    ```go
			if testCase.output.ok {
				assert.Equal(testCase.output.val, val)
				chain.assert(t, success)
			} else {
				chain.assert(t, failure)
			}
    ```

    we can write:

    ```go
			chain.assert(t, testCase.output.result)
			if testCase.output.result {
				assert.Equal(testCase.output.val, val)
			}
    ```

VikranthBala:
@gavv, I have made the changes, can you have a look at it, and please let me know, if I have to make any changes. 
Thank you

gavv:
When it's ready for review, please leave a comment or use "request review" button.

VikranthBala:
@gavv, I have made the changes, considering the previous suggestions. Do let me know, if any more changes are required

gavv:
Thanks for update, a few more cosmetic issues:

* we usually use separate "cases" variable for the table; both variants are good, but I'd prefer to keep all tests in similar format https://github.com/gavv/httpexpect/blob/eeead3e581cbee0068546a6502dd7a2c293ffaa5/number_test.go#L162

* we usually don't use `assert.New(t)`, but instead `assert.SomeAssertion(t, ...)` functions; again, unification is preferred

* please cleanup subtest names a bit: no need to repeat test name in subtest name; no need to mention specific values in subtest name; e.g. `testCanon_Number: input '123' as int` can become `input is int`, and so on

* please remove unnecessary `chain.clear()` calls

* instead of `result: true` / `result: false` please use `result: success` / `result: failure`

Issue #356, opened 2023-03-29
Convert WebsocketMessage tests to table-driven style


gavv:
Convert each test to table-driven style:

* TestWebsocketMessage_CloseMessage
* TestWebsocketMessage_TextMessage
* TestWebsocketMessage_BinaryMessage
* TestWebsocketMessage_MatchTypes
* TestWebsocketMessage_MatchCodes

`TestNumber_IsEqual/basic` can be used as a reference for table-driven style.

aliziyacevik:
Hello there, may I handle this one ?

gavv:
You're welcome, thanks!

gavv:
Landed

PR #375, opened 2023-04-05
websocket message tests have been converted to table-driven tests

aliziyacevik:
https://github.com/gavv/httpexpect/issues/356


aliziyacevik:
Please tell me if you have any problem so I can change the code according to your liking.. thanks.

aliziyacevik:
I accidentally drafted the pull request xd @coveralls

websocket_message_test.go
```diff
@@ -76,231 +76,435 @@ func TestWebsocketMessage_Alias(t *testing.T) {
...
-	msg.chain.clear()
-
-	msg.NotCloseMessage()
-	msg.chain.assert(t, failure)
-	msg.chain.clear()
-
-	msg.BinaryMessage()
-	msg.chain.assert(t, failure)
-	msg.chain.clear()
+	msg := NewWebsocketMessage(reporter, websocket.CloseMessage, nil, 10)
```
gavv, line 79: Instead of reusing same message and clearing it's chain, could you rework test to create a fresh reporter and message inside each subtest?

This is a much cleaner approach, because we can be sure that subtests don't affect each other, and we're trying to gradually switch tests to these approach.

websocket_message_test.go
```diff
@@ -76,231 +76,435 @@ func TestWebsocketMessage_Alias(t *testing.T) {
...
 
-	msg.NotBinaryMessage()
-	msg.chain.assert(t, success)
-	msg.chain.clear()
+	cases := []struct {
+		name   string
+		fn     func() *WebsocketMessage
```
gavv, line 83: Seems `fn` is not used.

websocket_message_test.go
```diff
@@ -76,231 +76,435 @@ func TestWebsocketMessage_Alias(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			tc.testFn().chain.assert(t, tc.want)
+			msg.chain.clear()
+		})
+	}
 
-	msg.NotTextMessage()
-	msg.chain.assert(t, success)
-	msg.chain.clear()
+	cases2 := []struct {
```
gavv, line 132: Can we move each group of cases (cases, cases2, cases3) into a subtest, ideally with a meaningful name? (so it will be three subtests, each one with a table with sub-subtests).

websocket_message_test.go
gavv: Same for other tests.

websocket_message_test.go
gavv: Same for other tests.

websocket_message_test.go
gavv: Same for other tests.

websocket_message_test.go
```diff
@@ -76,231 +76,435 @@ func TestWebsocketMessage_Alias(t *testing.T) {
...
-	msg.chain.clear()
-
-	msg.NotType(websocket.TextMessage, websocket.BinaryMessage)
-	msg.chain.assert(t, failure)
-	msg.chain.clear()
+	cases := []struct {
+		name        string
+		typ         int
+		testTyp     int
+		testTyp2    int
```
gavv, line 398: Instead of testTyp + testTyp2, I suggest to have a single slice testTypes and pass it to Type() using `...`.

gavv: Same for tc1 and tc2 in TestWebsocketMessage_MatchCodes.

gavv (commented):
Nice! Thanks for PR. A few comments.

websocket_message_test.go
aliziyacevik: done

websocket_message_test.go
aliziyacevik: done. But i divided case 1 into multiple subtests since there are different methods are used in case 1.

websocket_message_test.go
aliziyacevik: done

websocket_message_test.go
aliziyacevik: removed.

gavv:
Thanks for update!

> But i divided case 1 into multiple subtests since there are different methods used in case 1.

Actually I think it's pretty easy to use table-driven approach for them too. In the table entry, you can either store [pointer to method](https://go.dev/play/p/XF1AlEJgTnn), and use it like this:

```go
fn: (*WebsocketMessage).TextMessage,
```

or store lambda like this:

```go
fn: func(m *WebsocketMessage) { m.TextMessage() },
```

What do you think?

aliziyacevik:
converted case 1. thanks for the good review. @gavv

gavv:
Thanks!

PR #376, opened 2023-04-05
add WithReporter and WithAssertionHandler

diPhantxm:
* add WithReporter and WithAssertionHandler
* add tests and documentation in comments


request.go
```diff
@@ -1926,6 +1926,62 @@ func (r *Request) WithMultipart() *Request {
...
+//		func(message string, args ...interface{}) {
+//			// reporter code here
+//		}),
+//	req.WithReporter(reporter)
+func (r *Request) WithReporter(reporter Reporter) *Request {
+	return r.WithAssertionHandler(&DefaultAssertionHandler{
+		Reporter:  reporter,
+		Formatter: r.config.Formatter,
+	})
+}
```
gavv, line 1954: We shouldn't call WithAssertionHandler from WithReporter, because they should use different names for r.chain.enter. I think we can just copy-paste implementation, since it's small.

request.go
```diff
@@ -1926,6 +1926,62 @@ func (r *Request) WithMultipart() *Request {
...
 }
 
+//	req := NewRequestC(config, "GET", "/path")
+//	req.WithClient(&http.Client{
+//	  Transport: &http.Transport{
+//		DisableCompression: true,
+//	  },
+//	})
+
+// WithReporter sets reporter.
```
gavv, line 1936: Let's clarify this a bit: `WithReporter sets reporter to be use for this request`.

request.go
```diff
@@ -1926,6 +1926,62 @@ func (r *Request) WithMultipart() *Request {
...
+//
+// The new reporter overwrites AssertionHandler.Reporter. It will be used to
+// report formatted fatal failure messages
```
gavv, line 1939: Not quite true. It overwrites the whole AssertionHandler, not AssertionHandler.Reporter. And the new AssertionHandler is DefaulyAssertionHandler with specified reporter + existing formatter + nil Logger. Could you please mention all this in comment?

request.go
```diff
@@ -1926,6 +1926,62 @@ func (r *Request) WithMultipart() *Request {
...
+
+// WithAssertionHandler sets assertion handler.
+//
+// The new handler overwrites assertion handler. It will be used to
+// format and report test Failure or Success.
+//
+// Example:
+//
+//	req := NewRequestC(config, "GET", "http://example.com/path")
+//	req.WithAssertionHandler(DefaultAssertionHandler)
```
gavv, line 1964: Could you please use a correct argument in example, instead of "DefaultAssertionHandler".

I mean, something like:

```go
&DefaultAssertionHandler{
   <initializers here>
```

request_test.go
```diff
@@ -3726,3 +3726,40 @@ func TestRequest_Panics(t *testing.T) {
 		assert.Panics(t, func() { newRequest(newMockChain(t), config, "GET", "") })
 	})
 }
+
+func TestRequest_Reporter(t *testing.T) {
```
gavv, line 3730: Let's move new tests to the top, right after TestRequest_Basic.

request_test.go
```diff
@@ -3726,3 +3726,40 @@ func TestRequest_Panics(t *testing.T) {
...
+		Client:         &mockClient{},
+		Reporter:       reporter,
+	}
+
+	req := NewRequestC(config, "GET", "/")
+	req.WithReporter(reporter)
+	req.WithClient(nil)
+	req.chain.assert(t, failure)
+
+	assert.True(t, reporter.reported)
```
gavv, line 3744: Can you please rework this test to use two reporters?

* first reporter passed to config
* second passed to WithReporter
* we check that failure was reported to the second, and not to the first

request_test.go
```diff
@@ -3726,3 +3726,40 @@ func TestRequest_Panics(t *testing.T) {
...
+
+	req := NewRequestC(config, "GET", "/")
+	req.WithAssertionHandler(&DefaultAssertionHandler{
+		Reporter:  reporter,
+		Formatter: req.config.Formatter,
+	})
+	req.WithClient(nil)
+	req.chain.assert(t, failure)
+
+	assert.True(t, reporter.reported)
```
gavv, line 3764: Can you please rework this test to use two assertion handlers, one passed to config and another to WithAssertionHandler?

Also let's use mockAssertionHandler.

request_test.go
gavv, line 1: Some missing checks:

* we need to add WithReporter and WithAssertionHandler calls to TestRequest_FailedChain
* we need to update TestRequest_Usage (pass nil values)
* we need to update TestRequest_Order

In addition, it would be nice to add a subtests to both TestRequest_Order and TestRequest_AssertionHandler, that ensures that Response created by Expect uses the same reporter / assertion handler (i.e. if we call Expect, obtain response, and make it to report failure, it will report it to appropriate destination).

request_test.go
```diff
@@ -3726,3 +3726,40 @@ func TestRequest_Panics(t *testing.T) {
 		assert.Panics(t, func() { newRequest(newMockChain(t), config, "GET", "") })
 	})
 }
+
+func TestRequest_Reporter(t *testing.T) {
+	reporter := newMockReporter(t)
+
+	config := Config{
+		RequestFactory: DefaultRequestFactory{},
```
gavv, line 3734: I think RequestFactory is not needed here? (not sure)

request.go
gavv, line 1: In both WithReporter and WithAssertionHandler, we should check if reporter / handler is nil, and if so report usage error.

Example: https://github.com/gavv/httpexpect/blob/bf8aa9cf97484b8271bf64d9cb1728a4044bf5df/request.go#L270-L278

request.go
gavv: Same for WithAssertionHandler

request.go
```diff
@@ -1926,6 +1926,62 @@ func (r *Request) WithMultipart() *Request {
...
+	return r.WithAssertionHandler(&DefaultAssertionHandler{
+		Reporter:  reporter,
+		Formatter: r.config.Formatter,
+	})
+}
+
+// WithAssertionHandler sets assertion handler.
+//
+// The new handler overwrites assertion handler. It will be used to
+// format and report test Failure or Success.
```
gavv, line 1959: Let's also mention that it will be used in child objects too (Response and so on)

gavv:
I see some commits; please ping when it's ready for review

diPhantxm:
> I see some commits; please ping when it's ready for review

I think it's ready for review. I've fixed all the problems except 
> Also let's use mockAssertionHandler

`mockAssertionHandler` doesn't have a reporter, which I use to check if it reported failure

>In addition, it would be nice to add a subtests to both TestRequest_Order and TestRequest_AssertionHandler, that ensures that Response created by Expect uses the same reporter / assertion handler (i.e. if we call Expect, obtain response, and make it to report failure, it will report it to appropriate destination).

If it's necessary, I will do it, but for now I decided to leave it as it is

gavv:
Thanks for update!

I've pushed some cosmetic fixes, including using of mockAssertionHandler (we can actually easily use it).

> If it's necessary, I will do it, but for now I decided to leave it as it is

No worrier, I'll create a separate issue for that.

Issue #352, opened 2023-03-29
Convert Object tests to table-driven style


gavv:
Convert each subtest of the following tests to table-driven style:

* TestObject_ContainsKey
* TestObject_ContainsValue
* TestObject_ContainsSubset
* TestObject_HasValue

`TestNumber_IsEqual/basic` can be used as a reference for table-driven style.

For subtests that do only a few checks (e.g. `TestNumber_IsEqual/invalid argument` has only two checks), instead of table we can use the same approach as in `TestNumber_IsInt/invalid argument`, where we create a fresh number for each check.

Darrellbor:
@gavv i'd like to help out with this

gavv:
You're welcome!

PR #379, opened 2023-04-06
[352]:... convert object tests to table driven style

Darrellbor:
PR for #352 

Implemented table-driven style for:
- [x] TestObject_ContainsKey
- [x] TestObject_ContainsValue
- [x] TestObject_ContainsSubset
- [x] TestObject_HasValue


Darrellbor:
@gavv ready for review

object_test.go
```diff
@@ -665,64 +665,156 @@ func TestObject_InList(t *testing.T) {
...
-	value.chain.clear()
-
-	value.NotContainsKey("bar")
-	value.chain.assert(t, failure)
-	value.chain.clear()
-
-	value.ContainsKey("BAR")
-	value.chain.assert(t, failure)
-	value.chain.clear()
+	testObj := map[string]interface{}{"foo": 123, "bar": ""}
```
gavv, line 668: I think tests will be more readable if we inline (copy-paste) testObj into each test. This way you can quickly understand the test by just looking at the table entry.

We could do it for TestObject_ContainsKey and TestObject_ContainsValue - where testObj is small.

object_test.go
```diff
@@ -665,64 +665,156 @@ func TestObject_InList(t *testing.T) {
...
+			assertion: contains,
+			wantEqual: success,
+		},
+		{
+			name:      "1. correct key value, not contains assertion",
+			object:    testObj,
+			key:       "foo",
+			assertion: notContains,
+			wantEqual: failure,
+		},
```
gavv, line 694: Instead of two separate entries for ContainsKey and NotContainsKey, I suggest to use one entry for both, with two results instead of wantEqual:

* wantContainsKey (success or failure)
* wantNotContainsKey (success or failure)

contains/notContains enum won't be needed.

We use similar approach in other tests.

gavv: Same for other tests in this PR.

object_test.go
```diff
@@ -665,64 +665,156 @@ func TestObject_InList(t *testing.T) {
...
+
+	cases := []struct {
+		name      string
+		object    map[string]interface{}
+		key       string
+		assertion uint
+		wantEqual chainResult
+	}{
+		{
+			name:      "1. correct key value, contains assertion",
```
gavv, line 682: I'd avoid using numbers in subtest names. It'd require to keep attention that numbers should be unique and correctly ordered when editing tests. That looks like unnecessary work.

object_test.go
```diff
@@ -665,64 +665,156 @@ func TestObject_InList(t *testing.T) {
...
-	value.chain.clear()
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			reporter := newMockReporter(t)
+			value := NewObject(reporter, tc.object)
+
+			if tc.assertion == contains {
+				value.ContainsKey(tc.key)
+				value.chain.assert(t, tc.wantEqual)
+				value.chain.clear()
```
gavv, line 733: Could you please use the same approach as, for example, [here](https://github.com/gavv/httpexpect/blob/dad9859e10c0a94862136849954adb6bcfeb1ac8/number_test.go#L824-L825)?

Like:

```go
		NewObject(...).ContainsKey(...).
			chain.assert(...)
``` 

With this approach, each check uses a fresh value, instead of clearing chain and reusing old value. It's more clear and safe because there is no shared state.

object_test.go
gavv: Same for other tests in this PR.

object_test.go
gavv: Same for other tests in this PR.

gavv: Could you also apply this approach to subtests of the modified tests, which you didn't convert to table driven form?

(as mentioned in the last paragraph of the issue)

They include:

* [ ] TestObject_ContainsValue/struct
* [ ] TestObject_ContainsValue/canonization
* [ ] TestObject_ContainsValue/invalid argument
* [ ] TestObject_ContainsSubset/struct
* [ ] TestObject_ContainsSubset/canonization
* [ ] TestObject_HasValue/struct
* [ ] TestObject_HasValue/canonization

gavv (commented):

Darrellbor:
@gavv ready for re-review

gavv:
@Darrellbor What do you think about this comment? https://github.com/gavv/httpexpect/pull/379#discussion_r1159973052

Darrellbor:
I checked and used this method, i think it made sense

gavv:
Let's keep it as is. Thanks!

Issue #387, opened 2023-04-06
Convert Array tests to table-driven style


gavv:
Convert each subtest of the following tests to table-driven style:

* [x] TestArray_IsEmpty
* [x] TestArray_IsEqual
* [x] TestArray_IsEqualUnordered
* [x] TestArray_InList

`TestNumber_IsEqual/basic` can be used as a reference for table-driven style.

For subtests that do only a few checks (e.g. `TestNumber_IsEqual/invalid argument` has only two checks), instead of table we can use the same approach as in `TestNumber_IsInt/invalid argument`, where we create a fresh number for each check.

aliziyacevik:
hi @gavv , PR is ready for this issue. I can open it if you could assign me. Thanks.

gavv:
Great!

PR #389, opened 2023-04-07
converted array tests to table-driven style

aliziyacevik:
https://github.com/gavv/httpexpect/issues/387

### DIFF
--- a/array_test.go
+++ b/array_test.go
@@ -619,10 +619,11 @@ func TestArray_IsEqualUnordered(t *testing.T) {
 
 		reporter := newMockReporter(t)
 
-		NewArray(reporter, []interface{}{123, 456, "foo"}).IsEqualUnordered(myArray{myInt(456), 123.0, "foo"}).
+		testVal := myArray{myInt(456), 123.0, "foo"}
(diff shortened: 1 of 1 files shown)


gavv:
Thanks for PR, will review in upcoming days.

A few preliminary comments.

1. Could you please use the same approach as, for example, [here](https://github.com/gavv/httpexpect/blob/dad9859e10c0a94862136849954adb6bcfeb1ac8/number_test.go#L824-L825)?

   Like:

   ```go
		NewArray(...).SomeCheck(...).
			chain.assert(...)
   ``` 

   With this approach, each check uses a fresh value, instead of clearing chain and reusing old value. It's more clear and safe because there is no shared state.

2. As mentioned in the issue, if subtest has only one or two checks, I think the table is an overkill; you can just make those two checks, in the style provided above.

3. If in some subtest wantXxx and wantNotXxx always have opposite values, there is no need to have both. You can have only wantXxx, and instead of wantNotXxx you can use !wantXxx.

aliziyacevik:
Thanks for your comment, I did what you asked, hope you like it. @gavv e

gavv:
Thanks for PR, I've pushed commit with further refactoring: 451a4b362b1d1bf4c89fc3de12fb061021cd908b

Issue #255, opened 2023-01-30
Create reentrancy test for chain


gavv:
We need to be sure that `chain` does not invoke AssertionHandler under chain's lock. It's forbidden because AssertionHandler may be provided by user, and if user code will indirectly access the chain again, and the chain is locked, it would cause deadlock.

This already works as needed, but we need to cover this behavior with a unit test for chain.

The test will:

* set custom AssertionHandler which Success and Failure methods invoke chain.enter

* invoke chain methods to make it reporting success and failure 

* check that AssertionHandler methods were actually called

Here we simulate accessing chain from user code. If chain calls user code under its lock, this test will hang.

We already have a similar test for Environment: TestEnvironment_Reentrancy. It checks that AssertionHandler may access Environment and it wont cause deadlock.

regularcoder:
I'm interested in taking a shot at this.

gavv:
You're welcome!

gavv:
Landed

PR #391, opened 2023-04-08
Add reentrancy tests for chain assertion handler (#255)

regularcoder:
Tests for #255, I verified that adding Lock() calls in chain.go prior to calling handler.Success() or handler.Failure() caused the respective test to hang.

Please review @gavv

### DIFF
--- a/chain_test.go
+++ b/chain_test.go
@@ -17,6 +17,47 @@ func testFailure() AssertionFailure {
 	}
 }
 
+func TestChain_Reentrancy(t *testing.T) {
+	t.Run("Success", func(t *testing.T) {
+		assertionHandler := &mockAssertionHandler{}
+
+		chain := newChainWithConfig("root", Config{
+			AssertionHandler: assertionHandler,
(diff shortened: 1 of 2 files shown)


gavv:

Two preliminary comments:

* could you please merge two tests into one with two subtests?

* instead of creating new mock type reenteringAssertionHandler, could you please reuse mockAssertionHandler and adjust it for your needs?

   * add successCalled and failureCalled fields (BTW better counters than booleans)
   * add custom callback field which you can use to make call to `mh.chain.env()`; use mockReporter and its reportCb field as an example

gavv:
Also please make reentrancy test to be the very first test (like it is for other structs).

regularcoder:
Thank you for the feedback. Incorporated your comments.

gavv:
LGTM!

> I verified that adding Lock() calls in chain.go prior to calling handler.Success() or handler.Failure() caused the respective test to hang.

Appreciate that, thanks.

PR #400, opened 2023-04-11
refactor json path and schema tests

### LINKED ISSUE #359: Refactor json path and json schema tests
Tests for many matcher structs (e.g. Array, Object), have a test called TestXxx_Getters (e.g. TestArray_Getters, TestObject_Getters), which contain checks for struct's `Path` and `Schema` methods.

We need to find all these checks and extract them into separate tests TestXxx_Path and TestXxx_Schema. Please place new tests before TestXxx_Getters.

Besides extraction, we need to refactor the checks themselves. Instead of performing a series of checks on the same value and clearing the chain every time, we should perform a check on a new instance. This way checks will be more isolated (they won't use shared state) and clean (we won't mutate chain internal state from tests).

For example, consider this set of checks:

```go
value.Schema(`{"type": "object"}`)
value.chain.assert(t, success)
value.chain.clear()

value.Schema(`{"type": "array"}`)
value.chain.assert(t, failure)
value.chain.clear()
```

It can be refactored in one of the following ways:

1. We can re-recreate value for each check and re-assing var:

    ```go
	var value *Object

	value = Object(...)
	value.Schema(`{"type": "object"}`)
	value.chain.assert(t, success)

	value = Object(...)
	value.Schema(`{"type": "array"}`)
	value.chain.assert(t, failure)
    ```

2. Alternatively, we can split the test into subtests and create a variable inside each of them.

3. Alternatively, we can write the test in table-driver style.

Feel free to choose one of the approaches for each test based on your own judgement.

antara-gandhi:
fixes #359 

extracted all TestXXX_Getters that included Path and Schema tests into TestXXX_Path, TestXXX_Schema, and Test_Getters individual, respective tests. 

additionally, refactored each set of Schema tests into the following format: 

- create new instance of value based on first referenced type
- perform Schema check for first type
- perform Schema check for second type 

I considered splitting the tests into subtests and creating a new variable inside each one, but that ended up being too much repeat code. Although it may have been slightly more readable to a new user, this version is cleaner and more maintainable.


antara-gandhi:
@gavv PR is ready for review! Lmk if there are any changes to be made. Thanks!

gavv:
Nice!

Could you please remove now unnecessarily cslls to chain.clear in the end of some tests?

Also, could you rename TestXxx_Getters tests that now check only Raw method to TestXxx_Raw?

Thanks

antara-gandhi:
Should be done now @gavv! I wasn't sure for TestString_Getters though - I think we still need the `value.chain.clear()` in it but correct me if I am wrong. Also, since the test is also checking `num.chain.assert(t, success)`, I kept the test name as TestString_Getters instead of TestString_Raw, but again lmk if I should change that.

gavv:
LGTM!

Issue #381, opened 2023-04-06
Free memory after rewinds are disabled in bodyWrapper


gavv:
When bodyWrapper.DisableRewinds is called, we forbid further rewinds. From that point, user can only read response once until EOF or error.

However, if we have response in memory, we do not free this memory even though it's not needed anymore.

To free it, it's enough to do:

```go
bw.memReader = nil // or maybe bytes.NewReader(nil)
bw.memBytes = nil
```

The tricky part is that DisableRewinds can be called while we're in the middle of reading response from memory. In this case, DisableRewinds can't free memory right now; instead, it should be freed only when read encounters EOF.

I think no new tests are needed for this feature, it would be enough to modify existing tests, specifically TestBodyWrapper_DisableRewinds, which already has checks for allocated memory in various cases.

gpjservais:
Hey @gavv, could I work on this one?

gavv:
Sure!

gavv:
Landed via 6ff151ac30802786a1263874225a450f6f41df24, thanks!

PR #402, opened 2023-04-13
Update to free memory after disabling rewinds for bodyWrapper

gpjservais:
This addresses issue #381.

So far in this PR, cached HTTP response body is freed from memory in two cases:
1. `DisableRewinds` is called and the full HTTP response body has not been cached in memory.
2. A read from memory via `memReadNext` occurs and hits a EOF.

---
There are two additional cases where memory is not freed until a read from memory after `DisableRewinds` occurs (e.g. the test case at body_wrapper_test. go line 843):
1. Disabling rewinds immediately after a read via `httpReadNext`.
  - When switching to memReadNext, memReader is set to a new reader on nil. After disabling rewinds, only an additional attempt to Read will trigger a read from memory, which will result in an EOF.
2. Disabling rewinds for cases where a subsequent read will result in EOF.
  - You could assume that an additional read will occur and treat this as a case that's still in the middle of reading a response from memory.

I'm currently investigating if these could be moved to occur within the `DisableRewinds` method.

One option to address these additional cases could be to update the memReader type from `io.Reader` to `bytes.Reader`. This would allow us to implement additional methods like `Len()` ([see documentation](https://pkg.go.dev/bytes#NewReader)) rather than just the Read interface method. The we could do something like:
```
```

### DIFF
--- a/body_wrapper.go
+++ b/body_wrapper.go
@@ -186,14 +186,26 @@ func (bw *bodyWrapper) DisableRewinds() {
 	bw.mu.Lock()
 	defer bw.mu.Unlock()
 
+	// Free memory if still reading from original HTTP response.
+	if !bw.isFullyRead {
+		bw.memReader = bytes.NewReader(nil)
+		bw.memBytes = nil
+	}
+
 	bw.isRewindDisabled = true
 }
 
 func (bw *bodyWrapper) memReadNext(p []byte) (int, error) {
 	n, err := bw.memReader.Read(p)
 
-	if err == io.EOF && bw.readErr != nil {
-		err = bw.readErr
+	if err == io.EOF {
+		if bw.isRewindDisabled {
+			bw.memReader = bytes.NewReader(nil)
+			bw.memBytes = nil
+		}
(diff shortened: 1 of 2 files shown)


gpjservais:
In my last commit, I added the following:
- Converted `memReader` from io.Reader to *bytes.Reader. This allows us to make use of the Len() method for *bytes.Reader to free memory for `memBytes` when running `DisabledRewinds` if there's nothing left to read.
- Added functionality to the `Close` method to free memory after resetting the memory reader. This addresses the case where `DisableRewinds` is performed in the middle of a read from memory, right before running close. I also added a separate test case for it.

gavv:
Hi thanks a lot for the patch, and sorry for so long delay. I finally found some time to sort out PRs in this project.

The implementation looks fine, thanks for digging into it! I like the idea with bytes.Reader. I've merged it via 6ff151ac30802786a1263874225a450f6f41df24 (just squashed commits into one)

Also clarified comments a bit: 65e65e50fc180e9a8e583231a561a17bbd562ae1

Issue #383, opened 2023-04-06
Convert TestRequest_Conflicts to table-driven style


gavv:
Convert each subtest of TestRequest_Conflicts to table-driven style.

Use TestRequest_Usage and TestRequest_Order as a reference (but they have only one table, and TestRequest_Conflicts should have multiple tables, one per subtest).

Darrellbor:
is this task taken? @gavv @diPhantxm

diPhantxm:
> is this task taken? @gavv @diPhantxm

Hi. Yeah, I thought it wouldn't take so much time, so I decided not to assign to the task

gavv:
@Darrellbor Yep, there is a PR already.

@diPhantxm Thanks, merged!

gavv:
@diPhantxm Ideally, please leave a comment in an issue when you're working on a task. Apparently someone also also didn't find your PR and made a duplicate one: https://github.com/gavv/httpexpect/pull/404

PR #403, opened 2023-04-13
refactor TestRequest_Conflicts

diPhantxm:
#383
Hi @gavv. I refactored all the test, but there's one case I don't know how to refactor:
```Golang
req = NewRequestC(config, "GET", "url")
req.WithFileBytes("a", "a", []byte("a"))
req.chain.assert(t, failure)
```
It was in `multipart conflict`

UPDATE: seems that lint failed with timeout. I don't think that's my fault

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -3252,99 +3252,86 @@ func TestRequest_Conflicts(t *testing.T) {
 	}
 
 	t.Run("body conflict", func(t *testing.T) {
-		var req *Request
+		cases := []func(req *Request){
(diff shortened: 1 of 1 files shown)


gavv:
Thanks!

Follow-up: f4feb78e152617e6f2c718cb8845918d0ef5ec29

* add subtest names (for easier debugging, otherwise error log is too big because of many checks in one test)
* covert that WithFileBytes call in TestRequest_BodyMultipart (it's actually not a conflict test)

PR #404, opened 2023-04-14
Refactor TestRequest_Conflicts to table-driven style

saeid-a:
Hi @gavv, I refactored the requested test case similar to the other test cases with the following format.

```go
testCases := []struct {
		name       string
		beforeFunc func(req *Request)
		afterFunc  func(req *Request)
}{
```


gavv:
@saeid-a Hi, there was another PR solving this issue: https://github.com/gavv/httpexpect/pull/403

I've merged that PR because it was first. Anyway, thanks!

Issue #343, opened 2023-03-22
Colorize http requests and responses


gavv:
Follow-up for https://github.com/gavv/httpexpect/issues/159

Currently `defaultFailureTemplate` colorizes requests/responses as follows:

```
{{ .Request | indent | trim | color $.EnableColors "HiMagenta" }}
{{ .Response | indent | trim | color $.EnableColors "HiMagenta" }}
```

The following improvement is suggested:

* add new function "colorhttp" to defaultTemplateFuncs and use it for request and response
* if colors are disabled, the function will return original value
* the function will apply the following colorization:
  * in lines like `GET /test HTTP/1.1` it will highlight `GET` (method) and `HTTP/1.1` (protocol)
  * in lines like `HTTP/1.1 200 OK 229.642µs` it will highlight `HTTP/1.1` (protocol) and `200 OK` (status code)
  * in lines like `Host: 127.0.0.1:38681` it will highlight `Host:` (header field name)
* the suggested color for highlighting is HiMagenta (in line with #334)

Example for testing:

```
$ FORCE_COLOR=1 go test -v -run TestE2EReport_Names
=== RUN   TestE2EReport_Names
    e2e_report_test.go:52: 
        invalid "Content-Type" response header
          mime: no media type
        
        test name: TestExample
        
        request name: RequestExample
        
        request: GET /test HTTP/1.1
          Host: 127.0.0.1:35779
        
        response: HTTP/1.1 200 OK 246.75µs
          Date: Wed, 22 Mar 2023 18:38:49 GMT
          Content-Length: 0
        
        assertion:
          Request("GET").Expect().JSON()
        
        actual value:
          ""
--- PASS: TestE2EReport_Names (0.00s)
PASS
ok  	github.com/gavv/httpexpect/v2	0.003s
```

Related: #334, #335.

gdguesser:
Hey @gavv, can I take this one? 🙂

FillerFree:
Hey @gavv can I pick this task if @gdguesser is not working on this?. I have made code changes for this. https://github.com/gavv/httpexpect/pull/414

PR #414, opened 2023-04-27
CLose https://github.com/gavv/httpexpect/issues/343

FillerFree:
This wil close feature: https://github.com/gavv/httpexpect/issues/343

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -953,6 +953,57 @@ var defaultTemplateFuncs = template.FuncMap{
 		}
 		return color.New(colorAttr).Sprint(input)
 	},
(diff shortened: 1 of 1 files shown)


gavv:
Hi, thanks for the patch and sorry for delay.

Looks good. I've pushed a small follow-up commit: e6879c0c3e358e8400f3fc5e9677a48ceb661740

Changes:

- use Split() instead of Fields() to ensure that we won't squash subsequent spaces into one
- use SplitN to avoid altering tail part of the string (after header field) which we are not interested in
- change color scheme a bit
- fix indentation

Issue #409, opened 2023-04-25
Add test for reporter and assertion handler inheritance


gavv:
Add a test (with several subtests) to expect_test.go which checks the following use case:

* set reporter in config to REP1
* create two requests REQ1 and REQ2
* on REQ2, call WithReporter to set reporter to REP2
* call Expect on both requests to obtain two responses RESP1 and RESP2
* on both responses, make some assertion that fails
* check that:
  * RESP1 reported failure to REP1 (inherited from REQ1 which inherited it from config)
  * RESP2 reported failure to REP2 (inherited from REQ2 on which we called WithReporter)

Also add a similar test, but instead of Config.Reporter and Request.WithReporter, use Config.AssertionHandler and Request.WithAssertionHandler.

gavv:
Thanks!

PR #419, opened 2023-05-01
TestReporterAndAssertionHandler implementation

Darrellbor:
PR for #409 

checked that:
- RESP1 reported failure to REP1 (inherited from REQ1 which inherited it from config)
- RESP2 reported failure to REP2 (inherited from REQ2 on which we called WithReporter)
- RESP1 & RESP2 reported failure with Config.AssertionHandler and Request.WithAssertionHandler instead


Darrellbor:
@gavv I attempted writing the test for this but I'm not too sure if it's what's required exactly...

Darrellbor:
@gavv this is ready for review

gavv:
Thanks for the patch.

> I attempted writing the test for this but I'm not too sure if it's what's required exactly...

Yeah, actually this is not exactly what I was talking about. The key point is to check *which* reporter (or assertion handler) is used when failure happens on first or second response (first response inherits reporter from config, second from WithReporter on request).

I've reworked the test in a follow-up commit: 50355b80b56e935f740cc971d01df6023aebaf2f

Issue #413, opened 2023-04-26
Add macOS and Windows to github actions


gavv:
Currently our CI runs checks only on Linux.

We need to add macOS and Windows to the matrix of "build" and "examples" jobs, and fix possible issues.

AnandaIlyasa:
hi @gavv i want to try this one, assign me please

gavv:
@AnandaIlyasa Welcome, thanks!

reapedjuggler:
Hi, @gavv Is this in progress or shall I take it up? Thank you 🚀

neda1985:
@gavv does somebody already work on this issue or i can pick it ?

gavv:
Fixed in #426

PR #426, opened 2023-07-12
Adding macos and windows to github actions #413

FloSchl8:
PR for #413 

- added windows and macos in github actions for build, test and examples
- also replaced ioutil (deprecated since Go 1.16)
- Go version moved to 1.x for tests since some dependencies required Go >= 1.20

PR #428, opened 2023-09-29
feat(gha): support macos

georgettica:
related to #413 

not sure if someone did or didn't just though that as it's taking for so long I will just start and ask forgiveness if someone has already started and close mine


gavv:
Hi, thanks for PR, but we already had one: #426

PR #429, opened 2023-10-04
feat: add support for custom retry functions

### LINKED ISSUE #427: Add `WithRetryHandler` for a custom `RetryPolicy`
Context:
My test is as simple as:

```go
e.GET("/nginx").
		WithMaxRetries(10).
		WithRetryDelay(time.Second, time.Second*2).
		WithRetryPolicy(httpexpect.RetryAllErrors).
		Expect().
		Status(http.StatusNotFound).
		Body().Contains(`nginx/`)
```

I'm expecting a dynamic result from one call to the other. And I'd like to handle the retry policy myself here. Whuch does not seems doable right now.

So something like:

```go
e.GET("/nginx").
		WithMaxRetries(10).
		WithRetryDelay(time.Second, time.Second*2).
		WithRetryPolicy(httpexpect.RetryAllErrors).
		WithRetryHandler(func(resp *http.Response, err error) bool {
			// Check body if available, etc.
			return ...
		})
		Expect().
		Status(http.StatusNotFound).
		Body().Contains(`nginx/`)
```

What do you think ?

nkzren:
Closes #427

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -46,6 +46,9 @@ func TestRequest_FailedChain(t *testing.T) {
 	req.WithMaxRedirects(1)
 	req.WithRetryPolicy(RetryAllErrors)
 	req.WithMaxRetries(1)
+	req.WithCustomRetryFunc(func(r *http.Response, err error) bool {
+		return true
+	})
 	req.WithRetryDelay(time.Millisecond, time.Millisecond)
 	req.WithWebsocketUpgrade()
 	req.WithWebsocketDialer(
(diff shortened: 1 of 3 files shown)


request_test.go
```diff
@@ -3991,3 +4002,52 @@ func TestRequest_Panics(t *testing.T) {
 		assert.Panics(t, func() { newRequest(newMockChain(t), config, "GET", "") })
 	})
 }
+
+func TestRequest_RetriesCustomFunc(t *testing.T) {
```
rafiramadhana, line 4006: I think we should cover for both
- `WithCustomRetryFunc` that returns `true`
- `WithCustomRetryFunc` that returns `false`

Wdyt?

nkzren:
Hey, I'm dropping this since it's been a while and I'm currently stuck with other affairs, sorry

PR #431, opened 2023-10-11
Bump golangci-lint version and fix new warnings (closes #424)

### LINKED ISSUE #424: Linting outputs errors
I would like to contribute, but when running `make lint` on master I get errors like this:

```
$ make lint && echo $?
...
object.go:1204:2: unused-parameter: parameter 'opChain' seems to be unused, consider removing or renaming it as _ (revive)
	opChain *chain, obj map[string]interface{}, key string,
	^
mocks_test.go:79:25: unused-parameter: parameter 'failure' seems to be unused, consider removing or renaming it as _ (revive)
	ctx *AssertionContext, failure *AssertionFailure,
	                       ^
mocks_test.go:127:48: unused-parameter: parameter 'typ' seems to be unused, consider removing or renaming it as _ (revive)
func (mp *mockWebsocketPrinter) WebsocketWrite(typ int, content []byte, closeCode int) {
                                               ^
mocks_test.go:131:47: unused-parameter: parameter 'typ' seems to be unused, consider removing or renaming it as _ (revive)
func (mp *mockWebsocketPrinter) WebsocketRead(typ int, content []byte, closeCode int) {
                                              ^
mocks_test.go:155:46: unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _ (revive)
func (mc *mockWebsocketConn) SetReadDeadline(t time.Time) error {
                                             ^
mocks_test.go:159:47: unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _ (revive)
func (mc *mockWebsocketConn) SetWriteDeadline(t time.Time) error {
                                              ^
mocks_test.go:167:43: unused-parameter: parameter 'messageType' seems to be unused, consider removing or renaming it as _ (revive)
func (mc *mockWebsocketConn) WriteMessage(messageType int, data []byte) error {
                                          ^
formatter.go:356:20: unused-parameter: parameter 'ctx' seems to be unused, consider removing or renaming it as _ (revive)
	data *FormatData, ctx *AssertionContext, failure *AssertionFailure,
...
make: *** [lint] Error 1
2
```
I checked some of these files, and indeed the parameters are not being used in the function bodies. According to go docs, the compiler doesn't complain about unused function parameters because a developer may be writing a function with a specific signature in order to implement an interface. If this is the case, here, then there must be a way of ignoring these lint errors.

```
(master)$ go version && golangci-lint version
go version go1.20.4 darwin/arm64
golangci-lint has version 1.53.2 built with go1.20.4 from 59a7aaf on 2023-06-03T10:44:21Z
```
Happy to help with this, if this indeed is an oversight.

SergeAx:
Ah, GitHub isn't looking into the MR title. Okay, here's a tag for them: #424

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,7 +119,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         timeout-minutes: 20
         with:
-          version: v1.51.2
+          version: v1.54.2
           working-directory: ${{ matrix.dir }}
           args: --timeout=20m
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,10 @@ linters:
 linters-settings:
   lll:
     line-length: 90
+  revive:
+    rules:
+      - name: unused-parameter
+        disabled: true
 
 issues:
   exclude-rules:


gavv:
Hi! As I wrote in the issue, I'd prefer to disable this diagnostic: https://github.com/gavv/httpexpect/issues/424#issuecomment-1742875802

SergeAx:
Actually, about half of these warnings were interface implementations, but as you command)

gavv:
Thank you!

> Actually, about half of these warnings were interface implementations, but as you command)

Thanks :) Keeping argument names in these interfaces implementations is a kind of documentation - you may see what's the meaning of the arguments. And since unused parameters don't harm the code much, I'd like to keep them.

PR #436, opened 2023-11-13
Added standalone TLS example 

### LINKED ISSUE #205: Add standalone TLS example
Add new example to `_examples` directory (say, `tls.go` and `tls_test.go`) that runs small TLS server with hardcoded certificates and tests it using httpexpect. Example can be based on https://go.googlesource.com/go/+/dev.boringcrypto/src/crypto/tls/example_test.go

Add link to new example to README: https://github.com/gavv/httpexpect#examples

yoktoJH:
Closes #205


gavv:
Thanks. Small follow-up commit: 561d43e96bee03d443d291c4fd6a891769372d66

Those hard-coded certs have expired and tests were failing, so I re-generated them.

PR #446, opened 2024-05-21
Adds a new Query() method that works like Path()

stubents:
but uses ohler/ojg to evaluate the jsonPath expression.


yk-eukarya:
@gavv are there any plans to merge this PR?

PR #469, opened 2025-08-05
fix: preserve query from RequestFactory

vmihailenco:
`httpexpect.Request` holds a copy of query in `r.query`, but fails to copy/sync the query set in the RequestFactory.

### DIFF
--- a/request.go
+++ b/request.go
@@ -208,6 +208,10 @@ func (r *Request) initReq(opChain *chain, method string) {
 	}
 
 	r.httpReq = httpReq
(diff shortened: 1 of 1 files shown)


vmihailenco:
I also tried to write a test for `TestExpect_RequestFactory`, but it is not easy because `encodeRequest` is not called there and this code is never executed:

```go
func (r *Request) encodeRequest(opChain *chain) bool {
	r.httpReq.URL.Path = concatPaths(r.httpReq.URL.Path, r.path)

	if r.query != nil {
		r.httpReq.URL.RawQuery = r.query.Encode()           <--- here the query is overriden
	}
```

Issue #461, opened 2025-03-04
Support negative indexes in Array


gavv:
Support negative indexes in:

- `Array.Value`
- `Array.HasValue`
- `Array.NotHasValue`

Semantics should be the same as in Python. This feature is useful when you want to check boundary values in the array beginning and end. It becomes especially useful with deprecation of First and Last methods.

Bound checks in corresponding methods should be updated. Currently we check and report `AssertInRange` with `Min = 0, Max = len-1`. Now it would be convenient to report this range only if index is `>= 0`, and to check and report `Min = -len, Max = -1` range if index is negative.

Unit tests and documentation comments for all 3 methods should be updated as well.

ManojGadam:
Hi, is this issue still available? I would like to work on it

jgfranco17:
Hello is this issue still open? I would like to give it a shot.

antriantsek:
is there any assigns left for this issue?

PR #470, opened 2025-08-18
FEAT: Support negative indexes in Array

jgfranco17:
## Description

Add support for negative indexes in `Array` methods. Syntax follows style of Python, where the `-n` means "nth element from the right".

### Related issues

- https://github.com/gavv/httpexpect/issues/461

## Changes

- Add normalization step in `Value`, `HasValue`, and `NotHasValue` methods
- Update table tests

## Testing

Unit tested with Go Test


jgfranco17:
Linter failures seem unrelated to changes, should I address these in separate PR or is it okay to leave them be?

PR #476, opened 2026-06-05
bump go to 1.25 and update go deps

Nexucis:
Hello,

I would like to update the dependencies of this project. It's been a while it has not been done :)


e2e/chunked_test.go
```diff
@@ -39,8 +39,9 @@ func createChunkedFastHandler(t *testing.T) fasthttp.RequestHandler {
 	return func(ctx *fasthttp.RequestCtx) {
 		headers := map[string][]string{}
 
-		ctx.Request.Header.VisitAll(func(k, v []byte) {
```
Nexucis, line 42: Visitall is a deprecated function, that's why I have replaced it

array.go
```diff
@@ -715,7 +715,7 @@ func (a *Array) IsEmpty() *Array {
 		return a
 	}
 
-	if !(len(a.value) == 0) {
+	if len(a.value) > 0 {
```
Nexucis, line 718: it's easier and faster to read like that when you do not have to calculate in memory the boolean expression

boolean.go
```diff
@@ -117,7 +117,7 @@ func (b *Boolean) IsTrue() *Boolean {
 		return b
 	}
 
-	if !(b.value == true) {
+	if b.value {
```
Nexucis: value is already a boolean, so comparing to another boolean does not make really sense

formatter.go
```diff
@@ -796,13 +796,10 @@ func (f *DefaultFormatter) applySeparator(numStr string, dir int) string {
 	switch f.DigitSeparator {
 	case DigitSeparatorUnderscore:
 		separator = "_"
-		break
```
Nexucis, line 799: break is not necessary in Golang in a switch.

request.go
```diff
@@ -2601,25 +2601,22 @@ func (r *Request) setupRedirects(opChain *chain) {
...
 			})
-			return
-		}
-	} else {
-		if r.redirectPolicy != defaultRedirectPolicy || r.maxRedirects != -1 {
-			clientCopy := *httpClient
-			httpClient = &clientCopy
-			r.config.Client = &clientCopy
 		}
+		return
```
Nexucis, line 2613: here there was a possibility for `httpClient` to be nil. So I have moved the `return` statement to avoid this situation

Nexucis:
I don't get why the build is failing. If you are interested by this PR, do you mind to guide me @gavv ?

